PR #20, opened 2018-11-25
Fix template conditional in readme

spheromak:
The template as presented would result in an error.

---
This change is [Reviewable](https://reviewable.io/reviews/twpayne/chezmoi/20)


twpayne:
Many thanks for your contribution!

Issue #9, opened 2018-11-20
Build and distribute RedHat packages


twpayne:
`chezmoi` should be easily installable on RedHat systems. We should build and distribute a `.rpm` package.

twpayne:
Wow, thank you. goreleaser looks really nice and a PR adding support for it would be fantastic. I don't have a packagecloud.io repo, and the lowest price plan is $75/month, so let's start initially with Github releases.

I've tagged the current version as `v0.0.1` so you have a valid semantic version tag to start with.

If you don't have time to make a PR then please say and I'll have a look.

joemiller:
@twpayne working on it now

joemiller:
@twpayne I have a work in progress here: https://github.com/twpayne/chezmoi/compare/master...joemiller:goreleaser

Before sending a PR I wanted to ask if you had a preference on the release workflow. There are two different workflows I've used with goreleaser:

**Automated tagging and releasing.**  Combine autotag (https://github.com/pantheon-systems/autotag) with goreleaser for automated releases. In this mode, the workflow for releases goes something like this:
- a successful `master` build generates a new release: Autotag will increment the patch version automatically and push the tag back to github, eg: `v0.0.1` -> `v0.0.2`
- To create a new MINOR release: include `[minor]` or `#minor` in the git commit message. eg: `v0.0.1` -> `v0.1.0`
- To create a new MAJOR release: include `[major]` or `#major` in the git commit message. eg: `v0.0.1` -> `v1.0.0`

**Manually push tags to generate releases.** In this model, in order to create a new release you would create a tag locally and push to github, eg:
```
$ git tag -a v0.1.0 -m "First release"
$ git push origin v0.1.0
```

My team tends to prefer the autotag method since it conforms to the "what is on master is what is released" philosophy. The second method gives you a bit more control over when to release.

Let me know if you have a preference and I'll update the WIP and send the PR. I'm also testing on my own travis-ci account to ensure the releases work all the way through.

joemiller:
example releases here: https://github.com/joemiller/chezmoi/releases

twpayne:
In the short term, let's start with manual releases - `chezmoi` isn't quite settled yet, and I want to make internal changes (e.g. improved testing, doc fixes) without bumping the point release on each one. Once things are more stable, we can switch to the autotag approach.

Thanks for your work on this @joemiller - much appreciated.

Issue #7, opened 2018-11-20
Make chezmoi installable with homebrew


twpayne:
`chezmoi` should be easily installable with [Homebrew](https://brew.sh). This can easily be done by either:

* Creating a new tap, or
* Adding `chezmoi` to the default Homebrew tap.

PR #33, opened 2018-11-27
add goreleaser for building and pushing releases

joemiller:


.goreleaser.yaml
```diff
@@ -0,0 +1,129 @@
...
+        386: i686
+        arm: armhfp
+        arm64: aarch64
+    deb:
+      name_template: "{{ .ProjectName }}_{{ .Version }}-{{ .Env.TRAVIS_BUILD_NUMBER }}_{{ .Os }}_{{ .Arch }}"
+      replacements:
+        386: i386
+        arm: armel
+
+## generate a homebrew formula and push to github.com/twpayne/homebrew-taps
```
joemiller: If you want to host your own homebrew tap, do the following:

1. Create a new github repo `twpayne/homebrew-taps`
2. Uncomment this `brew:` block
3. Commit to master
4. create and push a new tag to github

When Goreleaser runs it will create a new file `chezmoi.rb` in the `twpayne/homebrew-taps` repo under the `Formula` directory. This allows macOS users to install with:

```console
brew install twpayne/taps/chezmoi
```

The kubefwd project has a good example of using goreleaser to manage a homebrew tap:

- https://github.com/txn2/kubefwd/blob/master/goreleaser.yml
- https://github.com/txn2/homebrew-tap/blob/master/Formula/kubefwd.rb

The homebrew formulas generated by goreleaser are not compatible with "homebrew-core", BTW. So if you intend this app to be part of the core homebrew package list (which it should be!) then this won't be necessary but it's great to have in the meantime.

.goreleaser.yaml
```diff
@@ -0,0 +1,129 @@
...
+#     owner: twpayne
+#     name: homebrew-taps
+#   commit_author:
+#     name: Tom Payne
+#     email: TODO@TODO.tld
+#   folder: Formula
+#   homepage: "https://github.com/twpayne/chezmoi"
+#   description: "chezmoi is a tool for managing your dotfiles across multiple machines."
+
+## generate and push docker images:
```
joemiller: Uncomment if you want to build and push docker images.

.goreleaser.yaml
```diff
@@ -0,0 +1,129 @@
...
+#
+project_name: chezmoi
+
+builds:
+- binary: chezmoi
+  env:
+    - CGO_ENABLED=0
+  ldflags:
+     - -s -w
+  goos:
```
joemiller: I went a bit overboard with the OS's and architectures, but goreleaser makes it so easy it is hard not to resist.

I selfishly request these platforms because it's where I intend to run chezmoi:
* linux/amd64
* openbsd/amd64
* macos

.goreleaser.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - linux
+    - darwin
+    # TODO: uncomment when compatible with windows. currently fails on 'undefined: syscall.Stat_t'
```
joemiller: I noticed I couldn't compile on windows at the moment. If that's ever supported you can comment out this line and you'll start generating windows/x86-64 releases

joemiller:
implements #9

.goreleaser.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - '^docs:'
+    - '^test:'
+    - Merge pull request
+    - Merge branch
+
+## generate RPM and DEB packages
+nfpm:
+  vendor: "TODO"
+  homepage: "https://github.com/twpayne/chezmoi"
+  maintainer: Tom Payne <TODO@TODO.tld>
```
joemiller: You probably want to modify this. I'm not sure if you need an email address here. My guess is that the field is generic and name + email is just a convention.

.goreleaser.yaml
```diff
@@ -0,0 +1,129 @@
...
+        arm: armel
+
+## generate a homebrew formula and push to github.com/twpayne/homebrew-taps
+# brew:
+#   github:
+#     owner: twpayne
+#     name: homebrew-taps
+#   commit_author:
+#     name: Tom Payne
+#     email: TODO@TODO.tld
```
joemiller: you'll want to update this. It'll be used for the git commit message generated during CI to the homebrew-taps repo

twpayne:
Thanks so much for this brilliant PR @joemiller. I'll work through the instructions you've given me to activate the commented-out bits over the next few days.

joemiller:
n/p

also noticed this covers some of #7

.goreleaser.yaml
twpayne: Right now, I don't think `chezmoi` is the sort of command that needs a Docker container, but I'll keep the commented-out code in case it becomes a requirement later.

PR #37, opened 2018-11-29
Activate goreleaser

twpayne:
This fixes a few TODOs left for me by @joemiller in #33.


joemiller:
lgtm! 👍 

could add a note in the readme for macos users to install the home brew tap... i think it'd be:  `brew install twpayne/taps/chezmoi`

twpayne:
Thanks, let me tag a new version and do an actual release - then I can also update the README to add installation instructions for the different OSes.

Issue #133, opened 2019-01-16
Add pass support


tasmo:
The yet supported password managers are well selected. Usually I use "[the standard unix password manager](https://www.passwordstore.org/)" which even has multi repo support.

`pass` should be a standard as well.

twpayne:
Great idea, will add it.

twpayne:
Fixed by @johnswanson in #138.

PR #138, opened 2019-01-17
Add support for pass

johnswanson:
As discussed in https://github.com/twpayne/chezmoi/issues/133

For now, this only takes the first line of output from `pass` (the same behavior as `pass -c`).

Thanks for chezmoi, very cool project!


tasmo:
Thank you for reacting this quick!

Issue #170, opened 2019-01-26
Add init command


fgrosse:
After testing chezmoi on different systems for a while I want to propose to add a new `init` command.

This command would be run when chezmoi is initially setup on a host. Its purpose is to handle the checkout of the `~/.local/share/chezmoi` directory from the source repository as well as setting up permissions along that path (i.e. `0700`).

This is useful because `~/.local/share` does usually not exist when I am installing dotfiles on a new _server_. In that case setting up the directories can be annoying. Also the init command can do submodule setup which I use for my dotfiles repo (e.g. for [antigen][1]). Maybe in the future it could also be extended to ask the user for parameters to automatically generate `~/.config/chezmoi/chezmoi.yaml` based on some definition in the source repo but I would not go that far yet and instead defer this topic, maybe into a later issue.

Example usage (tested inside fresh docker containter):

```sh
$ chezmoi init -v git@github.com:example/dotfiles.git
mkdir -m 700 /root/.local
mkdir -m 700 /root/.local/share
Created source directory "/root/.local/share" with file mode 0700 (-rwx------) 
Checking out dotfiles repository "git@github.com:example/dotfiles.git" into "/root/.local/share/chezmoi"
exec git clone git@github.com:example/dotfiles.git /root/.local/share/chezmoi
Enter passphrase for key '/root/.ssh/id_rsa': 
Cloning into '/root/.local/share/chezmoi'...

chmod 700 /root/.local/share/chezmoi
Initializing git submodules in "/root/.local/share/chezmoi"
cd /root/.local/share/chezmoi
exec git submodule init
Submodule 'antigen' (https://github.com/zsh-users/antigen.git) registered for path 'antigen'
```

If you agree that this could be a valuable addition I can provide a pull request (I implemented the command already on my fork). If however you do not want to add this to chezmoi then I will probably fall back to maintaining a bash setup script.

---

P.S. Thanks for this project. I think its a great idea and really motivated me to cleanup my dotfiles :sweat_smile: 

[1]: https://github.com/zsh-users/antigen

twpayne:
Yes, this is a great idea! Please do submit a PR.

twpayne:
Code in #174 merged, thank you!

fgrosse:
Came back here after I got distracted with other shiny things :)

I see chezmoi is doing great and still developing further :+1: 
Regarding this topic I would like to continue with this one part I mentioned in the issue description:

[...]

PR #150, opened 2019-01-17
Add first run git init instructions to README

bashfulrobot:
Good day, as per https://github.com/twpayne/chezmoi/issues/149 I am submitting a PR to add git init instructions on a fresh install for the README file.

Thank you in advance.

PR #174, opened 2019-01-27
Init command

fgrosse:
This PR implements the change I proposed in #170 

I had some problems testing it because of the exec that is required during init (for git commands). I wanted to avoid any bigger refactoring so I ended up not using the existing `Config.exec` function but instead provide a different one that does not use `syscall.Exec`.

Another noteworthy thing here is that I also made the `chezmoi init` command apply the source dir in order to avoid always having to run `chezmoi init  … && chezmoi apply`). I left a flag to opt out but I think its up to you to decide if that flag is really required or should be removed to reduce the surface of the CLI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,16 @@ For a full list of commands run:
 
     $ chezmoi help
 
+## Setting up from existing source
+
+If you put your source directory under version control you will eventually want
+to check it out on another host. `chezmoi` can do this for you and make sure all
+directory permissions are setup correctly:
+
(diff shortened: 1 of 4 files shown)


fgrosse:
This PR is somewhat related to #150

twpayne:
Thanks so much for this! I'm jumping on a plane right now and wanted a clean git history, so I've split this in to two PRs (#175 and #176) and merged them.

Thank you so much for the contribution!

PR #175, opened 2019-01-27
Init command

twpayne:
This is #174 but without a typo-fix commit that conflicted with `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,16 @@ For a full list of commands run:
 
     $ chezmoi help
 
+## Setting up from existing source
+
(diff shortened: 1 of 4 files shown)


fgrosse:
I really like the changes you added to the init command :+1: .

Thanks for merging this so fast, now I can continue integrating it deeper into my setup. As a follow up I will probably open a new issue to discuss whether we could ask the user to supply the parameters for chezmoi configuration (i.e. the ones used on templating). I should have time for that next weekend :grimacing:

PR #182, opened 2019-01-29
Consider files only containing whitespace as empty

twpayne:


twpayne:
cc @offbyone

offbyone:
You rock!

PR #183, opened 2019-01-29
fix: several plausible failure modes for the script during routine run

offbyone:
I found some cases where my dotfiles on another machine caused issues. So, more robust!


twpayne:
Thanks for the improvements :)

PR #187, opened 2019-01-29
Add a check for the ignore file in the `doctor` command

offbyone:
I got wondering whether I'd created the right ignore file (that double-i really is hard to brain around)

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


twpayne:
Interesting idea. I agree that the double "i" in `.chezmoiignore` is annoying. Sadly I'd picked the name `chezmoi` before designing the ignore functionality.

For this PR, there are a couple of problems:
* It's not a problem if the user doesn't have a `.chezmoiignore` file.
* `.chezmoiignore` files can appear anywhere in the source dir.

For these reason, I'd keep the idea of the test but instead walk the source dir (using `vfs.Walk`) and warn if any files called `.chezmoignore` (single "i") files are found. Feel free to update the PR or I can do this if you don't have time.

twpayne:
Thanks for the idea! Closing this in favor of #190.

Issue #192, opened 2019-01-31
Unable to add files -- "outside target directory"


offbyone:
On one of my servers, admittedly running an older Linux, I'm seeing this:

```
~/> chezmoi -v add snoop/*.sh
chmod 700 /home/offbyone/.local/share/chezmoi
chezmoi: /local/home/offbyone/snoop/cron_run.sh: outside target directory
~/> ls -ld /home/offbyone/.local/share/chezmoi
drwx------ 12 offbyone users 4096 Jan 30 16:14 /home/offbyone/.local/share/chezmoi/
~/> chezmoi --version
chezmoi version 1.3.0, commit 8d19d62fe1272ea162e4f881b9d56cc2b5990bfe, built at 2019-01-30T06:27:27Z
 ~/> chezmoi doctor
     ok: /home/offbyone/.local/share/chezmoi (source directory, perm 700)
     ok: /home/offbyone (destination directory, perm 750)
     ok: /home/offbyone/.config/chezmoi/chezmoi.yaml (configuration file)
     ok: /bin/vi (editor)
     ok: /usr/bin/git (source VCS command, version 2.16.5)
warning: op (1Password CLI, not found)
warning: bw (Bitwarden CLI, not found)
warning: lpass (LastPass CLI, not found)
warning: pass (pass CLI, not found)
warning: vault (Vault CLI, not found)
```

I can't figure this one out; this is the only host where I'm encountering the issue, so I'm suspicious that it's something about the OS.

offbyone:
```
~> pwd
/local/home/offbyone
```

I'm in `/local/home/offbyone` as well.

offbyone:
Ah, I see the surface problem: if `$HOME` is a symlink -- and it is, for me -- then the two values don't agree. The realpath of `$HOME` is `/local/home/offbyone` while the value of the `$HOME` environment variable is `/home/offbyone`.

twpayne:
Interesting. I don't think #193 is quite right -- see the comment there.

#194 might help.

offbyone:
I'd love to try it, but I can't build it on the machine that experiences the issue (no compiler modern enough to handle it) and I don't know enough Golang to cross-compile.

twpayne:
Assuming that you're in the `chezmoi` working copy checked out somewhere, you can cross compile with:

```
GOOS=linux GOARCH=amd64 go build -o chezmoi-linux-amd64 .
```

This will produce the `chezmoi-linux-amd64` Linux binary.

[...]

PR #193, opened 2019-01-31
Fix for #192 - set the DestDir to an absolute path

offbyone:
File adds are done using `filepath.Abs()` on the inputs, so make the `DestDir`
consistent so that when destdir contains symlinks it doesn't break resolution.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -283,6 +283,7 @@ func getDefaultData(fs vfs.FS) (map[string]interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	data["homedir"] = homedir
 
 	hostname, err := os.Hostname()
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks for the PR, but I don't think that this works quite the way you think it does. The destination directory is almost always already an absolute path, so `filepath.Abs()` should be no-op. `filepath.Abs()` also calls `filepath.Clean()` on the result, and this side-effect might be what is causing it to fix the problem in your case.

offbyone:
This update is based on #194 now, and at least in my usecase resolves the issue by resolving `$HOME` to a realpath if possible.

twpayne:
I added an explicit test for this in #252 but was unable to reproduce the problem. Please re-open if needed.

PR #196, opened 2019-02-01
Fix weird corner case

twpayne:
This fixes an unlikely-to-be-encountered failure mode. This could happen if you ran `chezmoi -n edit-config` (dry run mode) where the parent directory of the config file did not exist.

#195 used the most recent version of [`github.com/twpayne/go-vfs`](https://github.com/twpayne/go-vfs) that uses a minimal interfaces as argument to `vfs.Mkdir`. As a consequence, the `Mutator` interface needed a `Stat` method. The `Mutator` interface changed from being a write-only interface to a read-write interface. The question was: how should the `NullMutator` (which is supposed to do nothing) behave when read methods were called?

Prior to this PR, the `NullMutator.Stat` returned `nil` for the `os.FileInfo`, but this could lead to a panic if any methods were called on the `os.FileInfo` return value. This PR replaces the `NullMutator` with a `FSMutator` on a read-only filesystem. Consequently, `Stat` calls can return valid non-`nil` values (meaning no panics) while also ensuring that there is no modification of the underlying filesystem, as required by dry run mode.


twpayne:
OK, so this broke stuff. Reverting.

The problem is in the `Stat` call. Returning no error and a `nil` `os.FileInfo` means that the application stuff code will dereference the `os.FileInfo`, which leads to a panic. Better to pretend that the path doesn't exist.

Issue #205, opened 2019-02-03
alternate files? custom attributes in files names?


niksfirefly:
What do you  think about simple alternate files based on file suffix like in yadm
https://thelocehiliosan.github.io/yadm/docs/alternates
or custom files attributes/suffixes bound to some actions defined in .yaml config file?

I think yadm solution is very pretty and self explanatory

regards

twpayne:
This is an interesting idea. It's definitely easier than using templates if the files are significantly different across machines. Let me ponder on the idea for a while, but I suspect I'll end up adding this.

niksfirefly:
Is there any chance you implement this idea?
I see many advantages of chezmoi but this is the best and killer yadm option :)

twpayne:
Yes, I'd like to implement this. I'm not sure when I'll have time though.

twpayne:
After some thought, I think this can be achieved in chezmoi right now. The trick is to use a templated symlink.

Say you have two `.bashrc`, one for `darwin` and one for `linux`. You'd create the following in your source directory (warning: not tested, but should work):

`symlink_dot_bashrc.tmpl`:
```
.bashrc_{{ .chezmoi.os }}
```

`dot_bashrc_darwin`:
```
# .bashrc on darwin
```

`dot_bashrc_linux`:
```
# .bashrc on linux
```

`.chezmoiignore`
```
{{ if ne .chezmoi.os "darwin" }}
.bashrc_darwin
{{ end }}
{{ if ne .chezmoi.os }}
.bashrc_linux
{{ end }}
```

This setup means that `.bashrc` will be a symlink to `.bashrc_darwin` on `darwin` and `.bashrc_linux`, and that only one OS-specific `.bashrc_os` file will be installed on each OS.

Advantages of this system:

* Separate dotfiles for different machines.
* Fully customizable symlinks using whichever variables you want (yadm only allows a few predefined variables).
* Works with chezmoi's existing machinery now.

Disadvantages:

* Requires more manual setup compared to yadm (but you only have to do this once per file).
* Uses symlinks rather than writing real files.

On balance, I think this is a better solution than yadm's, so I don't plan to implement the original proposal in this PR. I'm happy to hear different perspectives however :)

PR #206, opened 2019-02-03
WIP: Script execution

schrej:
I was previously using Ben Alman's [dotfiles](github.com/cowboy/dotfiles), which contain an "init" step. It basically executes a few scripts that are in a directory. It has a prompt that allows disabling and enabling scripts before execution each time his `dotfiles` script is run.
It was very useful for stuff that can't be handled by copying files, like installing some tools or OS configuration.

As I was missing that feature in chezmoi, I started adding it. I know the contributing guidelines recommend to open an issue first, but I had time today and wanted to write some go.

The features I had in mind:
- A specific directory that contains scripts in the chezmoi source folder (`.scripts`, as that get's ignored and is good name)
- All scripts directly within that folder get detected (no recursion)
- Scripts run either once or every time the subcommand is called (currently configured by adding `#chezmoi: once` to the script)
- Template support, also to handle different operating systems

I've added a subcommand called `scripts` that runs all scripts. It supports passing a list of scripts to run as well as the `--prompt` flag.
Scripts are ran with the binary defined using a Shebang (`#!`) within the script.
Scripts support the template syntax, like all files do. This allows to handle different OSes and so on.
The "run once" feature isn't implemented yet, and because of that the `--force` flag isn't either.

I've opened this PR mainly to get feedback on the Idea itself, as well as some open decisions & future ideas I have:
- To allow single execution chezmoi needs to remember whether a script was executed already. I'm not sure how & where to store that, I was thinking in an extra file next to the local config.
- Whether the name of the scripts folder should be configurable (probably yes)
- How to handle dry-runs (maybe by just outputting the templated script?)
- A `edit-script` command to edit scripts
- adding os(<osname>) to the `#chezmoi:` to limit it to the specified os

Tests are also missing.

### DIFF
--- a/cmd/edit-script.go
+++ b/cmd/edit-script.go
@@ -1,0 +1,76 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/twpayne/chezmoi/lib/chezmoi"
+	vfs "github.com/twpayne/go-vfs"
+)
+
+var editScriptsCmd = &cobra.Command{
+	Use:   "edit-scripts scripts",
+	Args:  cobra.MinimumNArgs(1),
+	Short: "Edit scripts in the scripts folder",
+	RunE:  makeRunE(config.runEditScriptCmd),
+}
+
+type editScriptsCmdConfig struct {
+	run    bool
+	prompt bool
+}
+
+func init() {
+	rootCmd.AddCommand(editScriptsCmd)
+
+	persistentFlags := editScriptsCmd.PersistentFlags()
+	persistentFlags.BoolVarP(&config.editScripts.run, "run", "r", false, "run after editing")
+	persistentFlags.BoolVarP(&config.editScripts.prompt, "prompt", "p", false, "prompt before running")
+}
+
+func (c *Config) runEditScriptCmd(fs vfs.FS, args []string) error {
+	ts, err := c.getTargetState(fs)
+	if err != nil {
+		return err
+	}
+	//entries, err := c.getEntries(ts, args)
+	var scripts []*chezmoi.Script
+	for _, arg := range args {
+		if s, ok := ts.Scripts[chezmoi.StripTemplateExtension(arg)]; ok {
+			scripts = append(scripts, s)
+		}
+	}
+	println(scripts)
+	argv := []string{}
+	for _, s := range scripts {
+		argv = append(argv, s.SourcePath)
+	}
+	if !c.editScripts.run {
+		return c.execEditor(argv...)
+	}
+	if err := c.runEditor(argv...); err != nil {
+		return err
+	}
+
+	for _, s := range scripts {
+		if c.editScripts.prompt {
+			choice, err := prompt(fmt.Sprintf("Apply %s", s.Name), "ynqa")
+			if err != nil {
+				return err
+			}
+			switch choice {
+			case 'y':
+			case 'n':
+				continue
+			case 'q':
+				return nil
+			case 'a':
+				c.editScripts.prompt = false
+			}
+		}
+		if err := s.Apply(c.DestDir, c.DryRun); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/cmd/edit-script.go
+++ b/cmd/edit-script.go
@@ -33,14 +33,13 @@ func (c *Config) runEditScriptCmd(fs vfs.FS, args []string) error {
 	if err != nil {
 		return err
 	}
-	//entries, err := c.getEntries(ts, args)
 	var scripts []*chezmoi.Script
 	for _, arg := range args {
 		if s, ok := ts.Scripts[chezmoi.StripTemplateExtension(arg)]; ok {
 			scripts = append(scripts, s)
 		}
 	}
-	println(scripts)
+
 	argv := []string{}
 	for _, s := range scripts {
 		argv = append(argv, s.SourcePath)
--- a/cmd/edit-script.go
+++ b/cmd/edit-script.go
@@ -53,7 +53,7 @@ func (c *Config) runEditScriptCmd(fs vfs.FS, args []string) error {
 
 	for _, s := range scripts {
 		if c.editScripts.prompt {
-			choice, err := prompt(fmt.Sprintf("Apply %s", s.Name), "ynqa")
+			choice, err := c.prompt(fmt.Sprintf("Apply %s", s.Name), "ynqa")
 			if err != nil {
 				return err
 			}
--- a/cmd/scripts.go
+++ b/cmd/scripts.go
@@ -42,16 +42,6 @@ func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
 		return ts.ApplyScripts(fs, config.scripts.force)
 	}
 
-	var quit int
-	defer func() {
-		if r := recover(); r != nil {
-			if p, ok := r.(*int); ok && p == &quit {
-				err = nil
-			} else {
-				panic(r)
-			}
-		}
-	}()
 	var scripts []*chezmoi.Script
 	if len(args) == 0 {
 		for _, s := range ts.Scripts {
@@ -68,7 +58,7 @@ func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
 					scripts = append(scripts, s)
 				case 'n':
 				case 'q':
-					panic(&quit) // abort vfs.Walk by panicking
+					return nil
 				}
 			} else {
 				scripts = append(scripts, s)
@@ -90,7 +80,7 @@ func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
 						scripts = append(scripts, s)
 					case 'n':
 					case 'q':
-						panic(&quit) // abort vfs.Walk by panicking
+						return nil
 					case 'a':
 						c.scripts.prompt = false
 					}
--- a/cmd/scripts.go
+++ b/cmd/scripts.go
@@ -1,0 +1,24 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	vfs "github.com/twpayne/go-vfs"
+)
+
+var scriptsCmd = &cobra.Command{
+	Use:   "scripts [targets...]",
+	Short: "Run scripts that need to run",
+	RunE:  makeRunE(config.runScriptsCmd),
+}
+
+func init() {
+	rootCmd.AddCommand(scriptsCmd)
+}
+
+func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
+	ts, err := c.getTargetState(fs)
+	if err != nil {
+		return err
+	}
+	return ts.ApplyScripts()
+}
--- a/cmd/scripts.go
+++ b/cmd/scripts.go
@@ -28,18 +28,13 @@ func init() {
 }
 
 func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
-	if config.DryRun {
-		println("chezmoi: the scripts subcommand doesn't support dry-run yet")
-		return nil
-	}
-
 	ts, err := c.getTargetState(fs)
 	if err != nil {
 		return err
 	}
 
 	if len(args) == 0 && !config.scripts.prompt {
-		return ts.ApplyScripts(fs, config.scripts.force)
+		return ts.ApplyScripts(fs, config.scripts.force, c.DryRun)
 	}
 
 	var scripts []*chezmoi.Script
@@ -55,7 +50,7 @@ func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
 					c.scripts.prompt = false
 					fallthrough
 				case 'y':
-					if err := s.Apply(); err != nil {
+					if err := s.Apply(c.DestDir, c.DryRun); err != nil {
 						return err
 					}
 				case 'n':
@@ -79,7 +74,7 @@ func (c *Config) runScriptsCmd(fs vfs.FS, args []string) error {
 					}
 					switch choice {
 					case 'y':
-						if err := s.Apply(); err != nil {
+						if err := s.Apply(c.DestDir, c.DryRun); err != nil {
 							return err
 						}
 					case 'n':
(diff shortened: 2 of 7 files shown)


twpayne:
Thanks very much for this, it's very promising.

> To allow single execution chezmoi needs to remember whether a script was executed already. I'm not sure how & where to store that, I was thinking in an extra file next to the local config.

Sounds reasonable. I suspect that it might be a good idea to record which scripts have been run by a hash (e.g. SHA256) of their contents. This way, if the user adds or modifies a script then it can be re-run. This would also encourage the use of fine-grained scripts (e.g. a separate script for each package that you want installed, rather than one script that installs all packages). I think we'll also need a mechanism to mark scripts as run and reset this status. 

> Whether the name of the scripts folder should be configurable (probably yes)

Yes. I'd also consider having separate directories for scripts that are run just once vs. scripts that should be run every time. This is cleaner than parsing that script for `#chezmoi: once`.

> How to handle dry-runs (maybe by just outputting the templated script?)

👍 

> A edit-script command to edit scripts

👍

> adding os() to the #chezmoi: to limit it to the specified os

Please hold off implementing this right now as it overlaps with #205 and it makes sense to use a consistent mechanism.

Some general comments:

1. `chezmoi` tries hard to be idempotent. Having arbitrary scripts breaks this as the scripts can do arbitrary things. It's a shame, but I think the scripts functionality is sufficiently useful for it to be worthwhile. For what it's worth, my current personal solution is to use a [`Makefile`](https://github.com/twpayne/dotfiles/blob/master/Makefile) and relying on my memory to know what I need to run where, but I think this scripts approach is cleaner.

2. Rather can parsing the shebang line to determine the interpreter, I'd write the script contents to a temporary file, make that file executable, and then run that. This gives a bit more flexibility (users can add binaries as scripts if they really want), means less code in `chezmoi`, and a single source of truth for how the shebang line is interpreted (the kernel).

3. The scripts should always have a consistent working directory (`chezmoi`'s destination directory) irrespective of where the `chezmoi` command is run from (the `Dir` field of [`os/exec.Cmd`](https://godoc.org/os/exec#Cmd) is your friend here).

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"regexp"
+)
+
+var shebangRegex *regexp.Regexp
+var onceRegex *regexp.Regexp
```
twpayne: Please put these in a `var` block name them `Regexp` (with a `p` at the end) for consistency.

```go
var (
    shebanRegexp = // ...
    onceRegexp   = // ...
)
```

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,97 @@
...
+
+func init() {
+	shebangRegex = regexp.MustCompile(`(?m)^#!([^;\n]+)$`)
+	onceRegex = regexp.MustCompile(`(?m)^#\s*chezmoi:\s*once$`)
+}
```
twpayne: The rexgexps can be initialized where they are declared, no need for an `init` function.

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Script is a script supposed to run
```
twpayne: Nit: finish the comment with punctation (`.`).

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	s.executor = string(reg[1])
+	s.once = onceRegex.Match(s.contents)
+	return nil
+}
+
+func (s *Script) execute() error {
+	c := exec.Command(s.executor)
+	c.Dir = path.Dir(s.sourcePath)
+	in, err := c.StdinPipe()
```
twpayne: `c.Stdin` should be connected to `os.Stdin` in case the script is interactive.

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,97 @@
...
+		return fmt.Errorf("Shebang missing in script \"%s\"", s.Name)
+	}
+	s.executor = string(reg[1])
+	s.once = onceRegex.Match(s.contents)
+	return nil
+}
+
+func (s *Script) execute() error {
+	c := exec.Command(s.executor)
+	c.Dir = path.Dir(s.sourcePath)
```
twpayne: `c.Dir` should be the destination directory.

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,97 @@
...
+	if len(reg) < 2 {
+		return fmt.Errorf("Shebang missing in script \"%s\"", s.Name)
+	}
+	s.executor = string(reg[1])
+	s.once = onceRegex.Match(s.contents)
+	return nil
+}
+
+func (s *Script) execute() error {
+	c := exec.Command(s.executor)
```
twpayne: Rather than computing `c.executor`, instead write the script to temporary file and execute that.

lib/chezmoi/scripts.go
schrej: I did it this way as the script might need to access a file in the scripts folder (e.g. homebrew's Brewfile).
We could also expose a `CHEZMOI_SCRIPTS_PATH` environment variable for that though.

lib/chezmoi/chezmoi.go
```diff
@@ -94,9 +95,27 @@ func sortedEntryNames(entries map[string]Entry) []string {
...
 	return strings.Split(path, string(filepath.Separator))
 }
+
+// StripTemplateExtension removes the template extension at the end of the passed filename
+func StripTemplateExtension(name string) string {
+	i := strings.LastIndex(name, templateSuffix)
+	if i > 0 {
+		return name[0:i]
+	}
+	return name
```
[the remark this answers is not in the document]
schrej: I've kept it as a method as `templateSuffix` is not exposed (like the other constants) and it gets used in the command outside of the `chezmoi` package.

schrej:
> Yes. I'd also consider having separate directories for scripts that are run just once vs. scripts that should be run every time. This is cleaner than parsing that script for `#chezmoi: once`.
If we implement #205, maybe we could add `!once` (for single execution that is remembered) and `!manual` (or `!never` for execution that will only happen explicitly).
`!manual` is obviously debatable, but it could be useful for rarely used scripts. You could execute them manually of course, but I'd include it just for completeness.

> 1. `chezmoi` tries hard to be idempotent.
I know. This is the main reason for the "remember you executed this already" feature. I'd go for run every time by default to encourage users to write idempotent scripts and include the "once" option to avoid manual implementation of it.

> I suspect that it might be a good idea to record which scripts have been run by a hash (e.g. SHA256) of their contents. This way, if the user adds or modifies a script then it can be re-run.
While I like this idea, it might lead to unexpected behaviour. If just minor things get changed, maybe on another machine, the script might run on another machine on an update without the user expecting it to do so, which is against the idea of being idempotent.

cmd/edit-script.go
```diff
@@ -0,0 +1,76 @@
...
+	persistentFlags.BoolVarP(&config.editScripts.run, "run", "r", false, "run after editing")
+	persistentFlags.BoolVarP(&config.editScripts.prompt, "prompt", "p", false, "prompt before running")
+}
+
+func (c *Config) runEditScriptCmd(fs vfs.FS, args []string) error {
+	ts, err := c.getTargetState(fs)
+	if err != nil {
+		return err
+	}
+	//entries, err := c.getEntries(ts, args)
```
twpayne: This looks like debug code. Please remove it.

cmd/edit-script.go
```diff
@@ -0,0 +1,76 @@
...
+	var scripts []*chezmoi.Script
+	for _, arg := range args {
+		if s, ok := ts.Scripts[chezmoi.StripTemplateExtension(arg)]; ok {
+			scripts = append(scripts, s)
+		}
+	}
+	println(scripts)
```
twpayne: This looks like debug code. Please remove it.

cmd/scripts.go
```diff
@@ -0,0 +1,107 @@
...
+					}
+				}
+			}
+		}
+	}
+	for _, s := range scripts {
+		if err := s.Apply(); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
twpayne: This still looks like scripts are run all together at the end, rather than one-by-one after each prompt.

lib/chezmoi/target_state_test.go
```diff
@@ -137,6 +137,7 @@ func TestTargetStatePopulate(t *testing.T) {
 						contents:   []byte("bar"),
 					},
 				},
+				Scripts: map[string]*Script{},
```
twpayne: I don't think these empty maps are needed. They _should_ compare OK if they are left uninitialized (i.e. `nil`).

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,93 @@
...
+	if s.contentsErr != nil {
+		return s.contentsErr
+	}
+
+	if err := s.parse(); err != nil {
+		return err
+	}
+	fmt.Printf("chezmoi: Running script %s\n", s.Name)
+	if dryRun {
+		println(string(s.contents), "\n")
```
twpayne: Please don't use the `println` function, it is a leftover from an old version of go. Use `fmt.Println` instead, and only print if running in verbose mode.

lib/chezmoi/chezmoi.go
```diff
@@ -19,6 +19,7 @@ const (
 	executablePrefix = "executable_"
 	dotPrefix        = "dot_"
 	templateSuffix   = ".tmpl"
+	scriptsDir       = ".scripts"
```
twpayne: Please use separate directories for "run once" scripts and "run every time" scripts instead of checking for `chezmoi:once` regexps in the scripts.

lib/chezmoi/target_state_test.go
schrej: I tested it, they are required.
```
--- FAIL: TestTargetStatePopulate/script_force_executable (0.00s)
        target_state_test.go:530: ts.Populate(&{PathFS:{fs:0x17858e0 path:/var/folders/7b/_plxss852cn6ch8nfc4djxy40000gn/T/go-vfs-vfst228732403} tempDir:/var/folders/7b/_plxss852cn6ch8nfc4djxy40000gn/T/go-vfs-vfst228732403 keep:false}) diff:
            modified: .Scripts = map[string]*chezmoi.Script{}
```

cmd/edit-script.go
schrej: That's for running all remaining scripts at the end. When you select 'a', that's what get's invoked.
I've cleaned it up a little, I hope its more readable now.

cmd/scripts.go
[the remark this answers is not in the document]
schrej: That loops runs the rest of the scripts when pressing `a`. I've rewritten it now, so its a little clearer.

lib/chezmoi/chezmoi.go
schrej: I really don't like having separate directories. Could you read and comment on [this](https://github.com/twpayne/chezmoi/pull/206#issuecomment-465315877) (especially the first one) maybe?

schrej:
Sorry for the long pause, had issues with my notebook (MacBook chargers suck).

lib/chezmoi/scripts.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if err := f.Chmod(os.FileMode(0700)); err != nil {
+		return err
+	}
+	c := exec.Command(f.Name())
+	c.Dir = destDir
+	c.Env = append(os.Environ(), "CHEZMOI_SCRIPT_DIR="+path.Dir(s.SourcePath))
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	c.Stdin = os.Stdin
```
twpayne: Please use `c.Stdout()` and friends here. This was only recently added, but facilitates testing.

lib/chezmoi/chezmoi.go
twpayne: Thanks, I'll read [this](https://github.com/twpayne/chezmoi/pull/206#issuecomment-465315877).

Please can you make the scripts directory be `.chezmoiscripts` for consistency with `.chezmoiignore`.

lib/chezmoi/chezmoi.go
```diff
@@ -95,9 +96,23 @@ func sortedEntryNames(entries map[string]Entry) []string {
...
 	if strings.HasPrefix(path, string(filepath.Separator)) {
 		path = strings.TrimPrefix(path, string(filepath.Separator))
 	}
 	return strings.Split(path, string(filepath.Separator))
 }
+
+// StripTemplateExtension removes the template extension at the end of the passed filename
+func StripTemplateExtension(name string) string {
+	return strings.TrimSuffix(name, templateSuffix)
+}
```
twpayne: I suspect that we should be have `ScriptAttributes` and `Script` structs in the same way as we have a [`FileAttributes` and `File` structs](https://github.com/twpayne/chezmoi/blob/master/lib/chezmoi/file.go#L14-L34) (and also for [directories](https://github.com/twpayne/chezmoi/blob/master/lib/chezmoi/dir.go#L12-L26)), because scripts are effectively a new top-level concept in `chezmoi` but we can make this change in a follow-up PR.

twpayne:
For the tests to pass, please rebase on the latest `master`.

lib/chezmoi/target_state_test.go
twpayne: This might have changed following https://github.com/twpayne/chezmoi/pull/234. Could you do a quick re-check?

schrej:
I did a rebase already. I also fixed the tests and they pass. Only `goveralls` is failing, as Travis [removes encrypted tokens on PRs](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions) for security reasons.
```
goveralls -coverprofile=coverage.out -service=travis-ci -repotoken $COVERALLS_TOKEN
flag needs an argument: -repotoken
```

twpayne:
Thanks very much for continuing to work on this @schrej - I haven't forgotten about it.

FWIW I'd like to get this in to chezmoi soon - I'm eyeing a v2.0.0 release soon and I'd like this to be in it.

schrej:
Well, I keep forgetting to work on it as well ;)

The tests are actually passing now, but apparently my "fix" from #240 didn't really work...

twpayne:
Thanks very much for all your work on this @schrej. In the end, I've decided to go for a different approach to achieve equivalent functionality, implemented in #287, but it was very much inspired by what you developed here.

PR #215, opened 2019-02-15
Fix double word in init help

mgeisler:


twpayne:
Many thanks @mgeisler :)

PR #225, opened 2019-03-15
Fix #198 - Use os.UserHomeDir() instead of custom function

### LINKED ISSUE #198: Use os.UserHomeDir when Go 1.12 is released
#194 copied `os.UserHomeDir` from the forthcoming Go 1.12 release. Once Go 1.12 is released, `userHomeDir` should be replaced with `os.UserHomeDir`, and the copied function can be deleted. `.travis.yml` will also need to be updated to ensure that Go 1.12 is used to build the final binaries.

kkirsche:
Fix #198 - Use os.UserHomeDir() instead of custom function

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ addons:
     - rpm
 
 go:
-  - 1.11.x
+  - 1.12.x
 
 env:
   - GO111MODULE=on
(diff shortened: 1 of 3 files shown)


kkirsche:
Force push to avoid another commit to change .travis.yml file

twpayne:
Many thanks :)

PR #228, opened 2019-03-19
Fix for #227 - Add ppc64 and ppc64le targets

### LINKED ISSUE #227: Add ppc64 & ppc64le Architecture Targets to .goreleaser.yml
I'm really enjoying using chezmoi in my home-lab, however, my work environment is largely ppc64 (IBM POWER Big Endian) which currently isn't in release target architectures.

I'd like to see both ppc64 and ppc64le added to the goreleaser targets as I'd like to start using chezmoi throughout that environment.

pdemonaco:
Updates goreleaser to target ppc64 and ppc64le. I've run the build on my machine using the command below and it generates working binaries.

```bash
TRAVIS_BUILD_NUMBER=1 goreleaser --snapshot --rm-dist --debug --skip-publish
```

Note that I'm using go 1.12.1 on Gentoo to compile.

I also made a minor tweak to the contributing guide as it linked to the issues page for a different repo.

Closes #227

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -69,6 +69,8 @@ nfpm:
   formats:
     - deb
     - rpm
+  dependencies:
+    - git
   overrides:
     rpm:
       name_template: "{{ .ProjectName }}-{{ .Version }}-{{ .Env.TRAVIS_BUILD_NUMBER }}-{{ .Arch }}"
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -21,6 +21,8 @@ builds:
     - amd64
     - arm
     - arm64
+    - ppc64
+    - ppc64le
   goarm:
     - ''
   ignore:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 # Contributing
 
 `chezmoi` development happens on Github. When contributing, please first [open
-an issue](https://github.com/twpayne/dotfiles/issues/new) to discuss the change
+an issue](https://github.com/twpayne/chezmoi/issues/new) to discuss the change
 that you want to make. Bug reports and documentation improvements are
 particularly welcome.
 


pdemonaco:
Hmm. The build works properly but the Architecture is currently incorrect at least for the RPM packages. It's using linuxppc64 and linuxppc64le instead of ppc64 and ppc64le respectively.

goreleaser/nfpm documentation doesn't really detail the functionality of the override command, however, looking at the code I don't think the architecture can actually be overridden in this way. I think I know how to extend nfpm to correct this behavior.

twpayne:
Thanks for working on this. This is a nice contribution and I'll definitely accept it when it's working.

pdemonaco:
Sure thing. I've got a working fix for goreleaser (the problem is on that end, not in nfpm) which I've tested locally and I'm reading through their commit guidelines to get it submitted. I'll link that PR here once it's up.

pdemonaco:
Here's the corresponding [goreleaser PR](https://github.com/goreleaser/goreleaser/pull/987)

pdemonaco:
@twpayne The goreleaser changes are live and it now generates valid RPMs for ppc64 and ppc64le.

Before you merge I do have another change which might make sense to include in this PR. I've tested adding 'git' as a dependency to the RPM in goreleaser. While technically it's not mandatory to use for chezmoi I imagine it's the standard for most folks. It would certainly be a quality of live improvement when deploying chezmoi to a large environment.

I have the change ready, to push to this branch. Let me know what you think.

twpayne:
It makes sense to add a dependency on `git`. If possible, please make it a weak `rpm` `Recommends:` rather than a hard `Requires: (see https://rpm.org/user_doc/dependencies.html).

pdemonaco:
Both work from the goreleaser side of things, however, the version of RPM used by CentOS/RHEL 7.x doesn't actually support weak dependencies so adding the dependency in that way has no impact in my environment at least. While fedora and possibly the CentOS/RHEL 8 beta do support weak dependencies that leaves out a substantial portion of OSs using RPM.
https://bugzilla.redhat.com/show_bug.cgi?id=1427674

I'd prefer to add it as a regular dependency as this actually solves my use case, however, I'm not sure how this impacts other users of the application. Since git is in the standard repositories for the vast majority of distros it feels like a relatively safe change to me.

twpayne:
OK, it makes sense to add `git` as a regular dependency.

twpayne:
Thanks! I've pushed a new tag `v1.4.1` which should trigger the build of the RPMs. It normally takes about ~10 minutes.

Issue #10, opened 2018-11-20
Activate shell autocompletion


twpayne:
`chezmoi` uses [Cobra](https://github.com/spf13/cobra) for command line parsing, which can also generate autocompletion code. We should use this to generate `bash` and `zsh` autocompletion code.

kkirsche:
For anyone interested in working on this, the documentation specific to this task is:

https://github.com/spf13/cobra/blob/master/bash_completions.md

At least for bash

twpayne:
Fixed by @davewongillies in #231.

PR #158, opened 2019-01-20
Add REFERENCE.md

twpayne:

PR #231, opened 2019-03-22
Add bash & zsh completion

davewongillies:
Made a pass at implementing https://github.com/twpayne/chezmoi/issues/10

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -1,32 +1,34 @@
 package cmd
 
 import (
-	"github.com/spf13/cobra"
-
 	"errors"
 	"os"
+
+	"github.com/spf13/cobra"
+	vfs "github.com/twpayne/go-vfs"
 )
 
-// bashCompletionCmd represents the completion command
 var completionCmd = &cobra.Command{
-	Use:       "completion [shell]",
+	Use:       "completion shell",
 	Short:     "Output shell completion code for the specified shell (bash or zsh)",
-	Long:      "Output shell completion code for the specified shell (bash or zsh)",
-	ValidArgs: []string{"bash"},
-	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) != 1 {
-			return cmd.Usage()
-		}
-		switch args[0] {
-		case "bash":
-			return rootCmd.GenBashCompletion(os.Stdout)
-		case "zsh":
-			return rootCmd.GenZshCompletion(os.Stdout)
-		}
-		return errors.New("Unsupported shell")
-	},
+	ValidArgs: []string{"bash", "zsh"},
+	Args:      cobra.ExactArgs(1),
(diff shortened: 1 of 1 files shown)


cmd/completion.go
```diff
@@ -0,0 +1,32 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
+	"errors"
+	"os"
+)
```
twpayne: Please sort imports with `go imports`. Specifically, Go standard library imports should come before third-party imports.

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+
+// bashCompletionCmd represents the completion command
+var completionCmd = &cobra.Command{
+	Use:       "completion [shell]",
```
twpayne: `shell` is not optional so it shouldn't have square brackets.

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+	Short:     "Output shell completion code for the specified shell (bash or zsh)",
+	Long:      "Output shell completion code for the specified shell (bash or zsh)",
```
twpayne: Please remove `Long` as it's identical to `Short`. Long help will get generated from a reference manual eventually (#158).

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+	ValidArgs: []string{"bash"},
```
twpayne: I think this also needs to include `zsh`.

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
```
twpayne: Please follow the structure of other commands here and use a separate function wrapped with `makeRunE`. The `vfs.FS` argument will be unused, but that's OK.

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+		if len(args) != 1 {
+			return cmd.Usage()
+		}
```
twpayne: Use `cobra.Args: cobra.ExactArgs(1),` in the value of `completionCmd` instead of checking for the length of `args` here.

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+		switch args[0] {
+		case "bash":
+			return rootCmd.GenBashCompletion(os.Stdout)
+		case "zsh":
+			return rootCmd.GenZshCompletion(os.Stdout)
+		}
+		return errors.New("Unsupported shell")
```
twpayne: Please start the error message with a lowercase letter.

cmd/completion.go
```diff
@@ -0,0 +1,32 @@
...
+	},
+}
+
+func init() {
```
twpayne: Please make the `init` function the first function in the file.

Issue #241, opened 2019-04-01
Add method to create an initial config file


twpayne:
In #170, @fgrosse wrote:

> Maybe in the future it could also be extended to ask the user for parameters to automatically generate ~/.config/chezmoi/chezmoi.yaml based on some definition in the source repo but I would not go that far yet and instead defer this topic, maybe into a later issue.

This is that later issue for discussion.

twpayne:
Here's one way of doing it. @fgrosse, I'd be interested in your thoughts.

The template code below is untested, the actual syntax might be slightly different.

In the source directory, if the user has a file called `.chezmoi.yaml.tmpl` then, if no config file already exists, then `chezmoi init` will execute this file as a template and write the output to a new config file in the default location (`.config/chezmoi/chezmoi.yaml`).

We can use a template function to prompt the user for values, for example:

```
data:
  email: "{{ promptString "email" }}"
```

where we define `promptString` as a template function that prompts the user for a string value.

If the same value is needed in multiple places, then we can use a variable:

```
{{ $email := promptString "email" }}
data:
  email: "{{ $email }}"
  mailtoURL: "mailto:{{ $email }}"
```

We'll also need to check for other supported config file formats, e.g. `.chezmoi.toml.tmpl`, `.chezmoi.json.tmpl`, etc., as well as copying the config file verbatim if it doesn't have a `.tmpl` extension.

This approach should provide all the desired functionality while sharing a lot of common with normal `chezmoi` usage.

Thoughts?

fgrosse:
I was just about to open the issue :sweat_smile: 
The idea of using a template to generate this file is quite nice (and bit templateception :sunglasses:).
This provides very good flexibility and is nicely consistent with the way things work in chezmoi.

Some general thoughts:
- Just for clarification so we are on the same page: rendering the config template would be executed by `chezmoi init`, after the checkout but before creating the dotfiles. 
- Should there be a flag to skip this step? Probably no because if you are using templates with custom variables the apply step of the init command will fail with an error anyway (i.e. that's the problem that motivated this request).
- Can it be useful to define variables via environment variables so this remains scriptable and no interactive prompts are strictly required in the init command?
- If later required we can easily add new functions that can be used in the template. However this should be done with care to not bloat this part to much. Anyway just a thought as I'm pretty sure you'll have an eye on this.

[...]

PR #243, opened 2019-04-07
Implement config file creation from template during init command

fgrosse:
This implements using a config file template to bootstrap config files during `chezmoi init` as discussed via #241 

I did my best to cover this with tests but currently its not really possible to test the entire init command end to end. Instead I chose to test the config creation part that I added in isolation so no big refactoring was required.

Cheers


twpayne:
Thanks very much for this - this is pretty much spot on! There are some minor changes that I'll make in a separate PR, but this pretty much nails it. Many thanks for your excellent contribution :)

Issue #142, opened 2019-01-17
Plans for a snap package


bashfulrobot:
Do you plan to add a snap package for this?
 I see that you use goreleaser; which I believe supports it.

twpayne:
I haven't personally used snap, but it looks very nice indeed. So, yes, I will definitely add this.

twpayne:
I've [requested `classic` confinement for `chezmoi`](https://forum.snapcraft.io/t/request-classic-confinement-for-chezmoi/10903/1) and added Snap creation to `chezmoi`'s `.goreleaser.yaml` file. Once the confinement is granted I should be able to publish the snaps.

twpayne:
To update on the status of this, snap package support is currently pending on the following:

* [`classic` confinement request for `chezmoi`](https://forum.snapcraft.io/t/request-classic-confinement-for-chezmoi/10903)
* [Snapcraft support for `ppc64`, `ppc64le`, and `arm` architectures](https://forum.snapcraft.io/t/invalid-architecture-specified-in-the-manifest-ppc64/10904)
* [Improved goreleaser support for snapcraft](https://github.com/goreleaser/goreleaser/issues/1006)

Any contributions to moving any of these forward is much appreciated :)

twpayne:
@bashfulrobot I'm not getting any response from the Snapcraft folk for the [`classic` confinement request for `chezmoi`](https://forum.snapcraft.io/t/request-classic-confinement-for-chezmoi/10903/2). If you have any connections there, could you ping them please?

bashfulrobot:
Can you send me the link to the forum post?

bashfulrobot:
Disregard, found it!

bashfulrobot:
@twpayne I pinged the advocacy team. I suspect part of the delay was the Easter weekend. I however do know that a bunch of them are traveling for a conference. That could be part of the delay as well.

twpayne:
Brilliant, many thanks @bashfulrobot!

ivocavalcante:
Classic confinement is now granted in Snap Store. Can we have this now, @twpayne ?
Thanks!

twpayne:
Yes, it's coming. I have a WIP branch that adds snap support back but haven't had time to complete it yet.

[...]

PR #269, opened 2019-04-27
Remove snap infrastructure

twpayne:
Refs #142.

Sadly the `snapcraft.io` folk are [unresponsive](https://forum.snapcraft.io/t/request-classic-confinement-for-chezmoi/10903).

This PR removes the `snap` infrastructure. It can easily be reverted when the `snapcraft.io` people respond.


ivocavalcante:
@twpayne , you're granted classic confinement on Snap store. I'm following the thread, interested in using chezmoi. Think you can pull back Snap infra now...

PR #275, opened 2019-05-01
Fix destination flag name

2m:


twpayne:
Great catch, thanks for the contribution!

2m:
No problem. Migrating my current `stow` setup to `chezmoi`. Loving it so far! Especially the integration with `lpass`.

PR #285, opened 2019-05-06
Fixed typo in 1Password instructions

robeastham:
Current instructions result in this
```
 eval $(op login <subdomain>.1password.com <email>)
-bash: unexpected EOF while looking for matching `''
-bash: syntax error: unexpected end of file
```
correct instruction below result in user being prompted for Secret Key:

`eval $(op signin <subdomain>.1password.com <email>)`


twpayne:
Great catch, thanks!

Issue #22, opened 2018-11-26
KeePassX integration


twpayne:
[KeePassX](https://www.keepassx.org/) is a popular password manager. `chezmoi` could use this to store secrets, instead of storing them in `~/.chezmoi.yaml`.

This could possibly be exposed through a template function.

twpayne:
As far as I can tell, KeePassX does have an API or a specification of the database file format.

Closing this issue until at least one of the above becomes available.

franek:
It seems that https://github.com/keepassxreboot/keepassxc provide a CLI. I think we should be able to use [genericSecret.command](https://github.com/twpayne/chezmoi/blob/master/docs/HOWTO.md#use-a-generic-tool-to-keep-your-secrets) to use it. keepassxc-cli can be used like that : ```keepassxc-cli show keepass-file.kdbx passwordkey -a Password```

franek:
With : 
```
"genericSecret.command" = "keepassxc-cli"
``` 
into ~/.config/chezmoi/chezmoi.toml

I can run 
```
chezmoi secret generic -- show ./database.kdbx -a Password id
```
with success.

But, templating seems to not work : 
```
{{ secret "-- show ./database.kdbx -a Password id" }}
```
generate something like that : 
```
[85 115 97 103 46]
```
I don't know why.

twpayne:
I think this is because keepassxc-cli needs to read your password from stdin. I have a branch in progress to add KeePassXC support. Watch this space.

twpayne:
Fixed by #293.

franek:
Many thanks. It is awesome. It works ;)

PR #293, opened 2019-05-17
Add KeePassXC support

twpayne:


twpayne:
Fixes #22. @franek are you able to test this? See the changes to the documentation for how to use it.

twpayne:
A key assumption in this PR is that each user only has a single KeePassXC database file. I think this is reasonable, but please tell me if the assumption is incorrect.

twpayne:
@droidmonkey, CC'ing you out-of-the blue on this as you're a current major active developer of KeePassXC. Would appreciate any input you have on KeePassXC/chezmoi integration.

droidmonkey:
Cool! I think for the purposes of chezmoi a single database assumption is fine. Any secrets needed would be in that database.

twpayne:
Many thanks @droidmonkey 👍

PR #335, opened 2019-06-09
Exec new version after upgrade

### LINKED ISSUE #329: chezmoi upgrade shows old version after upgrading via snap
`chezmoi upgrade` upgrades chezmoi to the latest version. As a final check, it exec's `chemzoi --version` using [`os.Executable`](https://godoc.org/os#Executable) to find the executable.

This works for installations where chezmoi is installed either as a binary download or via a package manager, as the path does not change. With snap it does not work because the new executable is installed in a different location. Consequently, when chezmoi is installed with snap, the old version of chezmoi is printed after a `chezmoi upgrade`, even though the new version is installed.

This should be fixed.

twpayne:
Fixes #329.

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -163,8 +163,21 @@ func (c *Config) runUpgradeCmd(fs vfs.FS, args []string) error {
 		return fmt.Errorf("invalid --method value: %s", method)
 	}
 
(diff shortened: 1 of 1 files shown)


twpayne:
CI is currently failing due to https://github.com/mvdan/gofumpt/issues/20.

PR #339, opened 2019-06-11
chezmoi is now available in official Arch repository

daurnimator:


twpayne:
That's great news, thanks!

PR #343, opened 2019-06-12
Install completions

### LINKED ISSUE #341: wish: install completions so they work out of the box
As part of the packaging process, write the bash completions to:

    /etc/bash_completion.d/chezmoi

And Fish completions to:

   /etc/fish/completions/chezmoi

(I'm not certain about the correct zsh target, but I presume its similar). 

These are sensible paths at least on Debian and Ubuntu. I'm not sure about other Linux distributions or OS X. 

Perhaps this could be added to the `.deb` file as a first step, and other release targets could be supported in later patches.

twpayne:
This should fix #341 for `bash` and `fish`, but there's still some work to do. I'll open issues for these.


twpayne:
@daurnimator Could you add shell completions to the Arch Linux package?

markstos:
Thanks!

daurnimator:
> @daurnimator Could you add shell completions to the Arch Linux package?

Yes (once a release is made): do I just need to copy them across manually? or will go install them for us?

twpayne:
The shell completions need to copied across manually. You'll find them in the `completions/` directory. The files to be installed are:

`completions/bash/chezmoi` -> `/usr/share/bash-completion/completions/chezmoi`
`completions/fish/chezmoi.fish` -> `/usr/share/fish/completions/chezmoi.fish`

...assuming standard locations. I don't know where zsh completions should be installed (#344).

I'll do a release sometime in the next week. There are a couple of release-related things I'd like to get fixed first.

PR #354, opened 2019-06-18
Fix arch install command

plaindocs:


twpayne:
Great catch, many thanks :)

PR #375, opened 2019-06-29
Make remove command prompt by default

### LINKED ISSUE #373: `remove` should have a warning
I think maybe this should have had some kind of a prompt/warning.  I'm new here, so I was thinking of chezmoi as something like apt.  I inadvertently added my entire `~/.config/nvim` directory, realized my mistake, so I removed it.

....and I'm pretty sure I just nuked my entire vim config, and it's been awhile since I backed it up (which is why I'm here).  I think since this command is destructive (as best I can tell), it would be a very good idea to add a force flag to remove without prompt and make the default behavior require confirmation that the command is destructive.

This is my mistake, but I won't be the last one to make it most likely.

twpayne:
Fixes #373.

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -22,6 +22,9 @@ func TestExercise(t *testing.T) {
 		DestDir:   "/home/user",
 		Umask:     022,
 		Verbose:   true,
+		remove: removeCmdConfig{
+			force: true,
+		},
 	}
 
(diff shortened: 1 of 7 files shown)


jthn:
Thanks for the quick turnaround on this! I’ve stuck with the project and I am in love. Great tool and wonderful implementation.

Issue #23, opened 2018-11-26
Improve Windows support


twpayne:
`chezmoi` is currently targeted at macOS/Linux, but there is no reason why it shouldn't also run on Windows. Contributions improving Windows support are very welcome.

twpayne:
#43 adds CI on Windows, but it is not yet passing due to path issues.

roelofvkr:
I looked into it a little bit today and the issue lies in using `syscall.Stat_t` [here](https://github.com/twpayne/chezmoi/blob/2cd7d3c925de5f26545cf864db31f288f5374b9f/lib/chezmoi/fsmutator.go#L40). This is a direct linux systemcall and is unavailable on windows. 

From what I can find the best way to get an ID of a volume on windows would be using [GetVolumeInformationW](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getvolumeinformationw). Which can return a uint32 with the serial number the operating system generates when the volume is formatted.

Another option would be to use the drive letter (or in the case of network shares, the ip address and the share name), but that would require a change of `FSMutator.devCache` and `FSMutator.tempDirCache` to use strings instead of uints on windows.

zb140:
I've managed to get a reasonably usable Windows build going (code here: https://github.com/zb140/chezmoi/tree/windows).  I didn't see @roelofvkr 's  comment before I started down this path, so my fork isn't based on his, but we came up with very similar solutions using `GetVolumeInformation`.  I also implemented a custom `FSMutator.Chmod` function for Windows using ACLs to change permissions, and fixed `Config.exec` to work on Windows.

With these changes (plus a total hack in go-shell to return a value on Windows), all of the commands in the Quickstart guide work as expected, plus  a few others.  `archive` and `upgrade` don't work yet, and I haven't tested `chattr`, `import`, `remove`, `update`, or any of the secret or script stuff.  Everything else seems to work (though my experience with chezmoi is very minimal, so I'm not sure I'd notice if things were subtly broken).

It will take quite a bit of cleanup to be in a mergeable state (apart from anything else, I think it breaks all of the rules in the Contributing doc except the one about commit message formatting 😉 ) but I wanted to go ahead and share what I have.  There's also a couple of total hacks in there, plus this is the first or maybe second time I've ever worked on a project in Go, so I'm guessing my code will probably drive a legit Go programmer crazy 😄 But I'm super excited to have something working, and I'm sure everything else can be fixed.

[...]

PR #381, opened 2019-07-02
Improve Windows support.

zb140:
* Fix handling of paths and permissions to work on Windows
* Make sure all tests are passing


lib/chezmoi/fsmutator_windows.go
```diff
@@ -0,0 +1,148 @@
...
+	acl "github.com/hectane/go-acl"
+
+	"github.com/google/renameio"
+	"golang.org/x/sys/windows"
+
+	vfs "github.com/twpayne/go-vfs"
+	"github.com/twpayne/go-vfs/vfst"
+)
+
+func (a *FSMutator) getCorrectedPath(file string) string {
```
zb140: I don't know if there's a better way to deal with the fact that during the tests a path will get passed in that needs to have the test sandbox dir prepended to it.  This feels like a hack but it's probably better than pushing IsPrivate up into `go-vfs`.

twpayne:
This is really impressive work @zb140 :)

I'll only have time to review it fully at the weekend, but there a few changes that I would like to see:

* `IsPrivate` should not be a method on `Mutator`. Instead, I think it should be a top-level function in the `lib/chezmoi` package.

* I really don't like `if runtime.GOOS == ` statements. I'd rather factor out such code into separate POSIX and Windows source files and use build tags to control which gets built.

cmd/config.go
```diff
@@ -369,8 +364,9 @@ func getDefaultData(fs vfs.FS) (map[string]interface{}, error) {
 	group, err := user.LookupGroupId(currentUser.Gid)
 	if err == nil {
 		data["group"] = group.Name
-	} else if cgoEnabled {
-		// Only return an error if CGO is enabled.
+	} else if cgoEnabled && runtime.GOOS != "windows" {
```
zb140: I've removed nearly all of the uses of `runtime.GOOS` that I added.  I left this one in because I really think it makes more sense to just add this boolean check rather than try to refactor this into separate files.

zb140:
Thanks!  It's quite a big change, so I definitely understand it'll take a while to review.

In addition to removing the usages of `runtime.GOOS`, I had a go at making `IsPrivate` a top-level function.  Unfortunately that turned out to be a lot more complicated, again because of `TestFS`.  During the tests, the paths passed into `IsPrivate` do not include the test sandbox, but I need a full absolute path to pass into `acl.GetEffectiveAccessMode`.  I just don't see any way to get that path without some kind of involvement from `FSMutator`.

Next I revisited the idea discussed in #23 to override `Mutator.Stat` and return a Windows-specific `FileInfo` that understands ACLs.  So far I haven't been able to get that going either; I managed to get it to build but for some reason I haven't figured out yet, it's still not using my `Mode()` override.

I'm not sure right now how best to proceed.  If you have any thoughts or suggestions, I'd love to hear them.

lib/chezmoi/fsmutator_windows.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return false
+	}
+
+	return (uint32(mode) & 0007) == 0
+}
+
+// WriteFile implements Mutator.WriteFile.
+func (a *FSMutator) WriteFile(name string, data []byte, perm os.FileMode, currData []byte) error {
+	// Special case: if writing to the real filesystem, use github.com/google/renameio
```
twpayne: After a private chat with @stapelberg last week (author of `github.com/google/renameio`) I understood that atomic file replacement is not possible on Windows, so we can remove the use of `renameio` here. I've [asked for confirmation](https://github.com/google/renameio/issues/1#issuecomment-508942124).

lib/chezmoi/fsmutator_windows.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func getVolumeSerialNumber(Path string) (uint, error) {
+	fp, err := filepath.Abs(Path)
+	if err != nil {
+		return 0, err
+	}
+
+	// Input rootpath
+	RootPathName := filepath.VolumeName(fp) + "\\"
```
twpayne: nit: use `lowercaseSnakeCase` for local variables in Go (they are not exported).

lib/chezmoi/script.go
```diff
@@ -121,10 +121,14 @@ func (s *Script) Apply(fs vfs.FS, mutator Mutator, applyOptions *ApplyOptions) e
 	}
 
 	// Write the temporary script file.
-	f, err := ioutil.TempFile("", filepath.Base(s.targetName))
+	pathBase := filepath.Base(s.targetName)
+	pathBase = strings.Replace(pathBase, "#", "*", 1)
```
twpayne: Could you add a comment explaining why this replacement is necessary?

lib/chezmoi/fsmutator_windows.go
```diff
@@ -0,0 +1,148 @@
...
+
+	vfs "github.com/twpayne/go-vfs"
+	"github.com/twpayne/go-vfs/vfst"
+)
+
+func (a *FSMutator) getCorrectedPath(file string) string {
+	var pathfs *vfs.PathFS
+
+	// if it just is a pathfs, cool
+	maybe_pathfs, ok := a.FS.(*vfs.PathFS)
```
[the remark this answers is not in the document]
zb140: Now that `RawPath` exists in `go-vfs`, this code ended up being unnecessary.

lib/chezmoi/fsmutator_windows.go
```diff
@@ -0,0 +1,148 @@
...
+		if err == nil {
+			file = joined
+		}
+	}
+
+	return file
+}
+
+// Use the same default value as Linux (as of kernel 4.19)
+const MAXSYMLINKS = 40
```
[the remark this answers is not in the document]
zb140: Done. (note: this is in `private_windows.go` now)

lib/chezmoi/fsmutator_windows.go
zb140: I see.  I've left it in for now but once he confirms I'll go ahead and remove it.

lib/chezmoi/fsmutator_windows.go
zb140: Done.

lib/chezmoi/script.go
zb140: Turns out I massively overcomplicated this, which I realized once I started explaining the replacement.  My only defense is that some of this code was written pretty late at night :wink:

(For the record, the point of doing this was to preserve the file extension, which Windows uses to decide if it knows how to execute a file or not. Because of that, `TempFile`'s randomness couldn't just go on the end of the name, which is the default, so I needed a `*` to indicate where it should go.  But `*` is not a legal character in a Windows filename, thus the indirection.  Of course, the much simpler solution is... just put it on the front of whatever the real filename is.)

zb140:
I think I've addressed everything you mentioned (except removing `renamio`, for now).  Thanks a lot for the `RawPath` addition -- it really made things easier.  I've also included a bonus fix for prompting; I ran into this bug this evening while migrating some files from my old repo.

twpayne:
Brilliant! Thank you so much! I've merged this, will do a quick round of minor clean-ups, and tag a new version.

PR #389, opened 2019-07-07
Fix recursive add

zb140:
Fix Windows `IsPrivate` to check explicit permissions instead of effective permissions.  Effective permissions include values inherited from parent directories, which had the effect of making lots of things "private" that shouldn't have been.  Neither mode has exactly the same semantics as POSIX-style platforms, but explicit permissions should be a closer match to user intentions (ie, things are not private by default, and a user must take action to make them private).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,7 +1,6 @@
 linters:
   enable-all: true
   disable:
-    - goconst
     - gocyclo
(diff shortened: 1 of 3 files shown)


twpayne:
Thank you again @zb140 :)

Issue #392, opened 2019-07-08
On Windows only, some files written as public are detected as private


twpayne:
AppVeyor Windows builds, with *exactly the same code* sometimes pass, sometimes fail. For example:

* [Build 39](https://ci.appveyor.com/project/twpayne/chezmoi/builds/25833279), following a doc change (no change to the code), passes just fine.
* [Build 40](https://ci.appveyor.com/project/twpayne/chezmoi/builds/25833284) also passes just fine.
* [Build 41](https://ci.appveyor.com/project/twpayne/chezmoi/builds/25833560), which is exactly the same change merged into `master`, fails.

The PR is based on the latest `master` and only includes documentation changes, so the build result should be the same in all cases.

Ideas as to why this is welcome.

cc @zb140, as the Windows expert :)

twpayne:
Observations:

* All tests pass just fine on Linux/macOS.
* chezmoi uses a random temporary directory for each filesystem-related test (see https://godoc.org/github.com/twpayne/go-vfs/vfst#NewTestFS), and cleaning up afterwards.

Initial theories:

* AppVeyor is re-using VMs and random number seeds too aggressively, which seems unlikely.

It's more likely that the implicit assumptions in chezmoi's test suite are not valid on Windows, but I've no idea where the problem is. Input appreciated.

zb140:
I'll be happy to take a look, but I'm traveling this week. I'm sure I'll have time to look at it at some point, but I'll probably be a little slower to get things done than I have been in the past.

twpayne:
Of course, no rush. I know you're volunteering your time. Thank you so much for your contributions so far!

twpayne:
Here is an another example of two builds of the exact same code resulting in different outcomes on AppVeyor:

* [Build 49](https://ci.appveyor.com/project/twpayne/chezmoi/builds/26539090) (fails)
* [Build 50](https://ci.appveyor.com/project/twpayne/chezmoi/builds/26539097) (passes)

AppVeyor looks unreliable. I'll remove support for it. Maybe we can use Travis for Windows CI instead.

twpayne:
Let's try an alternative Windows CI instead, maybe Travis.

twpayne:
It seems that this flakiness is not limited to AppVeyor. In this build, something that build just fine on Windows before failed on Windows: https://github.com/twpayne/chezmoi/runs/274062037

twpayne:
[...]

PR #394, opened 2019-07-14
Fix #393 Disable the default template auto generating

### LINKED ISSUE #393: Don't auto generate the template by default when adding files as templates
The current behavior of "adding files as templates" is quite annoying as it always modify the template although I just want to add it as it is. Should we disable the template auto generating by default and let users add a flag such as `--template-auto-generate` for it when they intend to do so?

15cm:
Fix #393. Disable the default template auto generating, which can be enable by setting the
`--template-auto-generate` flag added by this commit.


twpayne:
Thanks very much for the PR :)

Please can you do the following:

* Rename the flag to `--autotemplate`.
* Rename the field in `struct AddOptions` to `AutoTemplate`.
* Add the short flag `-a`.
* Run `make completions` to generate the shell completions and commit them.
* Update the autotemplate test that currently fails (set `AutoTemplate: true` in `AddOptions`.
* Add a test the ensures that autotemplate is off by default.

Let me know if more information is needed. Equally, if you don't have time to work on this, please tell me and I can build on your work here.

15cm:
No probelm. Will take your suggestions and work on it these days.

15cm:
I have got your suggestions done but met an issue with CI. Could you please take a look for at why the CI failed to find some files when testing in the vfs. Actually the CI seems to be fail on your last commit 8 days ago: https://ci.appveyor.com/project/twpayne/chezmoi/builds/25833560

twpayne:
Thanks very much for making these changes. I think the Windows builds are flaky at the moment (#392) - the builds started failing after a documentation change. It is possible that I broke something somewhere else. I don't have a Windows machine to test on, so I can only use the AppVeyor CI to debug problems, which has a very slow feedback loop.

I won't have time to work on this for a couple of weeks due to other commitments, but I do plan to fix the Windows CI and get this PR merged.

twpayne:
Following #403, could you rebase this PR on the latest `master`?

15cm:
Done. Thanks for your help.

twpayne:
Thanks for updating the PR. It looks like you merged `master` rather than rebasing, and consequently the history is a little mixed up with multiple duplicate commits. I've squashed and rebased everything in #404.

Thanks again for the great contribution!

PR #395, opened 2019-07-19
Updated install script to be in line with shlib/uname_os.sh

tralston:
(see https://github.com/client9/shlib/commit/56b1c04bf46d992867a61ca70c03940d9047b4c2)

And updated the supported platforms to accept windows/amd64.


twpayne:
Thanks very much for this. The install script is generated by [godownloader](https://github.com/goreleaser/godownloader). Would you be able to submit a PR there instead? That way, all projects using godownloader will benefit from the fix.

twpayne:
I opened https://github.com/goreleaser/godownloader/pull/127 to fix this upstream.

twpayne:
Thanks very much for this PR. `godownloader` is slow to update so I created and merged #415 to include your changes (commit credit to you). Thanks again for your contribution :)

Issue #413, opened 2019-09-04
Feature request: autocommit and autopush


lewisacidic:
Thanks for this really cool project. Just thought I would stop by with a feature request.

It would be really useful to have a way of automatically committing changes within the source repo (with an appropriate generated commit message), and then pushing them to the remote - I know I'll forget to do this sometimes otherwise!  Perhaps this could be added as a new flag for commands that make changes to the source directory, such as `add` and `edit`?

twpayne:
Thanks for the kind words :)

This is a really nice idea.

twpayne:
Fixed in #432. I've tagged a new version, 1.7.2, that includes this functionality which should be on the [releases page](https://github.com/twpayne/chezmoi/releases) in about 15 minutes.

Documentation: https://github.com/twpayne/chezmoi/blob/master/docs/HOWTO.md#automatically-commit-and-push-changes-to-your-repo

Please test this and open issues for any problems you find.

cc @lolilijve

angristan:
I created a `.chezmoi.toml` in my chezmoi directory, and committed it. Afterwards, whenever I make a change, nothing is automatically committed or pushed... Am I missing something?

Contents of `/Users/stanislas/.local/share/chezmoi/.chezmoi.toml``:

```toml
[sourceVCS]
    autoCommit = true
    autoPush = true
```

twpayne:
Please [open a new bug report](https://github.com/twpayne/chezmoi/issues/new?assignees=&template=03_bug_report.md&title=) so I can gather the information required to debug your problem.

angristan:
(Sorry I deleted my comment right away, I didn't read the documentation enough!)

twpayne:
Great! Happy it's working for you :)

PR #432, opened 2019-09-24
Add autocommit and autopush

twpayne:
This is a draft PR working towards fixing #413.

cc @lewisacidic


lolilijve:
Looking forward, I think this feature is very useful

PR #434, opened 2019-09-24
Only open BoltDB when required

### LINKED ISSUE #433: Can't run chezmoi source-path within script
###### I did all the troubleshooting

I'm struggling with running chezmoi within chezmoi. I need to retrieve the chezmoi source path within my script. I used to retrieve it by running `sourcedir=$(chezmoi source-path)` and it worked until recently. Now that command just hangs forever while running `chezmoi apply`. If I run `chezmoi source-path` in a second terminal, it hangs until I terminate `chezmoi apply`.
Were there any changes that now only allow one chezmoi instance to execute in parallel?

It might also be an issue with my system, but I'm out of ideas.

#### Command that failed

```
chezmoi -v apply
```

#### Output of command with `--verbose`
It outputs my script, which contains `sourcedir=$(chezmoi source-path)`, then hangs.

#### Output of `chezmoi doctor`
Only missing a few password manager executables.

#### Output of `chezmoi --version`
```
chezmoi version 1.6.5, commit 3733485a2e018dfe174423c5c58edd6d03f0e3c6, built at 2019-09-20T12:17:47Z
```

schrej:
The current implementation blocks chezmoi from running in parallel, e.g. within chezmoi scripts. This was caused by BoltDB blocking when trying to open the same database in read-mode twice.

This PR changes the behavior to only open the db when it is required. It also opens it in read-mode only when doing a dry-run.

fixes #433

### DIFF
--- a/lib/chezmoi/boltpersistentstate.go
+++ b/lib/chezmoi/boltpersistentstate.go
@@ -16,7 +16,7 @@ type BoltPersistentState struct {
 	db   *bolt.DB
 }
 
-// NewBoltPersistentState returns a new BoltPersistentState.
+// NewBoltPersistentState returns a new, unopened BoltPersistentState.
 func NewBoltPersistentState(fs vfs.FS, path string) (*BoltPersistentState, error) {
 	b := &BoltPersistentState{
 		fs:   fs,
@@ -24,19 +24,18 @@ func NewBoltPersistentState(fs vfs.FS, path string) (*BoltPersistentState, error
 		perm: 0600,
 	}
 	_, err := fs.Stat(b.path)
-	switch {
-	case err == nil:
-		if err := b.openDB(); err != nil {
-			return nil, err
-		}
-	case os.IsNotExist(err):
-	default:
+	if err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
 	return b, nil
 }
 
-// Close closes b.
+// Open the connection to b.
+func (b *BoltPersistentState) Open(write bool) error {
+	return b.openDB(write)
+}
+
+// Close the connection to b.
 func (b *BoltPersistentState) Close() error {
 	if b.db == nil {
 		return nil
@@ -87,9 +86,10 @@ func (b *BoltPersistentState) Get(bucket, key []byte) ([]byte, error) {
 // it does not already exist.
 func (b *BoltPersistentState) Set(bucket, key, value []byte) error {
 	if b.db == nil {
-		if err := b.openDB(); err != nil {
+		if err := b.openDB(true); err != nil {
 			return err
 		}
+		defer b.Close()
 	}
 	return b.db.Update(func(tx *bolt.Tx) error {
 		b, err := tx.CreateBucketIfNotExists(bucket)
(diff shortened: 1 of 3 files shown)


lib/chezmoi/boltpersistentstate.go
```diff
@@ -87,9 +87,10 @@ func (b *BoltPersistentState) Get(bucket, key []byte) ([]byte, error) {
 // it does not already exist.
 func (b *BoltPersistentState) Set(bucket, key, value []byte) error {
 	if b.db == nil {
-		if err := b.openDB(); err != nil {
+		if err := b.openDB(true); err != nil {
 			return err
 		}
+		defer b.Close()
```
twpayne: I don't think this is correct: it causes the database to be opened and closed on every `Set`.

lib/chezmoi/boltpersistentstate.go
```diff
@@ -16,27 +16,27 @@ type BoltPersistentState struct {
 	db   *bolt.DB
 }
 
-// NewBoltPersistentState returns a new BoltPersistentState.
+// NewBoltPersistentState returns a new, unopened BoltPersistentState.
+// If you want to access a database b, open it with b.Open()
```
twpayne: Please drop the second line (`If you want...`): the statement is already implied by the inclusion of `unopened` in the first line.

lib/chezmoi/boltpersistentstate.go
```diff
@@ -16,27 +16,27 @@ type BoltPersistentState struct {
...
 	}
 	_, err := fs.Stat(b.path)
-	switch {
-	case err == nil:
-		if err := b.openDB(); err != nil {
-			return nil, err
-		}
-	case os.IsNotExist(err):
-	default:
+	if os.IsNotExist(err) {
```
twpayne: This should be

```go
if err != nil && !os.IsNotExist(err) {
    // ...
}
```

lib/chezmoi/boltpersistentstate.go
schrej: This just mimics the previous behavior. If the database gets opened automatically, it should also get closed automatically imo. If it's open already, that defer statement won't get called.

lib/chezmoi/boltpersistentstate.go
```diff
@@ -16,27 +16,26 @@ type BoltPersistentState struct {
...
 	}
 	_, err := fs.Stat(b.path)
-	switch {
-	case err == nil:
-		if err := b.openDB(); err != nil {
-			return nil, err
-		}
-	case os.IsNotExist(err):
-	default:
+	if err != nil && os.IsNotExist(err) {
```
twpayne: I think my previous comment was incorrect. This should be:

```go
if err != nil && !os.IsNotExist(err) {
```

The logic should be: "either the stat should succeed or the file doesn't exist, anything else is an error".

lib/chezmoi/boltpersistentstate.go
twpayne: The original intent is that the database should be opened lazily, i.e. only when accessed, but once opened it should stay open. I'll revisit this logic later.

lib/chezmoi/boltpersistentstate.go
schrej: That makes more sense, yes.

twpayne:
Thanks for your work here. I took a different approach and fixed some non-hermetic tests in #441.

PR #460, opened 2019-10-17
Remove non-git VCS support

### LINKED ISSUE #459: Remove  non-git VCS support
chezmoi currently includes support for plugging in your own VCS through the `sourceVCS.command` configuration value. However, this is hard to maintain without expert support for non-`git` version control systems as chezmoi moves towards closer integration with the underlying VCS, for example as required to implement #413.

This issue is a concrete proposal to remove non-git support from chezmoi. Doing so would simplify the implementation (e.g remove the [`VCS interface`](https://github.com/twpayne/chezmoi/blob/master/cmd/vcs.go)) and allow chezmoi to provide a consistent set of features to all users.

cc @mgeisler who might have a different perspective.

twpayne:
Fixes #459.


twpayne:
Closing this for now. Removing non-git VCS support might break some users, so should be done gradually with warnings before eventual deprecation.

PR #472, opened 2019-11-03
fix broken link

sseneca:
I was checking out the website and noticed the second link wasn't working. It's because it links to `what-are-chezmois-key-features`, [missing the hyphen](http://www.chezmoi.io/#what-are-chezmoi-s-key-features) which is meant to be in place of the quotation mark in "chezmoi's". There is the same problem elsewhere where there is certain punctuation e.g. [this](http://chezmoi.io/docs/faq/#if-theres-a-mechanism-in-place-for-the-above-is-there-also-a-way-to-tell-chezmoi-to-ignore-specific-files-or-groups-of-files-eg-by-directory-name-or-by-glob) link in the FAQs is [missing a hyphen as well](http://chezmoi.io/docs/faq/#if-there-s-a-mechanism-in-place-for-the-above-is-there-also-a-way-to-tell-chezmoi-to-ignore-specific-files-or-groups-of-files-e-g-by-directory-name-or-by-glob) in place of the quotation mark and full stop (afaict not on GitHub hence why it's not fixed in this commit)

### DIFF
--- a/chezmoi.io/content/_index.md
+++ b/chezmoi.io/content/_index.md
@@ -8,7 +8,7 @@ type: docs
 Manage your dotfiles across multiple machines, securely.
 
 * [What does chezmoi do and why should I use it?](#what-does-chezmoi-do-and-why-should-i-use-it)
-* [What are chezmoi's key features?](#what-are-chezmois-key-features)
+* [What are chezmoi's key features?](#what-are-chezmoi-s-key-features)
 * [I already have a system to manage my dotfiles, why should I use chezmoi?](#i-already-have-a-system-to-manage-my-dotfiles-why-should-i-use-chezmoi)
 * [How do I start with chezmoi?](#how-do-i-start-with-chezmoi)
 * [License](#license)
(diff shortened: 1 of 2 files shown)


twpayne:
Many thanks for the PR!

The tables of contents are generated automatically by the [Markdown All in One plugin](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) for VSCode and are not maintained manually.

It looks like the deeper issue here is that I haven't configured the plugin correctly, and there might be some tweaking required to get the automatically-generated table of contents working with both GitHub and the static website generator, [Hugo](https://gohugo.io/).

So, I'd like to hold off merging this PR until the deeper issue is fixed. It is likely that other documents are affected too and I'd like an automated solution that "just works".

sseneca:
> there might be some tweaking required to get the automatically-generated table of contents working with both GitHub and the static website generator, Hugo.

Yeah, I noticed the links seem to work with GitHub's flavour or Markdown but not Hugo's, which is annoying

> So, I'd like to hold off merging this PR until the deeper issue is fixed. It is likely that other documents are affected too and I'd like an automated solution that "just works".

Agreed! The more that's automated the better

twpayne:
Thanks @sseneca for your work here. In the end I fixed the problem by dropping the GitHub Markdown tables of contents from chezmoi.io and instead relying only on the tables of contents generated by Hugo.

PR #503, opened 2019-12-02
Add onepasswordDocument template function

### LINKED ISSUE #501: Wish: Add file fetching support from password managers
Hi,

For background, the way I deal with SSH key pairs is by keeping them in 1Password, mainly because in 1Password adding data with newlines is not possible in regular item fields, so
I'd like to have a way to retrieve files from 1Password without parsing them (`op get document id` instead of `op get item id`).

I've done some prototyping in code as to how this could work so I came up with a `file` template function which can be parameterized with the password manager you use, since I noticed at least Bitwarden supports this, so in the end it looks like `file onepassword id` in the templates.

I'm wondering if you have plans for this already and if you don't if this is something you'd like to integrate.

BTW, I tried adding this `file` function as a parameter of the already existing `onepassword` function but noticed it might have a significant impact to the codebase (might be that my Go skills are non-existant) if it's even possible.

twpayne:
Fixes #501.

@rcorrear are you able to test this? I'm not a 1Password user, but can do some tests late this evening otherwise.

### DIFF
--- a/cmd/secretonepassword.go
+++ b/cmd/secretonepassword.go
@@ -21,11 +21,15 @@ type onepasswordCmdConfig struct {
 	Command string
 }
 
-var onepasswordCache = make(map[string]interface{})
+var (
+	onepasswordCache         = make(map[string]interface{})
+	onepasswordDocumentCache = make(map[string]string)
+)
 
 func init() {
 	config.Onepassword.Command = "op"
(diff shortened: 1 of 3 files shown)


twpayne:
I was able to test and it needed a small fix, but it seems be working correctly now. I'll merge once CI passes and tag a new version.

rcorrear:
I also just tested, it works well. Thanks a lot for the quick fix!

PR #508, opened 2019-12-09
Fixed a type in completion example

ptxmac:
Just a quick fix, `fish` shell completions files must end in `.fish`


twpayne:
Thank you for the fix :)

The build is failing because chezmoi extracts the help for commands from `REFERENCE.md` and this needs an extra step to generate:

    make generate

This will generate `cmds/helps.gen.go`. The check in CI that is failing is the check that this file is up-to-date.

If you could run `make generate` and amend the commit (e.g. run `make generate && git add cmd/helps.gen.go && git commit --amend && git push --force`) that would be fantastic.

twpayne:
Thank you for your contribution :)

PR #528, opened 2019-12-22
Added kernel information to data

ptxmac:
The main purpose of this is to have a way to detect WSL (windows subsystem for linux).

From what I can gather, the best way is currently to look in `/proc/sys/kernel/osrelease` and look for "microsoft" in the kernel name: https://github.com/microsoft/WSL/issues/423

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -320,6 +320,12 @@ func (c *Config) getDefaultData() (map[string]interface{}, error) {
 		return nil, err
 	}
 
+	kernelInfo, err := getKernelInfo(c.fs)
+	if err == nil && kernelInfo != nil {
+		data["kernel"] = kernelInfo
+	} else if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
 	return data, nil
 }
 
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -713,17 +713,18 @@ chezmoi's configuration file.
 
 chezmoi provides the following automatically populated variables:
 
-| Variable                | Value                                                                                                                  |
-| ----------------------- | ---------------------------------------------------------------------------------------------------------------------- |
-| `.chezmoi.arch`         | Architecture, e.g. `amd64`, `arm`, etc. as returned by [runtime.GOARCH](https://godoc.org/runtime#pkg-constants).      |
-| `.chezmoi.fullHostname` | The full hostname of the machine chezmoi is running on.                                                                |
-| `.chezmoi.group`        | The group of the user running chezmoi.                                                                                 |
-| `.chezmoi.homedir`      | The home directory of the user running chezmoi.                                                                        |
-| `.chezmoi.hostname`     | The hostname of the machine chezmoi is running on, up to the first `.`.                                                |
-| `.chezmoi.os`           | Operating system, e.g. `darwin`, `linux`, etc. as returned by [runtime.GOOS](https://godoc.org/runtime#pkg-constants). |
-| `.chezmoi.osRelease`    | The information from `/etc/os-release`, Linux only, run `chezmoi data` to see its output.                              |
-| `.chezmoi.sourceDir`    | The source directory.                                                                                                  |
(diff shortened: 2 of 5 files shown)


twpayne:
Thank you for this very nice contribution!

Please can you:
* Use build flags so this is only built when `GOOS=linux` (you can create a `getKernelInfo` function that returns nothing on other OSs, only set `data["kernel"]` if the return value is non-`nil`).
* Add documentation of the variable to `docs/REFERENCE.md`.

ptxmac:
Sure, I'll get on that.

I based the implementation on `osrelease.go` since that's also a linux specific thing, but that isn't using build tags. Are there any reason for that?

twpayne:
Thank you. `osrelease.go` not doing this is an oversight on my part. I'll fix it.

cmd/config.go
```diff
@@ -320,6 +320,12 @@ func (c *Config) getDefaultData() (map[string]interface{}, error) {
 		return nil, err
 	}
 
+	kernelInfo, err := getKernelInfo(c.fs)
+	if err == nil {
+		data["kernel"] = kernelInfo
```
twpayne: Please only set `data["kernel"]` if `kernelInfo != nil`.

cmd/kernelinfo.go
```diff
@@ -0,0 +1,45 @@
+// +build linux
```
twpayne: Please rename this file to `kernelinfo_linux.go` and remove the build tag.

cmd/kernelinfo_test.go
```diff
@@ -0,0 +1,29 @@
+// +build linux
```
twpayne: Please rename this file to `kernelinfo_linux_test.go` and remove the build tag.

cmd/nokernelinfo.go
```diff
@@ -0,0 +1,9 @@
+// +build !linux
```
twpayne: Please rename this file to `kernelinfo_notlinux.go` and keep the build tag.

docs/REFERENCE.md
```diff
@@ -713,17 +713,18 @@ chezmoi's configuration file.
...
+| ----------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| `.chezmoi.arch`         | Architecture, e.g. `amd64`, `arm`, etc. as returned by [runtime.GOARCH](https://godoc.org/runtime#pkg-constants).             |
+| `.chezmoi.fullHostname` | The full hostname of the machine chezmoi is running on.                                                                       |
+| `.chezmoi.group`        | The group of the user running chezmoi.                                                                                        |
+| `.chezmoi.homedir`      | The home directory of the user running chezmoi.                                                                               |
+| `.chezmoi.hostname`     | The hostname of the machine chezmoi is running on, up to the first `.`.                                                       |
+| `.chezmoi.kernel`       | Contains information from `/proc/sys/kernel`. Linux only, useful for detecting specific kernels (I.e. microsfot's WSL kernel) |
```
twpayne: Please fix the typos and capitalization and add a closing full stop: `(i.e. Microsoft's WSL kernel).`

cmd/config.go
ptxmac: I guess I could also just return file not found from the no-op implementation

twpayne: Return a file not found error would have the same effect, but I think returning `nil, nil` more clearly signals the intent of the code.

cmd/nokernelinfo.go
ptxmac: Man, go's build tags are getting messy :D - some can be expressed by file names, but only a subset...

cmd/config.go
twpayne: This isn't quite right: if `err == nil` and `kernelInfo == nil` then the code ends up at line 327.

ptxmac: Yeah, I was a bit too quick with that fix

twpayne:
Thank you very much for the great contribution!

PR #555, opened 2020-01-10
Don't attempt to use ANSI colors or terminal width on Windows

### LINKED ISSUE #554: Diff does not show colors properly con Windows
Getting started on Windows, looks like the colors does not work properly.
(Or maybe the diff format is just too weird for me).

image

twpayne:
Fixes #554.


twpayne:
Actually, this is not the correct approach.

PR #560, opened 2020-01-11
Merge pull request #560 from twpayne/source-alias

twpayne:
As [suggested by](https://github.com/twpayne/chezmoi/issues/554#issuecomment-573196709) @omry.

### DIFF
--- a/cmd/git.go
+++ b/cmd/git.go
@@ -1,0 +1,27 @@
+package cmd
+
+import (
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+)
+
+var gitCmd = &cobra.Command{
+	Use:     "git [args...]",
+	Short:   "Run git in the source directory",
+	Long:    mustGetLongHelp("git"),
+	Example: getExample("git"),
+	RunE:    config.runGitCmd,
+}
+
+func init() {
+	rootCmd.AddCommand(gitCmd)
+}
+
+func (c *Config) runGitCmd(cmd *cobra.Command, args []string) error {
+	name := "git"
+	if trimExecutableSuffix(filepath.Base(c.SourceVCS.Command)) == "git" {
+		name = c.SourceVCS.Command
+	}
+	return c.run(c.SourceDir, name, args...)
+}
--- a/cmd/helps.gen.go
+++ b/cmd/helps.gen.go
@@ -228,6 +228,14 @@ var helps = map[string]help{
 			"Description:\n" +
 			"  Print the help associated with *command*.",
 	},
+	"hg": {
+		long: "" +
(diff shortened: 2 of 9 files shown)


omry:
Cool!
A nice touch can be that you dynamically show add the correct command based on the type of the source directory.

twpayne:
> A nice touch can be that you dynamically show add the correct command based on the type of the source directory.

I agree that this is possible but I like the predictability of always having both commands available.

omry:
Can you use the hg command if you have a git repo?

twpayne:
Yes. Why not? Having the output of `chezmoi help` vary depending on which vcs you're using would be quite confusing and I don't see any benefit to omitting `git` or `hg` or both from that output. Maybe you're converting from hg to git and have both present in your source directory. In practice, people will know exactly which vcs they're using and will use only that command.

omry:
Alright.
As a first time user, I was not aware of the repository (it was created as git automatically, I think).
I am not sure how it's possible to have two chezmoi repositories at the same time.

twpayne:
chezmoi tries to be vcs agnostic (which is a bit silly because almost everybody uses git these days). The creation of the git repo is mentioned in the [quickstart guide](https://github.com/twpayne/chezmoi/blob/master/docs/QUICKSTART.md#start-using-chezmoi-on-your-current-machine). Most of chezmoi only operates on the source directory, and is ignorant of how the files there are made to appear or change. Commands like `chezmoi update`, `chezmoi source`, `chezmoi hg`, `chezmoi git`, and the autocommit/autopush feature do interact with the underlying vcs, but there are not many of these commands (but they are important commands, of course).

omry:
Ah, yes.
so `chezmoi init https://github.com/username/dotfiles.git` would know to create a git repo?
How would one go about creating an hg repo?

I see that this command (in the example at least) does not take a target directory which suggests there is only one source repo used at any given time.
This is why I thought it might be good to only show the appropriate git/hg option to the user.
This is really not important.

twpayne:
> Ah, yes.
> so chezmoi init https://github.com/username/dotfiles.git would know to create a
> git repo?

Technically, this clones, rather than creates, a git repo, but yes.

> How would one go about creating an hg repo?

This is a missing feature in chezmoi. Right now there's no way to clone/create a hg repo with `chezmoi init`. At the same time, there has been no demand for this.

Thanks again for raising this - you are identifying completely valid gaps in chezmoi's functionality. All of this makes chezmoi better, so thank you :)

omry:
Happy to help!
As I start relying on Chezmoi I may have additional insight :)
For now, my usage is really minimal (just one file being synchronized).

PR #563, opened 2020-01-13
Merge pull request #563 from ndt/patch-1

ndt:
I could not install chezmoi 1.7.11 with `curl -sfL https://git.io/chezmoi | sh` due to a changed file name pattern of the checksum file.


twpayne:
Thanks very much for reporting this and the fix. There are actually some more problems related to how `install.sh` is generated, I'll open another PR to fix them.

PR #598, opened 2020-02-14
typo

pvinis:


twpayne:
Thanks for the fix. The CI failed because `go generate` needs to be run after the documentation is updated. I've done this in #602, maintaining you as author of the commit.

PR #635, opened 2020-03-22
Improve windows support by removing carriage return from prompts

ptxmac:
Windows uses `\r\n` for newlines, so when reading input it needs extra attention to avoid creating data with extra `\r` in the contents

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -177,5 +177,5 @@ func (c *Config) promptString(field string) string {
 	fmt.Fprintf(c.Stdout, "%s? ", field)
 	value, err := bufio.NewReader(c.Stdin).ReadString('\n')
 	panicOnError(err)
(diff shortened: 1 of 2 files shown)


twpayne:
Thank you very much for this nice contribution!

I think it makes sense to strip both leading and trailing whitespace and therefore use `strings.TrimSpace` instead of `strings.TrimRight`. Could you please update the PR and also make this change to [this line in `config.go`](https://github.com/twpayne/chezmoi/blob/master/cmd/config.go#L494)?

Many thanks!

twpayne:
I'll merge this now and make the requested changes in a follow-up PR. Thank you again for this really nice contribution :)

ptxmac:
Sorry about that, I was a bit busy and forgot to reply!

Issue #637, opened 2020-03-23
Misleading documentation: Handle configuration files which are externally modified


jsamr:
I think [the example](https://www.chezmoi.io/docs/how-to/#handle-configuration-files-which-are-externally-modified) has a typo (should replace `.` with `dot`):

```
mkdir -p $(chezmoi source-path)/.config/Code/User
echo -n "{{ .chezmoi.sourceDir }}/settings.json" > $(chezmoi source-path)/.config/Code/User/symlink_settings.json.tmpl
```
should be replaced with

```
mkdir -p $(chezmoi source-path)/dot_config/Code/User
echo -n "{{ .chezmoi.sourceDir }}/settings.json" > $(chezmoi source-path)/dot_config/Code/User/symlink_settings.json.tmpl
```

Otherwise
```
chezmoi apply -v
```
is no-op.

nawordar:
I was confused by that example too and I can confirm that it should be `dot_` not `.`

twpayne:
Thanks for reporting this @jsamr and thanks for the fix @nawordar :)

PR #638, opened 2020-03-23
Change . to dot_ in symlink template example

nawordar:
This is suggested fix for #637


twpayne:
Thanks for the contribution :)

Issue #424, opened 2019-09-12
wish: chezmoi status


markstos:
I keep expecting `chezmoi status` to exist.

Parallel to `git status`, it would work like a condensed version of `chezmoi diff` to only list files that that differ between source and target. 

Ideally, it would also tell me:

  * If there any patches to push or pull
  * When source and target files differ, which one has newer modifications.

twpayne:
This is a great idea, and, as you say, ties in very nicely with #423.

I'm not sure when I'll get time to implement this (hopefully sometime in the next few weeks), but it will definitely get done.

astorath:
It would also be nice to have `git ls-files` alternative as an option for `chezmoi status`. We can then be able to use it's output in completion scripts.

Issue #423, opened 2019-09-11
Wish: warn by default if there are differences before applying


markstos:
I enjoy `chezmoi` and use it regularly now. Thanks for the tool!

#### Description of issue

I have not adopted the use of use of `chezmoi edit` on a regular basis. Some tools, like Fish shell can end up adding or updating their own config files directly, so local changes outside of `chezmoi edit` are unavoidable. 

Instead, I edit my dot files directly and use `chezmoi diff` and then `add`, `apply` or `merge` to resolve differences.

Today I ran `chezmoi update` for the first time, not realizing beforehand that it would `apply` by default. I might have lost of some changes-- I'm not sure.

`apply` or `update` could be made safer with some kind pre-check to see if there are changes. I'm not sure which options which would best. Some warning signs there have been local changes:

 * Before a `pull`, does `chezmoi diff` show a change?
 * Has the local file changed since the most recent Git commit was created?

A more formal solution might also be possible: When `apply` updates all the files, also write out an inventory file that contains SHAs of each file written out by chezmoi.

The next time `apply` runs, check to see if the SHAs still match. If not, the file has been modified outside of `chezmoi`.

Whatever the mechanism, `chezmoi` could become "safe by default" by stopping an `apply` if any local changes are detected. Options could be presented to view or merge the differences or force the application.

twpayne:
Another great idea @markstos! I will implement it when I have time.

GabrielDuarteM:
Was about to suggest that. As @markstos said, some tools automatically add stuff to your configs, and I lost some of them due to an `apply` I made later. Checking for diffs before every `apply` would be really nice to avoid those cases.

lunik1:
I did this more often than I would like to admit :). Adding this snippet to your `.zshrc` or `.bashrc` (other shells untested) should stop chezmoi from overwriting any changes when `chezmoi apply` is run
```bash
# Don't let 'chezmoi apply' overwrite changes
chezmoi() {
  if [[ "$1" == apply ]] && [[ $(command chezmoi apply -n -v "${@:2}") ]]; then
    printf "This would overwrite changes! Aborting!\n"
  else
    command chezmoi "${@}"
  fi
}
```
A couple of limitations:
- It won't work if there are flags between `chezmoi` and `apply`
- It won't protect you from `chezmoi edit --apply` etc.

Hope it's useful! Would love to have this check built-in.

lunik1:
[...]

PR #654, opened 2020-04-04
Rework internals

twpayne:
This is a work-in-progress PR to update the internals of chezmoi to facilitate features like #423 and #424.


.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         os:
         - macos-latest
         - ubuntu-latest
-        - windows-latest
+        #- windows-latest # FIXME
```
zb140, line 15: Hey @twpayne, I just happened to come across this tonight and I wanted to let you know I'm available to help test or fix stuff on Windows if needed.

twpayne: Thanks @zb140, that's really appreciated :) I will definitely reach out to you.

twpayne:
@zb140 this isn't ready for running on Windows yet, but here's a very quick overview of what this PR hopes to achieve.

The primary goal is to enable new functionality that would be very tricky to implement with chezmoi's internal architecture in its current state. Specifically, these are:

* Implement `chezmoi status` to show the status of the destination state. This is  a three-way comparison between the current destination state, the target state, and the last destination state written by chezmoi. The idea is that if the current destination state is not equal to the last destination state written by chezmoi then the target has been modified by a third party and you can inform the user and take appropriate action. To achieve this, chezmoi will store a hash of the last destination state written in its persistent state and compare this with the equivalent hash of the current destination state.

* Provide more consistent handling of edge cases where multiple features intersect. One example of this is the interaction of `chezmoi add`, `chezmoi apply`, `.chezmoiignore`, and `.chezmoiremove`. The refactor should make it easier to detect when, say, a user adds a target that matches `.chezmoiignore` or `.chezmoiremove`. Another example, is how should encrypted scripts that are also templates be handled by `chezmoi edit`? There are many such edge cases.

chezmoi's internal architecture at the moment isn't terrible, but it does carry a fair amount of technical debt from its evolution over the past 18 months. Now chezmoi is stabilizing it's time for a clean-up.

Internally, the clean-up does a number of things:

* It makes a clear distinction between the source state, the target state, and the destination state. In v1 there is only the source state and the target state, which makes it hard to hook in the `chezmoi status` functionality.

* Representations have fewer assumptions built into them. For example, v1 packs the "private" and "executable" attributes into a single POSIX-like mode, and then has to unpack them when an individual attribute is needed. This complicates the Windows implementation as (as your work has been so invaluable in showing) "private" on Windows doesn't mean the same as "private" on POSIX. Now all of this is much cleaner and Windows support should be better.

* Internally, I'm hoping that I can use forward-slash separated paths everywhere, which will simplify a lot of logic when it comes to comparing and matching filenames. I'm not 100% sure that this will work, but I'm crossing my fingers.

* chezmoi v1 initially mutated file systems and then had running scripts and storing state bolted on to that. v2 unifies all of these into a single "system" that is mutated, which makes a lot of functionality much easier to implement. For example, modifying a tar archive "system" becomes writing files to a tar archive and executing a script on a git diff "system" becomes including the script's contents in the diff output.

At the moment, I still have a number of unanswered questions that are not reflected in the code yet, so I wouldn't recommend diving in here unless you're particularly keen or bored :)

zb140:
This sounds super exciting! Let me know how I can help.

twpayne:
For folk following along, I'm trying to maintain an up-to-date status of the `v2` branch in this file:

https://github.com/twpayne/chezmoi/blob/v2/next/docs/CHANGES.md

I still have work to do, but chezmoi v2 is getting closer and there are a lot of passing integration tests in this directory:

https://github.com/twpayne/chezmoi/tree/v2/next/testdata/scripts

zb140:
🎉 This is amazing!

twpayne:
Thanks @zb140 - I'll open an issue explaining how to beta test in the next hour or so.

PR #656, opened 2020-04-05
Include non-hermetic text functions from Sprig

### LINKED ISSUE #655: Get Chezmoi data from current environment variables

As I run chezmoi inside containerized dev environment, my control over the environment variable is complete. 
Template currently watch data values from the one defined in .config/chezmoi/chezmoi.toml. 

I would like `Chezmoi  apply` to expose a new flag to override or mix data values from the defined environement variables. By defaul it would be the value from the chezmoi config file. 

Stick to the config file.

## Additional context

Developement inside a container is eased by the VSCode remote -containers extension. 
https://code.visualstudio.com/docs/remote/containers
It would be neat to not have to create chezmoi.toml file to set template values and just rely on what is defined as env variables in the docker-compose.yml file. 
@twpayne , what do you think? I a willing to implement it once we agree on something.

twpayne:
Fixes #655.

@ihommani what do you think?

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -140,7 +140,7 @@ func newConfig(options ...configOption) *Config {
 			Command: "vimdiff",
 		},
 		maxDiffDataSize:   1 * 1024 * 1024, // 1MB
-		templateFuncs:     sprig.HermeticTxtFuncMap(),
+		templateFuncs:     sprig.TxtFuncMap(),
 		scriptStateBucket: []byte("script"),
 		Stdin:             os.Stdin,
 		Stdout:            os.Stdout,
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -19,7 +19,7 @@ import (
 func TestAutoCommitCommitMessage(t *testing.T) {
 	commitMessageText, err := getAsset(commitMessageTemplateAsset)
 	require.NoError(t, err)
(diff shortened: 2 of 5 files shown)


twpayne:
> To me the change is ok but in a product mindset, it's not as harmful as the change implies.
> What do you think?

I agree. chezmoi is a tool for power users so I think it's OK to include this functionality and rely on the use to wield it correctly.

maverick1872:
@twpayne Is there any examples of the `env` & `expandenv` functions being used? From the most basic of MWE I have been unable to get the following to function in my testing.

```shell
chezmoi execute-template '{{ env $EDITOR}}'
```

Or am I misunderstanding the usage?

halostatue:
> ```shell
> $ chezmoi execute-template '{{ env $EDITOR}}'
> chezmoi: template: arg1:1: undefined variable "$EDITOR"
> ```

There are examples on the [sprig page](http://masterminds.github.io/sprig/os.html). In all cases, you cannot simply use `{{ env $EDITOR }}` because that’s trying to read the Go template variable `$EDITOR`. The two options are `{{ env "EDITOR" }}` or `{{ expandenv "EDITOR=$EDITOR" }}`.

In the future, please open a new issue instead of commenting on a three-year-old PR.

PR #665, opened 2020-04-11
Add managed command

### LINKED ISSUE #659: Managed command

I generally like to have a quick way to access my configuration files for edits and the fact that chezmoi offers the ability to edit and apply in one go is very useful to me. Unfortunately there doesn't seem to be a command that allows listing the target filenames in order to select from them and pass them to `chezmoi edit -a`.

I see 2 options to bring this capability:

* a **managed** command that would list the target filenames managed by chezmoi
* allow the **edit** command to take source filenames (maybe using an optional switch ?)

I would definitely favor the first option as getting an accurate listing of source filenames can be tricky (there are files in the chezmoi directory that aren't actually managed configuration files). I can also see a list of managed target filenames being useful in other contexts (not only for edits).

Currently I resort to a couple bash scripts I wrote that parse the chezmoi source directory and output the matching target filenames, which I then select from using fzf and pass to `chezmoi edit -a`.

## Additional comments

I have some basic go knowledge so if one one the 2 solutions appear like a good one I would be happy to try to implement it myself and submit a pull request.

Thanks from a happy (new) user of chezmoi!

remigourdon:
Closes #659

### DIFF
--- a/cmd/managed.go
+++ b/cmd/managed.go
@@ -1,0 +1,66 @@
+package cmd
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"github.com/twpayne/chezmoi/internal/chezmoi"
+)
+
+var managedCmd = &cobra.Command{
+	Use:     "managed",
+	Args:    cobra.NoArgs,
+	Short:   "List the managed files in the destination directory",
+	Long:    mustGetLongHelp("managed"),
+	Example: getExample("managed"),
+	PreRunE: config.ensureNoError,
+	RunE:    config.runManagedCmd,
+}
+
+func init() {
+	rootCmd.AddCommand(managedCmd)
+}
+
+func recurseEntries(in []chezmoi.Entry) []string {
+	out := make([]string, 0, len(in))
+	for _, entry := range in {
+		switch v := entry.(type) {
+		case *chezmoi.Dir:
+			entries := make([]chezmoi.Entry, 0, len(v.Entries))
+			for _, entry := range v.Entries {
+				entries = append(entries, entry)
+			}
+			results := recurseEntries(entries)
+			out = append(out, results...)
+		case *chezmoi.File, *chezmoi.Symlink:
+			out = append(out, v.TargetName())
+		}
+	}
+	return out
+}
+
+func (c *Config) runManagedCmd(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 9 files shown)


cmd/managed.go
```diff
@@ -0,0 +1,63 @@
...
+	PreRunE: config.ensureNoError,
+	RunE:    config.runManagedCmd,
+}
+
+func init() {
+	rootCmd.AddCommand(managedCmd)
+}
+
+func recurseEntries(in []chezmoi.Entry) []string {
+	out := make([]string, len(in))
```
twpayne, line 26: I think this should be

```go
out := make([]string, 0, len(in))
```

Similarly for other calls to `make`.

cmd/managed.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+	entries := make([]chezmoi.Entry, len(ts.Entries))
+	for _, entry := range ts.Entries {
+		entries = append(entries, entry)
+	}
+	allManaged := recurseEntries(entries)
+	for _, tn := range allManaged {
+		path := filepath.Join(ts.DestDir, tn)
+		entry, _ := ts.Get(c.fs, path)
```
twpayne, line 55: This should return any error, rather than discarding it.

cmd/managed.go
```diff
@@ -0,0 +1,63 @@
...
+		managed := entry != nil
+		ignored := ts.TargetIgnore.Match(tn)
+		if !ignored && managed {
+			fmt.Println(path)
```
twpayne, line 59: Please use `fmt.Printf(c.Stdout, ...)` as this makes it easier to test.

twpayne:
For extra bonus points, would you mind adding a test? This test should be a good skeleton:

  https://github.com/twpayne/chezmoi/blob/master/cmd/dump_test.go#L14-L30

After equivalent code that runs the `managed` command rather than the `dump` command, you should be able to test that `stdout.String()` matches the output that you're expecting.

remigourdon:
Just pushed some new commits following your comments. Thanks a lot for walking me through the process as this is actually my very first pull request on a public repository!

twpayne:
Thanks - this looks great.

Please can you squash everything into a single commit and then we can merge this.

Let me know if you need the git commands to squash everything.

remigourdon:
Done, thanks!

twpayne:
Thank you very much for this great PR :)

PR #667, opened 2020-04-13
Add possibility to configure GPG cli command in config file

seberm:
There are cases where we want to be able configure the gpg CLI command.

For example I am user of the QubesOS and I use a separate VM for the gpg (split GPG). That's why I need to use the `gpg-client-wrapper` when working with the keys.

Ref.:
- https://github.com/QubesOS/qubes-app-linux-split-gpg

### DIFF
--- a/internal/chezmoi/gpg.go
+++ b/internal/chezmoi/gpg.go
@@ -9,6 +9,7 @@ import (
 
 // GPG interfaces with gpg.
 type GPG struct {
+	Command   string
 	Recipient string
 	Symmetric bool
 }
@@ -28,8 +29,9 @@ func (g *GPG) Decrypt(filename string, ciphertext []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	//nolint:gosec
 	cmd := exec.Command(
-		"gpg",
+		g.Command,
 		"--output", outputFilename,
 		"--quiet",
 		"--decrypt", inputFilename,
@@ -73,7 +75,9 @@ func (g *GPG) Encrypt(filename string, plaintext []byte) ([]byte, error) {
 		args = append(args, "--encrypt")
 	}
 	args = append(args, filename)
-	cmd := exec.Command("gpg", args...)
+
+	//nolint:gosec
+	cmd := exec.Command(g.Command, args...)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -139,6 +139,9 @@ func newConfig(options ...configOption) *Config {
 		Merge: mergeConfig{
 			Command: "vimdiff",
 		},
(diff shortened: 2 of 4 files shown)


internal/chezmoi/gpg.go
```diff
@@ -9,10 +9,12 @@ import (
 
 // GPG interfaces with gpg.
 type GPG struct {
+	Command string
```
twpayne, line 12: Please run `gofmt` to format this file correctly.

internal/chezmoi/gpg.go
```diff
@@ -9,10 +9,12 @@ import (
...
 	Recipient string
 	Symmetric bool
 }
 
+
```
twpayne, line 17: nit: stray newline.

internal/chezmoi/gpg.go
```diff
@@ -28,8 +30,12 @@ func (g *GPG) Decrypt(filename string, ciphertext []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	if g.Command == "" {
+		g.Command = "gpg"
+	}
```
twpayne, line 35: Please set the default command to be `gpg` in [this function](https://github.com/twpayne/chezmoi/blob/master/cmd/config.go#L129) rather than using `if` statements here.

internal/chezmoi/gpg.go
```diff
@@ -59,6 +65,10 @@ func (g *GPG) Encrypt(filename string, plaintext []byte) ([]byte, error) {
 	}
 	outputFilename := inputFilename + ".gpg"
 
+	if g.Command == "" {
+		g.Command = "gpg"
+	}
```
twpayne, line 70: See previous comment. This can be removed.

internal/chezmoi/gpg.go
seberm: I moved it, but I am still getting `Subprocess launched with function call as argument or cmd arguments (gosec)` error from the CI. I am still very new to go, could you please help me to fix it?

Thank you for review!

twpayne: Of course!

This warning comes from a very strict linter that gets run as part of chezmoi's build process. This linter spots things which are valid Go but might cause unintended problems. In this case, it's warning that we're launching a command (subprocess) where the command to launch is controlled by a variable. This means that if someone can change the value of the variable they can change which command is launch. This might be a security problem in a web service but of course here it's exactly what what we want, so we should disable the lint warning for this line.

The way to do this is to put on the line before:

```go
//nolint:gosec
cmd := exec.Command(
	gpg.Command,
```

This disables the gosec linter for this single line.

chezmoi uses a lot of strict automation that makes it more solid but also makes it a bit intimidating to contribute to. Thanks for persisting with this :)

seberm: Thank you!

Hopefully everything is ok now :).

twpayne:
Perfect! Thank you very much for the excellent contribution :)

PR #672, opened 2020-04-16
Iterate on managed command

twpayne:
This builds on @remigourdon's excellent work in #665.

More comments on this PR to follow (but first I want to see if this passes on Windows).


twpayne:
What changed in this PR:

* The output is now sorted. This works around Go's undefined order of iteration over maps and means that the output, given the same input (source state), will always be the same, which makes testing easier and the program more predictable to anything that consumes its output.

* The recursive function to iterate over all entries has moved from a typeswitch in the `managed` command to a method on the `Entry` interface. Which is best is really not obvious.

* The output of the recursive function includes a return value of the same type as the accumulator input to the recursive function. This makes it a true accumulator, avoiding conversions between types on each recursive call.

* The original code did not include directories in its output, but did include symlinks. This change adds a command line option to make it explicit what's included in the output.

* The tests have been simplified and expanded to include the new functionality. They also shift the perspective from changing the expected paths to match the actual paths to computing a normalized actual paths to and comparing that to an unchanged expected paths, which shifts most of the logic out of the test.

What was already excellent in #665:

* The author talked about the change before making it, see https://dave.cheney.net/2019/02/18/talk-then-code as to why this is important.

* The author provided a working implementation that demonstrated a clear understanding of the project's internals.

* The author implemented a comprehensive test as part of the PR.

* The author used the CI system to refine their code.

So thank you @remigourdon for an excellent first PR on a public repo :)

PR #674, opened 2020-04-17
Add diff.pager config option

### LINKED ISSUE #673: Add optional support for diff-so-fancy
It would be really cool if `chezmoi diff` / `chezmoi <...> -v` could also use [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy), if it's available.
 
Since [this issue](https://github.com/so-fancy/diff-so-fancy#220), diff-so-fancy can display arbitrary unified diffs from stdin:
```diff -u file1 file2 | diff-so-fancy```

Alternatively I could also think of an option to make the diff tool completely configurable to the users choice, but that might be a little more difficult.

twpayne:
Fixes #673. cc @Gerrit-K.

This is isn't quite ready to merge yet, as it only uses the pager for `chezmoi diff`, not for `chezmoi -v`.

@Gerrit-K is this what you had in mind? Are you able to test this?

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -66,6 +66,7 @@ type Config struct {
 	Merge             mergeConfig
 	Bitwarden         bitwardenCmdConfig
 	CD                cdCmdConfig
+	Diff              diffCmdConfig
 	GenericSecret     genericSecretCmdConfig
 	Gopass            gopassCmdConfig
 	KeePassXC         keePassXCCmdConfig
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -1,10 +1,21 @@
 package cmd
 
 import (
+	"io"
+	"os/exec"
+	"strings"
+	"unicode"
+
 	"github.com/spf13/cobra"
 	"github.com/twpayne/chezmoi/internal/chezmoi"
+	"github.com/twpayne/go-shell"
 	bolt "go.etcd.io/bbolt"
 )
+
+type diffCmdConfig struct {
+	NoPager bool
+	Pager   string
+}
 
 var diffCmd = &cobra.Command{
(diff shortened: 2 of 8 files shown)


twpayne:
Note that the CI failure is due to an unrelated issue.

Gerrit-K:
Thanks for the quick PR! Yes that's exactly what I had in mind and personally I don't mind that `chezmoi -v` is not yet supported (as I only used it in `chezmoi apply -vn` which I believe is exactly the same as `chezmoi diff`). 

I tested the branch and it's already looking great! I've only spotted a few minor differences which you can see on the screenshot:
image
* Line numbers don't seem to be picked up correctly: `@@ -0,0 +1,7 @@` vs `@ .local/share/chezmoi/dot_vimrc:4 @`
* The `+` and `-` signs still appear on each modified line
* The colorings for completely added or removed lines are slightly off (it's not easy to see, but the green on the left is different from the green on the right, e.g for all lines in `.vimrc`)

It doesn't bother me that much, but if you happen to find and eliminate the cause that would be awesome!

twpayne:
Thanks for the detailed testing.

I think what's happening is:

* `diff-so-fancy` expects diffs to be in [git's format](https://git-scm.com/docs/diff-format).
* `chezmoi diff` is producing a mix of unified diffs and pseudo shell code, which `diff-so-fancy` only partially understands.

As people are much more used to git's diff format, I think it's worth changing the output of `chezmoi diff` to be compatible with git's. I'll open an issue for this.

In the meantime, I think this PR adding support for `diff.pager` can be merged. I did a bit of minor refactoring, and will merge it once CI passes.

twpayne:
#677 covers the diff output formt.

PR #678, opened 2020-04-18
Add git diff format

### LINKED ISSUE #677: Generate `git diff` compatible diffs from `diff` command

`chezmoi diff` (and `--verbose` output) produce output that is a mix of pseudo shell code and unified diffs. This does not follow any standard and cannot easily be consumed by other tools, e.g. as required for #673.

`chezmoi diff` should produce output in [git's diff format](https://git-scm.com/docs/diff-format), which is the effective standard given the widespread use of git.

None.

## Additional context

* [git's diff format](https://git-scm.com/docs/diff-format).
* [`go-git/go-git`](https://github.com/go-git/go-git) looks like it contains a diff formatter in [`plumbing/format/diff`](https://github.com/go-git/go-git/tree/master/plumbing/format/diff).
* [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy) consumes git diffs .

twpayne:
Fixes #677 and #673.

@Gerrit-K can you test this and check that the output looks sensible?

Automatic `diff-so-fancy` support (#673) can be done by adding the following to your config file:

```toml
[diff]
    format = "git"
    pager = "diff-so-fancy"
```


Gerrit-K:
Yep, works now:

image

Thanks!

twpayne:
Great! Thank you for the fast testing!

PR #697, opened 2020-04-23
install from homebrew-core

stephengroat:
https://github.com/Homebrew/homebrew-core/commit/caa44f8aac7347b1846df70e66a73044123b4d62

this change will eventually be mirrored to Linuxbrew, but I'm not 100% sure of that process


twpayne:
Thanks very much!

Looking at the Homebrew formula in https://github.com/Homebrew/homebrew-core/pull/53431 there are a couple of things missing from the formula:

1. The build command doesn't set the version of chezmoi.
2. The fish shell completion is not installed.

Please could you update the formula in homebrew core before we change the installation instructions. It's particularly important that the version is set as this is critical for debugging any problems that users report. There are [guidelines for packaging chezmoi in the contributing doc](https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md#packaging).

twpayne:
Will merge this after https://github.com/Homebrew/homebrew-core/pull/53636 is merged.

PR #716, opened 2020-04-28
Improved howto for tracking externally modified files

awilts:
The provided example for handling externally modified files did not work for me.
It assumes that the '.config' and 'Code' folders are public. But they are private by default.
I updated the howto to better reflect that.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -363,7 +363,7 @@ Some programs modify their configuration files. When you next run `chezmoi
 apply`, any modifications made by the program will be lost.
 
 You can track changes to these files by replacing with a symlink back to a file
-in your source directory, which is under version control. Here is a worked
+in your source directory, which is under version control. Here is a working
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks very much for the PR!

As it stands, there are a number of problems:

* The PR doesn't follow chezmoi's [contributing guide](https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md). Specifically, generated files have not been updated, the commits are not logically separate, the commit messages do not follow chezmoi's guidelines, and there's a merge commit.
* The markdown isn't correctly formatted (it uses single quotes instead of backquotes, the line is too long).
* There are two grammatical errors ("worked example" is deliberate, the added sentence is missing closing punctuation.

I've corrected these problems, while maintaining you as the author of the commit, in #716. Thank you again for your contribution :)

awilts:
Thank you for the feedback! I will watch out for that next time.

PR #737, opened 2020-05-09
Add a Windows setup implementation for testscript

zb140:
Also enable as many of the test scripts as possible on Windows and update them as needed.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -99,9 +99,33 @@ func setupPOSIXEnv(env *testscript.Env) error {
 	})
 }
 
+// Set up the testing environment for Windows.  Works the same as on POSIX with
+// the exception that there isn't currently a convenient way to override the
+// shell, so that feature is skipped.
 func setupWindowsEnv(env *testscript.Env) error {
-	// FIXME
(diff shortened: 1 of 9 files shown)


main_test.go
```diff
@@ -99,9 +99,33 @@ func setupPOSIXEnv(env *testscript.Env) error {
 	})
 }
 
+// Set up the testing environment for Windows.  Works the same as on POSIX with
```
twpayne, line 102: nit: Standard Go doc comments begin with the function name, so this should be:

```go
// setupWindowsEnv sets up the testing environment for Windows. Works the same as...
```

zb140: Fixed. You can tell I'm not usually a Go programmer 😄

twpayne:
Brilliant, thank you @zb140 :)

For info, I've got a PR in progress that adds a load more integration tests, which, thanks to your work, should work on Windows too.

I also updated `README.md` to make it clearer that Windows support is a first-class citizen alongside Linux and macOS.

Once the integration tests are in, I'll head back to the internal refactoring.

PR #738, opened 2020-05-09
Add more integration tests

twpayne:


testdata/scripts/hg.txt
```diff
@@ -0,0 +1,42 @@
...
+chezmoi source -- --version
+stdout 'Mercurial Distributed SCM'
+
+# create a commit
+chezmoi add $HOME${/}.bashrc
+chezmoi hg -- add dot_bashrc
+chezmoi hg -- commit -m 'Add dot_bashrc'
+
+# test that chezmoi init clones a mercurial repo
+env HOME=$WORK${/}home2${/}user
```
zb140, line 17: Anything that relies on changing `$HOME` will most likely fail on Windows, which uses `$USERPROFILE` instead.  In the other tests, I fixed this by putting `[windows] env USERPROFILE=$HOME` after every place that changes `$HOME`.

twpayne: Yeah, I just found that out!

I've added a `chhome` command ... am about to update the PR.

PR #756, opened 2020-05-18
Fix hostname condition in how-to page

ehudon:
On line 219, we can read the following:

```
{{- if ne .chezmoi.hostname "work-laptop" }}
    .work # only manage .work on work-laptop
{{- end }}
```
The comment in the condition hints at the body of the condition will be evaluated if this is the hostname is `work-laptop`. However, the condition uses negation instead of equality.

I suggest to update it to:

```
{{- if eq .chezmoi.hostname "work-laptop" }}
```
so the condition reflects the intention and also matches the same condition on line 176.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -216,7 +216,7 @@ that chezmoi should ignore, and are interpreted as templates. An example
 `.chezmoiignore` file might look like:
 
     README.md
-    {{- if ne .chezmoi.hostname "work-laptop" }}
+    {{- if eq .chezmoi.hostname "work-laptop" }}
     .work # only manage .work on work-laptop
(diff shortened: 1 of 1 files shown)


twpayne:
This patch is not correct.

I agree that it's better to use positive logic where possible, but in this case the negative logic is required.

What we want to achieve is "only install `.work` if hostname is `work-laptop`" but chezmoi installs everything by default, so we have to turn the logic around to be "ignore `.work` unless the hostname is `work-laptop`".

ehudon:
Sorry. My mistake. :) Will close this PR.

twpayne:
No worries, thanks for contributing :)

PR #759, opened 2020-05-22
Add Chocolatey to docs

rasmuskriest:
As someone who heavily relies on [Chocolatey ](https://chocolatey.org/) instead of Scoop for installing most software, I took the liberty to create a [package](https://chocolatey.org/packages/chezmoi) ([package source](https://github.com/rasmuskriest/Chocolatey-Packages/tree/master/chezmoi)) that updates automatically based on the GitHub releases.

This PR updates all documentation to include the Chocolatey package after/below Scoop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ before you can use your system.
 chezmoi is distributed as a single stand-alone statically-linked binary with no
 dependencies that you can simply copy onto your machine and run. chezmoi
 provides one-line installs, pre-built binaries, packages for Linux and BSD
-distributions, Homebrew formulae, Scoop support on Windows, and a initial config
+distributions, Homebrew formulae, Scoop and Chocolatey support on Windows, and a initial config
 file generation mechanism to make installing your dotfiles on a new machine as
 painless as possible.
 
--- a/chezmoi.io/content/_index.md
+++ b/chezmoi.io/content/_index.md
@@ -153,7 +153,7 @@ before you can use your system.
 chezmoi is distributed as a single stand-alone statically-linked binary with no
 dependencies that you can simply copy onto your machine and run. chezmoi
 provides one-line installs, pre-built binaries, packages for Linux and BSD
-distributions, Homebrew formulae, Scoop support on Windows, and a initial config
+distributions, Homebrew formulae, Scoop and Chocolatey support on Windows, and a initial config
 file generation mechanism to make installing your dotfiles on a new machine as
 painless as possible.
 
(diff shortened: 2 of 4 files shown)


twpayne:
Thanks very much for this!

I tried installing chezmoi using `choco install chezmoi` and got the error:

```
chezmoi not installed. The package was not found with source(s) listed.
```

I [searched for chezmoi on chocolately.org](https://chocolatey.org/packages?q=chezmoi) but didn't find any packages.

Are some extra steps required, e.g. adding another source?

rasmuskriest:
Thanks for the quick reply - I did not think that the PR get checked faster than the actual package.

For newly created packages, the approval process sometimes takes a couple of days since manual (as in: human) action is required. The package will only be listed publicly after this and thereby show up upon search/allowed to be installed without any further commands. From my experience, new versions are approved much faster after initial release.

In case you would like to check whether the package works, `choco install chezmoi --version=1.8.1` should work fine.

I will get back to this PR as soon as I hear back from the Chocolatey moderators.

twpayne:
Excellent, thanks for the update. Is there a public issue that I can subscribe to track the progress of chezmoi in Chocolately? I had a quick poke around the chocolately.org website and its [GitHub repos](https://github.com/chocolatey) but couldn't immediately spot where its package definitions are kept.

Is there anything that we can add to chezmoi's release process to help Chocolately get notified of new releases of chezmoi?

rasmuskriest:
Chocolatey is quite a bit different to Homebrew/Linuxbrew as it utilizes NuGet packages in feeds (the one you linked above begin the community one) instead of formulas in Git repositories. Therefore, package definitions are separate from the actual repository (although the downloaded NuGet package is just a compressed archive of said definition).

You may find the current package status on the already available [package page](https://chocolatey.org/packages/chezmoi). As mentioned above, the package is already uploaded but pending moderation so that a search does result in "not found".

The package definition for `chezmoi` is [in my GitHub](https://github.com/rasmuskriest/Chocolatey-Packages/tree/master/chezmoi). I utilize the [recommended method for automated packaging](https://chocolatey.org/docs/automatic-packages), handled by AppVeyor. Since the `[update.ps1](https://github.com/rasmuskriest/Chocolatey-Packages/blob/master/chezmoi/update.ps1)` is being executed multiple times every day and simply checks GitHub releases everything should work fine - as long as the Go packaging continues to name the Windows versions `_windows_amd64.zip` and `_windows_i386.zip`.

rasmuskriest:
The Chocolatey package has been approved. @twpayne Feel free to merge!

twpayne:
Merged :) Thanks for a really nice contribution :)

Issue #763, opened 2020-05-23
Allow specifying vault when fetching 1password secrets


zb140:
## Is your feature request related to a problem? Please describe.

Fetching secrets from 1password is frustratingly slow. For example, on my system, each call to `op get` takes around 1800ms.  For my repo with 4 secret fetches, this means `chezmoi apply` takes around 5 full seconds to complete ☹️ 

## Describe the solution you'd like

I recently learned that specifying which vault to fetch a secret item from (using the `--vault` argument) significantly speeds up access -- calls to `op get item --vault` only take around 800ms.  I put together a prototype (over here: zb140/chezmoi@0cf25b57608566f0a2fa2b7f024b7214ce5a51cc), and with the same repo, `chezmoi apply` now completes in about 2.5 seconds.  Still quite a bit slower than I'd like, but a definite improvement.

## Describe alternatives you've considered

I don't think there's a whole lot `chezmoi` itself can do to help with performance of the secret commands -- in my situation, for example, nearly all of the slowness is directly attributable to the `op get` call.

One idea I did have was that perhaps there could be an option to `chezmoi apply` that would skip processing templates with secrets in them. Since my secrets don't change terribly often, as a user I generally know whether it's safe to skip re-fetching them.  But that sounds like it would be hard to implement.

Or perhaps it could be a property of some templates that they're only re-evaluated sometimes, but that isn't great either -- it seems like a big potential source of confusion for users.

## Additional context

Once there's a clear path forward, I'll fix up my prototype as needed and submit a PR.

twpayne:
> Fetching secrets from 1password is frustratingly slow.

Oof. Sad face.

A few thoughts:

> One idea I did have was that perhaps there could be an option to `chezmoi apply` that would skip processing templates with secrets in them. Since my secrets don't change terribly often, as a user I generally know whether it's safe to skip re-fetching them. But that sounds like it would be hard to implement.

Skipping processing templates with secrets in them would be tricky to implement indeed. However, chezmoi already maintains a persistent state to remember which `run_once_` scripts have been run (and soon for the `chezmoi status` command to remember what the last `chezmoi apply` wrote). We could extend this persistent state to optionally cache 1password secrets too. As all of the secrets already exist in plaintext in the user's home directory, I don't think there are any adverse security consequences to this. A command like `chezmoi secret onepassword flush-cache` could then flush the cache so chezmoi re-requests the new secrets from 1password.

[...]

PR #764, opened 2020-05-24
Allow specifying vault when fetching 1password secrets

zb140:
This fixes the part of #763 targeted at v1 and provides a significant performance boost when fetching items from 1password (if the vault is also specified).  Other perf improvements may come in later versions.

### DIFF
--- a/cmd/secretonepassword.go
+++ b/cmd/secretonepassword.go
@@ -39,12 +39,27 @@ func (c *Config) runOnepasswordCmd(cmd *cobra.Command, args []string) error {
 	return c.run("", c.Onepassword.Command, args...)
 }
 
-func (c *Config) onepasswordFunc(item string) interface{} {
-	if data, ok := onepasswordCache[item]; ok {
+func (c *Config) onepasswordFunc(args ...string) interface{} {
+	if len(args) < 1 || len(args) > 2 {
+		panic(fmt.Sprintf("onepassword: expected 1 or 2 arguments, got %d", len(args)))
+	}
+
+	item := args[0]
+	vault := ""
+	key := item
+	if len(args) == 2 {
+		vault = args[1]
+		key += "\x00" + vault
+	}
+
+	if data, ok := onepasswordCache[key]; ok {
 		return data
 	}
 	name := c.Onepassword.Command
-	args := []string{"get", "item", item}
+	args = []string{"get", "item", item}
+	if vault != "" {
+		args = append(args, "--vault", vault)
+	}
 	cmd := exec.Command(name, args...)
 	cmd.Stdin = os.Stdin
 	cmd.Stderr = os.Stderr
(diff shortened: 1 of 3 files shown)


docs/REFERENCE.md
```diff
@@ -1023,31 +1023,37 @@ further parsing is done on the `note` field.
 
     {{ (index (lastpassRaw "SSH Private Key") 0).note }}
 
-### `onepassword` *uuid*
+### `onepassword` *uuid* [*vault uuid*]
```
twpayne, line 1026: nit: please change `*vault uuid*` to `*vault-uuid*` (i.e. add a hyphen in the middle) so the new argument visually parses as a single token.

twpayne: Same comment in the use of `*vault uuid*` below.

docs/REFERENCE.md
```diff
@@ -1023,31 +1023,37 @@ further parsing is done on the `note` field.
...
 The output from `op` is cached so calling `onepasswordDocument` multiple times with the
-same *uuid* will only invoke `op` once.
+same *uuid* will only invoke `op` once.  If the optional *vault uuid* is supplied,
+it will be passed along to the `op get` call, which can significantly improve
+performance.
 
 #### `onepasswordDocument` examples
 
     {{- onepasswordDocument "<uuid>" -}}
+    {{- onepasswordDocument "<uuid>" "<vault>" -}}
```
twpayne, line 1056: crazy nit: please make the second argument `<vault-uuid>` to match the syntax used in my previous comments :)

cmd/secretonepassword.go
```diff
@@ -39,12 +39,27 @@ func (c *Config) runOnepasswordCmd(cmd *cobra.Command, args []string) error {
 	return c.run("", c.Onepassword.Command, args...)
 }
 
-func (c *Config) onepasswordFunc(item string) interface{} {
-	if data, ok := onepasswordCache[item]; ok {
+func (c *Config) onepasswordFunc(args ...string) interface{} {
+	if len(args) < 1 || len(args) > 2 {
+		panic(fmt.Errorf("function takes an item ID and optionally a vault ID"))
```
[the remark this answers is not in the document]
zb140: Sure, I'm happy to make these changes :smile:

I actually deleted the `onepassword:` prefix because it was already being added by whatever ultimately prints out the message.  Before I deleted it, this was the message I was getting:
`chezmoi: template: C:\Users\zb\.local\share\chezmoi\1pass-test.tmpl:2:16: executing "C:\\Users\\zb\\.local\\share\\chezmoi\\1pass-test.tmpl" at <onepassword "uuid" "vault" "foo" "bar" "baz">: error calling onepassword: onepassword: function takes an item ID and optionally a vault ID`

But I totally get not wanting to rely on that behavior, especially since it's not obvious from the call site that it will happen.

twpayne:
Perfect, thank you!

PR #771, opened 2020-05-27
 Read KeepassXC passwords from stdin rather than stdout

### LINKED ISSUE #768: Keepass password input bug?
## Describe the bug

Not sure if this is just on Windows, but I ran across a problem while trying to put together a secret caching script with KeepassXC as suggested in #763 .  When I try to use the `keepassxc` template function, I get this error:
```
Insert password to unlock c:\users\zb\.op_cache\cache.db:
chezmoi: template: C:\Users\zb\.local\share\chezmoi\1pass-test.tmpl:1:13: executing "C:\\Users\\zb\\.local\\share\\chezmoi\\1pass-test.tmpl" at <keepassxc "foo/bar">: error calling keepassxc: keepassxc: keepassxc-cli.exe show --show-protected 'c:\\users\\zb\\.op_cache\\cache.db' foo/bar: read stdin: The handle is invalid.
```
I believe it's due to this line: https://github.com/twpayne/chezmoi/blame/master/cmd/secretkeepassxc.go#L133 . It looks like it's trying to read the password from stdout.  Indeed, changing it to `os.Stdin.Fd()` fixes the error and I'm able to enter the password and read secrets from my db successfully.

## To reproduce

1. Create a keepassxc database with some stuff in it
1. Attempt to read it with the `keepassxc` template function

```
$ chezmoi command --verbose
```
The error is pasted above.

## Output of `chezmoi doctor`

```
$ chezmoi doctor
     ok: version 1.8.1, commit 4def4f3b96cb01cefe8d7887e8dbfb77908ef267, built at 2020-05-06T18:05:06Z, built by goreleaser
     ok: runtime.GOOS windows, runtime.GOARCH amd64
  ERROR: C:\Users\zb\.local\share\chezmoi (source directory, perm 777)
     ok: C:\Users\zb (destination directory, perm 777)
     ok: C:\Users\zb\.config\chezmoi\chezmoi.toml (configuration file)
     ok: C:\WINDOWS\system32\cmd.exe (shell)
     ok: c:\users\zb\.op_cache\cache.db (KeePassXC database)
     ok: C:\WINDOWS\gvim.bat (editor)
     ok: C:\WINDOWS\vimdiff.bat (merge command)
     ok: C:\Program Files\Git\cmd\git.exe (source VCS command, version 2.26.2)
     ok: C:\Program Files\Git\usr\bin\gpg.exe (GnuPG, version 2.2.20)
     ok: C:\Users\zb\scoop\shims\op.exe (1Password CLI, version 1.0.0)
warning: bw (Bitwarden CLI, not found)
     ok: C:\Users\zb\scoop\shims\gopass.exe (gopass CLI, version 1.9.2)
     ok: C:\Program Files\KeePassXC\keepassxc-cli.exe (KeePassXC CLI, version 2.5.4)
warning: lpass (LastPass CLI, not found)
warning: pass (pass CLI, not found)
warning: vault (Vault CLI, not found)
```

## Additional context

Like I said, I'm not sure if this is just a Windows problem -- I haven't tried it on Linux yet, but it seems like others would have encountered this if it were a more widespread issue.

Also, ultimately I'd like to be able to pipe the password in (or provide it some way other than interactively), but I've just tried that and it doesn't work either with `os.Stdout` or with my updated `os.Stdin` version. No idea why yet...

zb140:
* Fix a bug that tried to read KeepassXC passwords from stdout rather than stdin (Fixes #768)
 * Allow reading passwords from stdin even when it's not a terminal.
 * Add an integration test for KeepassXC features

### DIFF
--- a/cmd/secretkeepassxc.go
+++ b/cmd/secretkeepassxc.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"regexp"
@@ -127,10 +128,43 @@ func (c *Config) keePassXCAttributeFunc(entry, attribute string) string {
 	return outputStr
 }
 
+func readPassword(prompt string) (pw []byte, err error) {
+	fd := int(os.Stdin.Fd())
+	if terminal.IsTerminal(fd) {
+		fmt.Print(prompt)
+		pw, err = terminal.ReadPassword(fd)
+		fmt.Println()
+		return
+	}
+
+	var b [1]byte
+	for {
+		n, err := os.Stdin.Read(b[:])
+		// terminal.ReadPassword discards any '\r', so do the same
+		if n > 0 && b[0] != '\r' {
+			if b[0] == '\n' {
+				return pw, nil
+			}
+			pw = append(pw, b[0])
+			// limit size, so that a wrong input won't fill up the memory
+			if len(pw) > 1024 {
+				err = errors.New("password too long")
+			}
+		}
+		if err != nil {
+			// terminal.ReadPassword accepts EOF-terminated passwords
+			// if non-empty, so do the same
+			if err == io.EOF && len(pw) > 0 {
+				err = nil
+			}
+			return pw, err
+		}
+	}
+}
+
 func (c *Config) runKeePassXCCLICommand(name string, args []string) ([]byte, error) {
 	if keePassXCPassword == "" {
-		fmt.Printf("Insert password to unlock %s: ", c.KeePassXC.Database)
-		password, err := terminal.ReadPassword(int(os.Stdout.Fd()))
+		password, err := readPassword(fmt.Sprintf("Insert password to unlock %s: ", c.KeePassXC.Database))
 		fmt.Println()
 		if err != nil {
 			return nil, err
--- a/testdata/scripts/secretkeepassxc.txt
+++ b/testdata/scripts/secretkeepassxc.txt
@@ -1,0 +1,66 @@
+stdin $HOME/input
+chezmoi secret keepassxc -- show --show-protected secrets.kdbx example.com
+stdout examplelogin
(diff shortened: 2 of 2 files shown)


cmd/secretkeepassxc.go
```diff
@@ -127,10 +128,43 @@ func (c *Config) keePassXCAttributeFunc(entry, attribute string) string {
 	return outputStr
 }
 
+func readPassword(prompt string) (pw []byte, err error) {
+	fd := int(os.Stdin.Fd())
+	if terminal.IsTerminal(fd) {
+		fmt.Fprint(os.Stderr, prompt)
```
twpayne, line 134: I think this should write to `os.Stdout` not `os.Stderr`, same for line 136.

cmd/secretkeepassxc.go
```diff
@@ -127,10 +128,43 @@ func (c *Config) keePassXCAttributeFunc(entry, attribute string) string {
...
+		pw, err = terminal.ReadPassword(fd)
+		fmt.Fprintln(os.Stderr)
+		return
+	}
+
+	var b [1]byte
```
twpayne, line 140: Note that this branch does not print the prompt, is that deliberate?

Also, I'm a bit surprised that `terminal.ReadPassword` doesn't work on Windows. Are you sure that's the case?

twpayne: Ah, I've just seen [your comment in the other issue explaining this](https://github.com/twpayne/chezmoi/issues/768#issuecomment-634388601), thanks.

cmd/secretkeepassxc.go
zb140: Sure, I'll go ahead and update it.

zb140: Fixed

twpayne:
Great, thank you for another excellent contribution!

Issue #787, opened 2020-06-19
Include other files in templates


hg-zt:
This would make managing configurations vary from machines more flexible.

## Describe the solution you'd like

`config.tmpl`
```ini
# Common
## Keybindings
"Ctrl-A" = ...
"Ctrl-B" = ...
## Other settings
show_line_numbers = true

# OS specific
{{ if eq .chezmoi.os "linux" -}}
cache_dir = $XDG_CACHE_HOME/my-app
  {{ include linux.conf }}
{{ else if eq .chezmoi.os "darwin" -}}
  ...
{{ end -}}

# Machine specific
; Or maybe in this form? I'm not familiar with golang so I'don't know if it's possible
{{ include {{ .chezmoi.hostname }}.conf }}

```

In fact, I'm already managing my emacs config in this way. Since elisp is a programming language, it's easy to load(include) other files. But many config file types (`.toml`, `.conf`, etc.) are lack of such abilities.

twpayne:
You can use a `.chezmoitemplates` directory to do almost exactly this, see the [section in the reference manual](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md#chezmoitemplates). See also [this issue discussion](https://github.com/twpayne/chezmoi/issues/749#issuecomment-628048523) for a worked example including passing parameters.

chezmoi could certainly be better documented here.

hg-zt:
Thanks for your quick reply!
I did find `{{ template ... }}` on the golang template reference page but do not have idea of how to use it.

Legion2:
@hg-zt you use `{{ template ... }}` just like `{{ include... }}` but you only can include files from the `.chezmoitemplates` directory.

@twpayne can we also "include" files outside `.chezmoitemplates`? For Example:
`.chezmoiignore`
```
.bashrc_*
```

`dot_bashrc_linux`
...
`dot_bashrc_windows`
...

`dot_bashrc.tmpl`
```
{{ if eq .chezmoi.os "windows" }}
{{ include "dot_bashrc_windows" }}
{{ end }}
{{ if eq .chezmoi.os "linux" }}
{{ include "dot_bashrc_linux" }}
{{ end }}
´´´
So we don't have to use symlinks

twpayne:
> @twpayne can we also "include" files outside `.chezmoitemplates`?

Right now, you can only include files in `.chezmoitemplates`, but I can add a template function to include arbitrary files.

PR #790, opened 2020-06-20
Add include template function

twpayne:
As [suggested by](https://github.com/twpayne/chezmoi/issues/787#issuecomment-646823638) @Legion2.

This adds a template function `include` that includes the contents of a file relative to the source directory.

Notes:
* The literal contents of the file are included, the file is not interpreted as a template.
* If you want to include a template, then you need to put the file in `.chezmoitemplates` and use the `template` function.

@Legion2 I'm not sure whether we should add this there are a number of downsides:
* The `include`d file has to be added to `.chezmoiignore` or start with a `.` to prevent chezmoi from creating a file in the target state of the same name.
* The `include`d file cannot be a template itself.

In short, `include` does not add anything that cannot already be achieved with `.chezmoitemplates`, with the exception of allowing the `include`d file to be outside the `.chezmoitemplates` directory.

What are your thoughts on this?

### DIFF
--- a/cmd/docs.gen.go
+++ b/cmd/docs.gen.go
@@ -1661,6 +1661,7 @@ func init() {
 		"* [Template functions](#template-functions)\n" +
 		"  * [`bitwarden` [*args*]](#bitwarden-args)\n" +
 		"  * [`gopass` *gopass-name*](#gopass-gopass-name)\n" +
+		"  * [`include` *filename*](#include-filename)\n" +
 		"  * [`keepassxc` *entry*](#keepassxc-entry)\n" +
 		"  * [`keepassxcAttribute` *entry* *attribute*](#keepassxcattribute-entry-attribute)\n" +
 		"  * [`keyring` *service* *user*](#keyring-service-user)\n" +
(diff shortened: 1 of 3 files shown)


Legion2:
I currently have the setup 
`.chezmoiignore`
```
{{ if ne .chezmoi.os "windows" }}
*_windows
{{ end }}
{{ if ne .chezmoi.os "linux" }}
*_linux
{{ end }}
```

`dot_bashrc_linux`
...
`dot_bashrc_windows`
...


And then have symlink templates to link the correct os version file. For example
`symlink_dot_bashrc.tmpl`
`.bashrc_{{ .chezmoi.os }}`

But this creates two files in my home dir. The `.bashrc_linux` and the symlink `.bashrc` pointing to `.bashrc_linux`.
When using include this does only generate one file im the home dir.

Using the `.chezmoitemplates` directory looks like a good alternative, but then the files are not at the correct position in the git repository.

twpayne:
Thanks for the feedback, so it looks like this is a useful addition. Let's go with it :)

Legion2:
I think this new feature should be at least mentioned in the [How To guide](https://github.com/twpayne/chezmoi/blob/master/docs/HOWTO.md#use-completely-separate-config-files-on-different-machines) for "Use completely separate config files on different machines".

twpayne:
Totally agree! I'm planning a re-work of the templates documentation. I'll open an issue for this.

PR #815, opened 2020-07-22
Add timeout when opening bolt database

ptxmac:
I've recently had issues where a stuck chezmoi process kept the lock on the database causing a new run to stall forever. This ensures the user at least get an error instead of an infinite wait.

The 2 seconds are chosen completely arbitrarily :smile:

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -14,6 +14,7 @@ import (
 	"runtime"
 	"strings"
 	"text/template"
+	"time"
 	"unicode"
 
 	"github.com/Masterminds/sprig"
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks for this. Do you have any idea what caused the chezmoi process to get stuck in the first place?

ptxmac:
> Thanks for this. Do you have any idea what caused the chezmoi process to get stuck in the first place?

yeah, it's basically a user error - I had a "run_once" script that installs oh my fish (extension manager for fish shell).
As part of the installation, it starts a new fish shell.

So my process tree would end up like this:

```
+ fish
 + chezmoi
  + install_omf
   + fish
```
 So the apply was never allowed to terminate

twpayne:
Ah, thanks for the explanation. That makes sense.

PR #826, opened 2020-07-28
Added description how to add folders

ASDFGamer:
This adds a info that folders can be added via the -r argument to the quick-start guide.
fixes #825

PR #827, opened 2020-07-30
Added description how to add folders

### LINKED ISSUE #825: Add info how to add folders to quick-start
I just started using Chezmoi and used the quick-start guide to learn how to use it.
There is a example where the bash config is added via the add command. I thought that the add command could also be used to add whole folders without other arguments. This lead to confusion as i looked into the generated repo.

I would propose to add a short statement into the quick-start, that whole folders can be added via the -r argument.

There could also be a example for that, but I think that is too much for the quick-start.

## Additional context
I don't have any additional context

twpayne:
Contributed by @ASDFGamer, replaces #826 and fixes #825.

@ASDFGamer the commit has you as the author, to preserve credit.

### DIFF
--- a/cmd/docs.gen.go
+++ b/cmd/docs.gen.go
@@ -1538,7 +1538,8 @@ func init() {
 		"\n" +
 		"    chezmoi add ~/.bashrc\n" +
 		"\n" +
-		"This will copy `~/.bashrc` to `~/.local/share/chezmoi/dot_bashrc`.\n" +
+		"This will copy `~/.bashrc` to `~/.local/share/chezmoi/dot_bashrc`. If you want\n" +
+		"to add a whole folder to chezmoi, you have to add the `-r` argument after `add`.\n" +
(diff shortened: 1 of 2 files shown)


aniketgm:
@twpayne I am here after searching for a way to add folders instead of files into chezmoi. The website doesn't show the changes of this PR. I had to search through multiple issues, which lead me here. Can you have a quick look why the website is not updated ?

twpayne:
This documentation was for an earlier version of chezmoi.

To add a folder, but not its contents, use `chezmoi add --recursive=false $FOLDER`.

twpayne:
See https://www.chezmoi.io/developer/website/.

This would probably best fit in https://github.com/twpayne/chezmoi/blob/d16ce211bcd2c55cc1c9ad8df13f3222944caf5f/assets/chezmoi.io/docs/user-guide/manage-different-types-of-file.md.

Issue #784, opened 2020-06-11
`error calling pass: pass: pass show foo/bar: exit status 2` in ssh shells


Grimler91:
First of all, thanks for chezmoi. This is the third dotfile handling system I try and it is more flexible and less complex than what I previously tried (a homemade solution and dotdrop).

## Describe the bug

`chezmoi apply` or `chezmoi diff` gives an error 

```
/usr/bin/pass show foo/bar: exit status 2
chezmoi: template: /home/grimler/.local/share/chezmoi/dot_foo.tmpl:6:25: executing "/home/grimler/.local/share/chezmoi/dot_foo.tmpl" at <pass "foo/bar">: error calling pass: pass: pass show foo/bar: exit status 2
```

if run from a terminal without Xorg/wayland running, so for example in an ssh shell (without the -X option).

## To reproduce

Create a dotfile.tmpl with `{{ pass foo/bar }}`, create a `foo/bar` file in pass and then ssh into localhost to get a ssh shell `ssh localhost`, and then try `chezmoi diff`.

I also run into this problem when running chezmoi on my phone, in [termux](https://termux.com/).

## Expected behavior

`chezmoi diff` show show me pending changes instead of giving the `/usr/bin/pass show foo/bar: exit status 2` error.

## Output of command with `--verbose`

Exactly the same output
```
$ chezmoi diff --verbose
/usr/bin/pass show foo/bar: exit status 2
chezmoi: template: /home/grimler/.local/share/chezmoi/dot_foo.tmpl:6:25: executing "/home/grimler/.local/share/chezmoi/dot_foo.tmpl" at <pass "foo/bar">: error calling pass: pass: pass show foo/bar: exit status 2
```

## Output of `chezmoi doctor`

```
$ chezmoi doctor
warning: version dev
     ok: runtime.GOOS linux, runtime.GOARCH amd64
     ok: /home/grimler/.local/share/chezmoi (source directory, perm 700)
     ok: /home/grimler (destination directory, perm 710)
     ok: /home/grimler/.config/chezmoi/chezmoi.toml (configuration file)
     ok: /bin/bash (shell)
     ok: /usr/bin/vi (editor)
     ok: /usr/bin/vimdiff (merge command)
     ok: /usr/bin/git (source VCS command, version 2.27.0)
     ok: /usr/bin/gpg (GnuPG, version 2.2.20)
warning: op (1Password CLI, not found)
warning: bw (Bitwarden CLI, not found)
[...]

PR #839, opened 2020-08-07
Change pass unlock command

Grimler91:
Hi,

This changes the unlock command for pass from `^$` to `$^`. `^$` matches empty lines, causing the unlock step to print some grep matches if empty lines occur in the pass secrets. `$^` on the other hand (end of line followed by beginning of line) should never match anything. 

Note that `$^` was what @m-rey suggested from the beginning in https://github.com/twpayne/chezmoi/issues/784. 

Both `make test` and `make lint` succeed without error. I do not have access to a windows box though so I am not sure if the test on windows is correct. Based on the comment it seems that only leading ^'s are dropped, so that we can match `grep $^`, but I have not been able to verify this (we'll see when the CI runs I guess).

### DIFF
--- a/cmd/secretpass.go
+++ b/cmd/secretpass.go
@@ -41,7 +41,7 @@ func (c *Config) passFunc(id string) string {
 	}
 	name := c.Pass.Command
 	if !c.Pass.unlocked {
-		args := []string{"grep", "^$"}
+		args := []string{"grep", "$^"}
 		cmd := exec.Command(name, args...)
 		cmd.Stdin = c.Stdin
 		cmd.Stdout = c.Stdout
--- a/testdata/scripts/secretpass.txt
+++ b/testdata/scripts/secretpass.txt
@@ -8,7 +8,7 @@ cmp $HOME/.netrc golden/.netrc
 #!/bin/sh
 
 case "$*" in
-"grep ^$")
+"grep $^")
     ;;
 "show misc/example.com")
     echo "examplepassword"
@@ -19,8 +19,9 @@ case "$*" in
 esac
 -- bin/pass.cmd --
 @echo off
-REM Windows drops the leading ^ from the ^$ argument that chezmoi passes to "pass grep" so match on $ only.
-REM For background information, read http://daviddeley.com/autohotkey/parameters/parameters.htm#WIN.
+REM Matching "grep $^" seems more or less impossible on windows as it
+REM is an escape character in the cmd.exe shell. See
+REM https://github.com/twpayne/chezmoi/pull/839#issuecomment-670800462
 IF "%*" == "grep $" (
     exit /b 0
 ) ELSE IF "%*" == "show misc/example.com" (


Grimler91:
I think the failed windows test means that '^' is still dropped (and I should revert the second commit in this PR). Is this the correct interpretation?

twpayne:
Thanks for this. I suspect that Windows is interpreting `$^` as some kind of environment variable. I'll be able to test on Windows either tonight or sometime over the weekend.

twpayne:
Ah, correction, I won't be able to test on Windows for at least a couple of weeks. @zb140 might be able to help here.

zb140:
Sure thing, I can take a look at it a little later today.

zb140:
Oh, now that I've read the whole history of comments, I'm pretty sure I know what's going on here. The issue with the `^` symbol is that Windows batch script uses it as an escape character for pipe and redirection operators. I'm fairly certain that if you change the grep call to `grep $^^` (ie, escape the escape) it will work. I will be able to verify this after work this afternoon, but if someone gets to it sooner you could give it a try.

Grimler91:
I've pushed an update to add a second '^' to the windows testdata. If it solves it I can squash and re-write the git history of this PR to get something more tidy

zb140:
Thanks for giving it a try; too bad it didn't work.  My work day will be over in about an hour or so, and I'll be able to properly debug it then.

zb140:
Ok, so I understand what's going on here now.  I think the TL;DR is, just undo commit 1a9236b and let it go back to testing for `grep $` on Windows.

Here's the full explanation: as the comment in the test script implies, command line parsing is... quirky... on Windows.  The OS expects each process receiving arguments to parse the command-line string itself.  Which is weird, but not necessarily an issue until it's combined with the fact that the `secretpass.txt` test script uses a `pass.cmd` batch script to emulate the real `pass` command.  The thing responsible for parsing command lines in this situation, then, is the `cmd.exe` shell, which uses `^` as an escape character -- including for command line arguments that get passed into batch scripts!  So the only way to get a `^` into the script as an argument would be to go over [here](https://github.com/Grimler91/chezmoi/blob/pass-unlock/cmd/secretpass.go#L44) and replace `$^` with `$^^`, but only in the case when the command being run is a Windows batch script.  Which, apart from the fact that it would be difficult and fragile to detect, only has the single benefit of helping make this integration test script very slightly more technically "correct".

And, as if that weren't enough quirkiness (:wink:), the batch scripting language has a few other gems for us.  First, when a `^` appears inside a quoted string, it is always interpreted literally, meaning using 2 of them as an escape won't work correctly.  Which is understandable if maybe a little unexpected, but it does mean that the if condition in the test script actually needs to _not_ have an escaped `^` like I suggested before.  Lastly, when interpolating a variable into the script, any `^` that has been passed in will actually act as though you had written a literal `^` into the script, so something like these lines from the test script:
```
echo pass: invalid command %*
exit /b 1
```
will actually write `pass: invalid command grep $    exit /b 1` to stdout if `%*` ends with a `^` (because `^` is also a line-continuation character).

So, after explaining all that (if anyone's still reading :wink:) I just don't see the value in jumping through these hoops just for the sake of this test.  I can't imagine any bugs that doing all this would catch that wouldn't be caught by the existing script if we just didn't bother.

twpayne:
Wow, thank you @zb140 for teaching me much more than I ever wanted to know about scripts on Windows. I'm still not sure if I want to commit your amazing explanation to long term memory or forget it as quickly as I can.

Totally agree that this is test code and therefore more relaxed matching (e.g. matching on looser condition) is completely fine.

Thank you both very much for your work here.

zb140:
Oh, you definitely don't want to have to remember this nonsense -- definitely let GitHub do that part for you :wink:

My favorite part of the whole thing is that Microsoft didn't want to make any changes to the old scripting language for fear of backwards compatibility issues (fair enough) so they invented a whole new shell (Powershell) with an actual, decent scripting language, then recommended to everyone to stop using the old one.  So far so good.  But they made the default interpreter for the new language `notepad.exe` so people couldn't be tricked into "accidentally" running potentially dangerous scripts.  Which has the side effect of making it useless for these sorts of applications unless you can guarantee that whatever system these scripts run on has been preconfigured to actually be able to run Powershell scripts 😭

Grimler91:
Alright, thanks! I updated the comment in the testdata to point to the explaination above. I think this should be good to go then!

Issue #818, opened 2020-07-27
Support GitHub Codespaces, Visual Studio Codespaces and VS Code Dev Containers


felipecassiors:
Would be good if `chezmoi` could _support_ how `dotfiles` are managed in these services, using its default settings. 

## What needs to be made

There is no much work to do on `chezmoi` itself (I believe) and the outcome would be to probably have everything on the documentation.

### Write a `bootstrap.sh` script 

This script must be able to install `chezmoi` and reuse the current folder for `chezmoi init`. I have one [here](https://github.com/felipecassiors/dotfiles/blob/master/install) (with a bunch of unnecessary fashion).

### Support `~/dotfiles` as permanent source dir

This can be actually anywhere since the user can set it in the settings of these services. I believe this can be made with:

```
# bootstrap.sh
script_dir="$(cd "${BASH_SOURCE[0]%/*}" && pwd)"
chezmoi init --apply --source $script_dir
```

And have the following in the `.chezmoi.toml.tmpl`

```toml
# This key must be at the top level of the configuration file,
# for TOML, it means before any [field] such as [data]
sourceDir = "{{ .chezmoi.sourceDir }}"
```
This will pick up whatever source dir was passed as a parameter in the `chezmoi init`.

### Document everything, and:

How about adding a flag to `chezmoi init` to create the `bootstrap.sh` and the snippet on `.chezmoi.toml.tmpl`?

How about creating a [template repository](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-template-repository) with everything already set, to make it easy for newcomers? They wouldn't even need to install `chezmoi` by themselves to get started.

## References

- Visual Studio Codespaces: https://docs.microsoft.com/pt-br/visualstudio/codespaces/reference/personalizing
- GitHub Codespaces: https://docs.github.com/en/github/developing-online-with-codespaces/personalizing-codespaces-for-your-account
- Visual Studio Code Remote Development - Containers: https://code.visualstudio.com/docs/remote/containers#_personalizing-with-dotfile-repositories

twpayne:
This is a great suggestion, thanks.

I just got accepted to the Codespaces beta, #834 is a work in progress PR.

felipecassiors:
@twpayne, any words about this?

> How about adding a flag to chezmoi init to create the bootstrap.sh and the snippet on .chezmoi.toml.tmpl?
> 
[...]

Issue #760, opened 2020-05-23
chezmoi and VS Code remote containers dotfile support


thernstig:
## Is your feature request related to a problem? Please describe.

I am using VS Code as my main editor, and a growing pattern is to use [VS Code Remote - Containers](https://code.visualstudio.com/docs/remote/containers) (devcontainer.json) to spin up a development environment. If you work over multiple projects, with various technology stacks, this is a great way to setup each projects own stack.

But since the development is then run inside a container, I am missing my personal dotfiles for an efficient workflow. This is a perfect example for when dotfiles really shines.

VS Code dev containers has support for this, https://code.visualstudio.com/docs/remote/containers#_personalizing-with-dotfile-repositories, but the problem is that it expects it to work with e.g. a `~dotfiles/install.sh` command that it launches after having spun up the container.

I could find no information anywhere on how to get this to work in a seamless way with chezmoi. One problem is that chezmoi requires a binary. One option might be to have an install.sh file that actually installs chezmoi and runs the proper commands to setup the environment, but I am unsure that is the best way.

## Describe the solution you'd like

I am not sure of the solution, but it would be nice with a FAQ entry describing how to best achieve this. It is possible that even some feature should be added into chezmoi to support this, but I'm unsure how.

## Describe alternatives you've considered

An alternative is to use some other project than chezmoi to handle the dotfile management. But since chezmoi does so many things good, it'd be a shame to have to resort to that.

twpayne:
Thanks for the suggestion. I haven't used VS Code Remote Containers myself, but reading through the docs I think you can do the following. I haven't tried it though.

In your dotfiles repo, create the following:

`install.sh`
```sh
#!/bin/sh

curl -sfL https://git.io/chezmoi | sh
bin/chezmoi apply
```

`.chezmoiignore`
```
install.sh
```

Make `install.sh` executable (`chmod a+x install.sh`) and commit it and `.chezmoiignore` to your repo and push the changes.

Then, add the following to your `settings.json`:

```json
{
  "dotfiles.repository": "your-github-id/your-dotfiles-repo",
  "dotfiles.targetPath": "~/.local/share/chezmoi",
  "dotfiles.installCommand": "~/.local/share/chezmoi/install.sh"
}
```

I _think_ this will work. It's possible that VS Code won't create the nested target path. If that's the case, please report here and I'll suggest a workaround.

I'll add this to the how-to once we've worked out how to do it :)

thernstig:
[...]

PR #848, opened 2020-08-12
Enhance Codespaces documentation

felipecassiors:
- Set `sourceDir` so further calls of `chezmoi` will pick-up the right source directory without need to `--source`.

- Install `chezmoi` to a known binary folder, only if it's not installed already.

- A safer installation script (because of `set -e`).

Closes https://github.com/twpayne/chezmoi/issues/760 and closes https://github.com/twpayne/chezmoi/issues/818.

### DIFF
--- a/cmd/docs.gen.go
+++ b/cmd/docs.gen.go
@@ -1333,7 +1333,7 @@ func init() {
 		"The following assumes you are using chezmoi 1.8.4 or later. It does not work\n" +
 		"with earlier versions of chezmoi.\n" +
 		"\n" +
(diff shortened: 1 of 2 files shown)


felipecassiors:
The `install` script can be used anywhere, not only on Codespaces. For example, I always install my dotfiles in a new machine with a singles command:

```bash
sh -c "$(curl -fsSL https://git.io/felipecassiors-dotfiles)"
```
https://github.com/felipecassiors/dotfiles#using-the-convenience-script

Under the hoods, it installs `git` if needed, clones the repository to my preferred location (~/.dotfiles) and then calls the script `install`.

felipecassiors:
There is a bunch of verbosities though, do you think it would be better to take them off the documentation?

felipecassiors:
I don't understand why the CI is complaining...

twpayne:
Thanks very much for this -- these are really nice improvements.

The CI is complaining because it is necessary to run `go generate` and include any changes it makes after making changes to the documentation, see [this section of the contributing guide](https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md#generated-code) for an explanation.

felipecassiors:
My bad, I should have looked at the contribution guidelines. Fixed now, by the way.

twpayne:
Thank you @felipecassiors for this great contribution :)

felipecassiors:
By the way, I used the `devcontainer` to run the `go generate` and it worked quite well.

PR #850, opened 2020-08-16
Allow getting pass pw from pinentry instead of 'unlocking' pass

Grimler91:
Tested for a week on archlinux without any observed side effects.

Here's a gif showing that a password prompt is opened both in a desktop environment and in a terminal without x:
chezmoi

Saves quite a lot of time when `chezmoi apply`ing or `chezmoi diff`ing. 

I have not tested this on mac or windows.

### DIFF
--- a/cmd/secretpass.go
+++ b/cmd/secretpass.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"os/exec"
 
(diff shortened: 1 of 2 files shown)


Grimler91:
I'll link this issue for future reference: https://github.com/twpayne/chezmoi/issues/784

twpayne:
Thanks very much @Grimler91 and thanks for the thumbs up @m-rey.

PR #856, opened 2020-08-17
Remove init --clone flag and only clone repo if needed

### LINKED ISSUE #855: Discussion: is `chezmoi init --clone=false` really needed?
This is not a bug report, feature request, or support request. 

So, `chezmoi` can surely detect if the `source` folder is already "`git initiated`", is there a need for including such an option?

It would help people like me who like to change the `.chezmoi.toml.tmpl` a lot and thus need to run `chezmoi init --clone=false` to apply the changes.

twpayne:
As suggested by @felipecassiors in #855. Fixes #855.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -807,19 +807,18 @@ fi
 # POSIX way to get script's dir: https://stackoverflow.com/a/29834779/12156188
 script_dir="$(cd -P -- "$(dirname -- "$(command -v -- "$0")")" && pwd -P)"
 # exec: replace current process with chezmoi init
-exec $chezmoi init --apply --clone=false --source $script_dir 
+exec "$chezmoi" init --apply "--source=$script_dir"
 ```
 
 Ensure that this file is executable (`chmod a+x install.sh`), and add
 `install.sh` to your `.chezmoiignore` file.
 
 It installs the latest version of chezmoi in `~/.local/bin` if needed, and then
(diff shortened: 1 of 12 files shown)


docs/HOWTO.md
```diff
@@ -807,19 +807,18 @@ fi
 # POSIX way to get script's dir: https://stackoverflow.com/a/29834779/12156188
 script_dir="$(cd -P -- "$(dirname -- "$(command -v -- "$0")")" && pwd -P)"
 # exec: replace current process with chezmoi init
-exec $chezmoi init --apply --clone=false --source $script_dir 
+exec $chezmoi init --apply --source=$script_dir
```
felipecassiors, line 810: I believe you should wrap between quotes to prevent a case where the directory have spaces.

```suggestion
exec "$chezmoi" init --apply "--source=$script_dir"
```

Or:

```suggestion
exec "$chezmoi" init --apply --source "$script_dir"
```

twpayne: Thanks, I've updated the commit with your first suggestion (accepting the suggestion in GitHub would create a new commit, which is not necessary).

felipecassiors: I see. I think this suggestion mechanism is only really useful when you merge commits using squash (which I also do not prefer).

felipecrs:
@twpayne I believe you should release a new version of chezmoi ASAP to prevent people from using the `--clone=false` introduced early.

Issue #858, opened 2020-08-19
https://git.io/chezmoi stopped working


felipecrs:
Problem:

```sh-session
$ curl -L https://git.io/chezmoi
No url found for chezmoi
```

Workaround:

```sh-session
$ curl -L https://raw.githubusercontent.com/twpayne/chezmoi/master/scripts/install.sh
👍
```

twpayne:
Sorry about that, I wanted to update the short url to point to https://github.com/twpayne/chezmoi/blob/master/assets/scripts/install.sh instead of https://github.com/twpayne/chezmoi/blob/master/scripts/install.sh, and there was a bit of downtime as GitHub deleted the link (in the US timezone) so I could re-create it and me creating the link the European time zone.

Anyhow, it's fixed now.

twpayne:
@felipecrs this is now fixed, please use https://github.com/twpayne/chezmoi/blob/master/assets/scripts/install.sh as the work-around.

The reason for updating the location is to keep all assets in the same directory and reduce the number of top level directories.

felipecrs:
@twpayne I wonder, how did you update the URL that your custom short link points to? Every time I try to update mine, (by running the same command I used for creating it) it says `but "existing" already is!`.

twpayne:
You need to ask GitHub to delete the old short URL, you can then re-create it. If an existing short URL already points to the URL you want to shorten then GitHub returns the existing short URL instead of creating a new short URL.

felipecrs:
> You need to ask GitHub to delete the old short URL, you can then re-create it. If an existing short URL already points to the URL you want to shorten then GitHub returns the existing short URL instead of creating a new short URL.

Awesome, thank you very much!

PR #860, opened 2020-08-20
Remove install script from old location

twpayne:
Refs #858.


twpayne:
@felipecrs please tell me when you've updated your scripts to use `https://git.io/chezmoi` or `https://github.com/twpayne/chezmoi/blob/master/assets/scripts/install.sh` so I can merge this.

felipecrs:
Done from my side, thank you so much.

PR #866, opened 2020-08-20
added example for detecting WSL 2

Legion2:
Updated the documentation for detecting WSL according to https://github.com/Microsoft/WSL/issues/423#issuecomment-611086412. In WSL 2 the string `microsoft` is lowercase and in newer versions there is also `-WSL2` in the string to unambiguously identify it.


twpayne:
Thank you :)

PR #878, opened 2020-09-02
Revert "Add work-around for GitHub Codespaces permission mangling"

twpayne:
This reverts commit 5949726fff0fb22d12c25bd206f148b4c92099fc.


twpayne:
The GitHub Codespaces bug is described in https://github.community/t/bug-umask-does-not-seem-to-be-respected/129638 and should be fixed soon.

PR #883, opened 2020-09-06
Fixing onepasswordDetailsFields typo in the HOWTO docs

elazkani:
Insignificant change in the documentation that caused me some time and reading the code to find the missing (s), which was in the previous line. Might save someone else the time.


twpayne:
Thank you very much -- really appreciated :)

PR #890, opened 2020-09-23
Update devcontainer

twpayne:
This is not quite ready yet, it still needs a full work-around for setting the ACL on `$TMPDIR` where chezmoi creates its temporary test directories. See https://github.community/t/bug-umask-does-not-seem-to-be-respected/129638 for discussion.


twpayne:
Superseded by #907.

PR #893, opened 2020-09-28
Statically link Linux binary with musl

### LINKED ISSUE #892: Linux binaries aren't compatible with musl libc systems (dynamic glibc depency)
## Describe the bug

The linux binaries are dynamically linked to glibc. This breaks compatibility with musl libc systems, like voidlinux musl and alpine (the package on `apk` is built correctly).

## To reproduce

1. Try to use a linux build on a musl libc system.

## Expected behavior

The binaries are *truly* statically linked an just work.

## Additional info

Following the build instruction on a musl system works just fine, but I expect those builds to not work on glibc systems.

twpayne:
Fixes #892.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,6 +38,9 @@ jobs:
   test-release:
     runs-on: ubuntu-latest
     steps:
+    - name: Install musl
+      run: |
+        sudo apt install -y musl-tools
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -62,7 +65,7 @@ jobs:
     - name: Test release
       run: |
         # verify that version information is embedded correctly
-        ./dist/chezmoi-cgo_linux_amd64/chezmoi --version | tee /dev/stderr | grep -q "chezmoi version v"
+        file ./dist/chezmoi-cgo_linux_amd64/chezmoi --version | tee /dev/stderr | grep -q "statically linked"
```
Legion2, line 68:
```suggestion
        file ./dist/chezmoi-cgo_linux_amd64/chezmoi | tee /dev/stderr | grep -q "statically linked"
```

twpayne: Good catch, thanks! I was a bit sloppy with cut-and-paste.

Issue #905, opened 2020-10-07
chezmoi fails with userid lookup error


marascio:
## Describe the bug

Running `chezmoi diff` or any other command that performs work (`help` doesn't fail for example) results in the following error: 

```
$ chezmoi diff
chezmoi: user: lookup userid 10001: input/output error
```
## To reproduce

The only special consideration of our environment is that users and groups are defined in LDAP and are not in `/etc/passwd` and `/etc/group`

## Expected behavior

I would expect it to provide the diff of the config to what is currently installed.

## Output of command with `--verbose`

The output does not change, see below.

```
$ bin/chezmoi diff --verbose
chezmoi: user: lookup userid 10001: input/output error
```

## Output of `chezmoi doctor`

```
$ chezmoi doctor
     ok: version 1.8.6, commit cfa3fd2674bc7cd74cbf506f52ebe15723365133, built at 2020-09-28T15:41:30Z, built by goreleaser
     ok: runtime.GOOS linux, runtime.GOARCH amd64
     ok: /home/REDACTED_PATH_TO_DOTFILES (source directory, perm 700)
     ok: /home/REDACTED_HOME (destination directory, perm 755)
     ok: /home/REDACTED_HOME/.config/chezmoi/chezmoi.toml (configuration file)
     ok: /bin/bash (shell)
     ok: /usr/bin/vi (editor)
     ok: /usr/bin/vimdiff (merge command)
     ok: /usr/bin/git (source VCS command, version 2.25.1)
     ok: /usr/bin/gpg (GnuPG, version 2.2.19)
warning: op (1Password CLI, not found)
warning: bw (Bitwarden CLI, not found)
warning: gopass (gopass CLI, not found)
warning: keepassxc-cli (KeePassXC CLI, not found)
warning: lpass (LastPass CLI, not found)
warning: pass (pass CLI, not found)
warning: vault (Vault CLI, not found)
```

twpayne:
chezmoi uses the Go standard library's [`user.Current()`](https://pkg.go.dev/os/user#Current) function to look up your username. It looks like this is returning an error in your case.

Please can you create a file called `user.go` containing:

```go
package main

import (
	"fmt"
	"os/user"
)

func main() {
	fmt.Println(user.Current())
}
```

and then run it with

```
go run user.go
```

and report the output here.

For reference, on my macOS machine, it outputs

```
&{501 20 twp Tom Payne /Users/twp} <nil>
```

twpayne:
[...]

PR #906, opened 2020-10-09
Use glibc for packages

twpayne:


twpayne:
Fixes #905. Refs #893.

PR #912, opened 2020-10-16
Add void linux installation instructions to INSTALL.md

classabbyamp:
The package is in their repositories now (search [here](https://voidlinux.org/packages/) to see).

PR #913, opened 2020-10-16
Add void linux installation instructions to install (hugo docs)

classabbyamp:
Related to #912, see there for details.


twpayne:
Thanks for this, but this is not needed. The website docs are generated with [Hugo](https://gohugo.io) from the Markdown files in the `docs/` directory.

PR #934, opened 2020-11-19
Fix umask on Windows so new files are not created read-only by default.

zb140:
This commit (https://github.com/twpayne/chezmoi/commit/592a363977a83945308018a6b1748647403bed6c) inadvertently made it so that whenever `chezmoi` updates one of my files (on Windows), it's set to "read only".  So if I make some changes to a file, apply it, then make some more changes, the second apply will fail because it can't overwrite the read-only file.

Going back to using `0` for the umask on Windows fixes the issue, but I'm not sure if this change was intentional.  Let me know if you foresee any problems with this and I can look for a different solution.

### DIFF
--- a/internal/chezmoi/chezmoi_windows.go
+++ b/internal/chezmoi/chezmoi_windows.go
@@ -4,7 +4,7 @@ import "os"
 
 // GetUmask returns the umask.
 func GetUmask() os.FileMode {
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks for this and sorry about that!

PR #936, opened 2020-11-20
Opt in to data caching provided by newer versions of 1password-cli

zb140:
The new `--cache` option is provided on all platforms even though no actual caching is done on Windows (🤞 for the future). Also, the option is _not_ automatically added to commands run with `chezmoi secret onepassword` to keep it a pure pass-through.  If people need caching that way, they can simply specify the option themselves.


twpayne (approved):
Perfect, thank you!

PR #940, opened 2020-11-23
Fix fullname for WSL

garyng:
Should be _Windows Subsystem for Linux_ right? (https://docs.microsoft.com/en-us/windows/wsl/install-win10)

Btw, is small changes like this an okay?


twpayne:
Thanks for catching this, and yes small changes like this are very welcome.

Note that documentation changes also require updates to the generated code [as documented in the contributing guide](https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md#generated-code). If you don't have Go to hand then I'm happy to fix this after the PR is merged.

garyng:
Oh! I didn't have Go installed on my system. 
In fact I was reading the documentation and edited the typo right on GitHub, without checking it out locally.
Sorry 😅

PR #951, opened 2020-12-08
Fix GOLANGCI missing version patch

benmezger:
When running `make lint`, it fails locally with the following message:

```
golangci/golangci-lint info checking GitHub for tag 'v1.33'
golangci/golangci-lint crit unable to find 'v1.33' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
```

This PR fixes this issue.


twpayne:
Thank you :)

PR #952, opened 2020-12-08
Check if SourceDir begins with a home dir prefix

benmezger:
Running `chezmoi {appy,doctor}` fails because chezmoi does not expand the home's tilde. This PR check if `SourceDir` has a prefix of `~/` or `~` and expands the home directory accordingly.

Example:

```toml
sourceDir: ~/dotfiles
data:
        ...
```

Before:
```
warning: version dev
     ok: runtime.GOOS darwin, runtime.GOARCH amd64
  ERROR: ~/dotfiles: (source directory, not found)
       : lstat ~/dotfiles: no such file or directory
....
```
After:
```
warning: version dev
     ok: runtime.GOOS darwin, runtime.GOARCH amd64
     ok: /Users/seds/dotfiles (source directory, perm 700)
     ok: /Users/seds (destination directory, perm 755)
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -115,6 +115,15 @@ func init() {
 					"warning: to disable this warning, set sourceVCS.notGit = true in your config file\n",
 				)
 			}
+
+			if strings.HasPrefix(config.SourceDir, "~/") {
+				home, _ := os.UserHomeDir()
+				config.SourceDir = filepath.Join(home, config.SourceDir[2:])
+			} else if strings.HasPrefix(config.SourceDir, "~") {
+				home, _ := os.UserHomeDir()
+				config.SourceDir = filepath.Join(home, config.SourceDir[1:])
+			}
+
 		case os.IsNotExist(err):
 		default:
 			initErr = err
--- a/main_test.go
+++ b/main_test.go
@@ -288,7 +288,7 @@ func cmdUNIX2DOS(ts *testscript.TestScript, neg bool, args []string) {
 		dosData, err := unix2DOS(data)
 		ts.Check(err)
 		//nolint:gosec
-		if err := ioutil.WriteFile(filename, dosData, 0666); err != nil {
+		if err := ioutil.WriteFile(filename, dosData, 0o666); err != nil {
 			ts.Fatalf("%s: %v", filename, err)
 		}
 	}


twpayne:
Tilde expansion is typically the responsibility of the shell, not the program. From looking at the example in the issue, it looks like you want tilde expansion in the config file. What's the exact use case for this that cannot be solved with a full path?

Note that if you want your dotfiles to be in `~/dotfiles` rather than `~/.local/share/chezmoi` you can do the following:

1. Update `.chezmoi.yaml.tmpl` in your source state to contain:

        sourceDir: {{ .chezmoi.sourceDir }}

2. Run `chezmoi init` with the `-S` (`--source-dir`) option:

        chezmoi init -S ~/dotfiles https://github.com/benmezger/dotfiles.git

Note that `sourceDir` in your config file should be an absolute path, not a relative one ([as it is at the moment](https://github.com/benmezger/dotfiles/blob/main/.chezmoi.yaml.tmpl#L1)), as it's interpreted relative to the directory that you run chezmoi from.

A couple of comments on the PR:
* If tilde expansion is added for the source directory, then it should also be added for the destination directory for consistency.
* The code currently ignores any errors returned from `os.UserHomeDir()`.
* The expansion of `~foo` is not correct: `~foo` means "the home directory of user `foo`, not `foo` in the current user's home directory.

Hopefully the update to `.chezmoi.yaml.tmpl` proposed above should be sufficient for now, however.

twpayne:
Thanks for spotting the wayward octal notation by the way. I committed it in b8d2dae6f61d4d5ab14f7df238c6dd7b7482aff4 with you as the author.

zb140:
I know this is only for SourceDir, but FWIW there is a case to be made for `chezmoi` doing tilde expansion for most or all paths, at least on Windows.  Powershell doesn't expand `~` unless you explicitly ask it to by calling `Resolve-Path` or with tab expansion.  So if I run `chezmoi edit ~/.vimrc`, it will fail unless I remember to hit `<TAB>` first -- which I rarely remember to do :wink:

A minor annoyance, but since it's related I thought I'd mention it.

twpayne:
A corner case, consider setting template data in your config file:

```yaml
data:
  someDirectory: "~/dir"
  somethingWithALiteralTildeSlash: "~/somethingElse"
```

Should `~/` at the start of a string always be expanded to the user's home directory? What if the user actually wants a literal `~/`? How can you distinguish between the two?

I think the answer is "in reality, as we're dealing with users' home directories, this is entire reasonable behavior" but I'd like to know if there are any other edge cases.

zb140:
I agree that it seems like reasonable behavior.  The biggest (potentially) problematic assumption that would be built into this change is that strings starting with `~/` are intended to be file paths.  While it's possible that someone might want a string that breaks this assumption, it seems pretty unlikely.

If someone really needed a literal `~/`, could they do something like this:
```yaml
data:
  someDirectory: "~/dir"
  somethingWithALiteralTildeSlash: "\u007E/somethingElse" # U+007E = ~
```
or would the escape get processed too early?

Another option that I'm somewhat more confident would work would be something like this:
```yaml
data:
  tilde: "~" # no slash means no substitution
  someDirectory: "~/dir"
  somethingThatNeedsATilde: "/somethingElse"
```
and then in the template that uses it:
```go
{{ $somethingWIthALiteralTildeSlash := print .tilde .somethingThatNeedsATilde }}
{{/* used to be: {{ .somethingWIthALiteralTildeSlash }} */}}
{{ $somethingWithALiteralTildeSlash }}
```

Kind of clunky, but it should be pretty rare that it's necessary, and if someone finds themselves in this position, it's not the _most_ horrible thing I've ever seen :smile:

twpayne:
Thinking about this further, I think the assumption that all strings in the `data` section are file paths is too strong, and although strings are unlikely to begin with `~/`, having chezmoi always expand this to `$HOME` is too weird of a corner case for too little gain and risks causing extreme confusion later.

However, when it's known that the string is a path, namely for the paths to the configuration file, source directory, destination directory, and targets, I think it's OK that chezmoi does tilde expansion in case the shell didn't.

Rather than shoe-horning this into the current `master`, it's straightforward to add consistently to `v2` as v2 has a consistent mechanism for handling external paths (`OSPath`).

zb140:
That sounds like a great compromise.  I hadn't realized it would be reasonably easy to do source dir etc without involving the data section.  Adding expansion for target paths is enough to solve the annoyance I encountered, and handling source dir and friends covers the original intent of the PR.  In the future, if someone encounters a need to expand `~/` somewhere else (like in a data string), we could always add a `resolveHome` template function to do it explicitly.

benmezger:
Tom,

Regarding `sourceDir: {{ .chezmoi.sourceDir }}`: thanks, this will solve
the issue I am having with Chezmoi.

Instead of using an absolute path, I was executing Chezmoi like this
[1]. Using a relative path was working fine, however, the issue I was
having is when I ran `chezmoid add <path>` outside the home directory.
Chezmoi was creating a `~` directory in my `$HOME`, due to `sourceDir`
being `~/dotfiles` in my `chezmoi.yaml`.




[1] -
<https://github.com/benmezger/dotfiles/blob/77338034ad052320e65ca3dcf2f810c9293089de/roles/common/tasks/main.yml#L23>

benmezger:
As expected, .chezmoi.sourceDir worked fine. I won't close this PR as
there's some discussion going on, but feel free to close it. In case
this PR is still required, I would be happy to update with the comments
pointed out.

twpayne:
Thanks, I've moved the tracking of tilde expansion to #963 and so will close this now. Thanks again for the proposal and discussion :)

Issue #987, opened 2021-01-17
chezmoi2: Contributors, please help with beta testing


twpayne:
With #654 merged, chezmoi version 2 is now ready for beta testing. I'd really appreciate it if you could test it on your machines. At the moment, this requires you need to be familiar with Go development to build and install it locally. Packages, snaps, etc. will follow later.

Quick notes:
* You should not need to change anything in your source state. Note that when you run `chezmoi2 apply` for the first time your `run_once_` scripts will run again.
* The goal of chezmoi2 is to enable some new features (e.g. `chezmoi2 status`) and fix various corner-case bugs that were hard to fix with chezmoi1's architecture. Even though chezmoi2 is a complete re-write internally, to the end user, chezmoi2 is not radically different to chezmoi1.
* Read [a rough draft of what's changed](https://github.com/twpayne/chezmoi/blob/master/chezmoi2/docs/CHANGES.md).
* See the [v2.0.0 milestone](https://github.com/twpayne/chezmoi/issues?q=is%3Aissue+milestone%3Av2.0.0) for individual issues.
* Right now it's in a separate subdirectory, install it by checking out the `master` branch and running `go install ./chezmoi2`.
* It should be safe to use, but please use `chezmoi2 diff` and/or use the `--dry-run` option to check changes before you apply them.
* The separate `chezmoi2` directory and command name are temporary: eventually chezmoi2 will replace chezmoi1.

Specific ways to help:
* Please install and run chezmoi2 and report anything that seems strange or suspicious *in this issue*, please don't open new issues yet.
* There are a lot of integration tests in [`chezmoi2/testdata/scripts`](https://github.com/twpayne/chezmoi/tree/master/chezmoi2/testdata/scripts). More tests are very welcome!
* Please prefix PRs that affect chezmoi2 with the string `chezmoi2:`.

What happens next:
* chezmoi2 gets tested a bit more, hopefully by you, to shake out the initial bugs.
* Once chezmoi2 seems mostly stable, I'll update the release processes to produce beta releases for wider testing.
* Once the beta has been more widely tested, chezmoi2 will replace chezmoi.

cc @benmezger @Grimler91 @felipecrs @fwSmit @Legion2 @zb140

zb140:
I've installed `chezmoi2` and things seem to be going pretty well.  Here's a couple of things I've noticed after my first `chezmoi2 apply`:
[...]

PR #989, opened 2021-01-17
chezmoi2: Restore secret keyring command

twpayne:
Refs #987.

As [reported by](https://github.com/twpayne/chezmoi/issues/987#issuecomment-761884922) @zb140.

> Is the `chezmoi secret` subcommand permanently gone?

No :) I just didn't realize that it was actively used. This PR restores it for `secret keyring`. For other password managers, I think that the password manager's CLI should be a better option than using `chezmoi secret` but let's see what users report.

> I've been using it as part of my secret caching scheme to work around the perf issues in 1password. Specifically, I'm caching secrets in keepass and storing the keepass db password in the OS keyring (which gets unlocked when I sign into the computer). Then I use a shell alias to do `chezmoi secret keyring get --service secret_cache_pw --user zb | chezmoi apply` to apply dotfile changes. I'm sure I can find a workaround if needed.

This PR should fix this, but there's a deeper issue here that this PR doesn't address. chezmoi2 tries to read passwords from a TTY, and will get one if necessary. It so happens that I don't know how to do this on Windows yet (see https://github.com/twpayne/chezmoi/blob/master/chezmoi2/cmd/config.go#L634-L651) so on Windows chezmoi2 will happily read a password from stdin. The long term fix is probably to (a) read passwords from a TTY on Windows and (b) add a flag to allow passwords to be read from stdin if needed.

@zb140 would you mind testing this PR? Please do merge it or push fixes to this branch if needed.

### DIFF
--- a/chezmoi2/cmd/config.go
+++ b/chezmoi2/cmd/config.go
@@ -111,6 +111,7 @@ type Config struct {
 	init            initCmdConfig
 	managed         managedCmdConfig
 	purge           purgeCmdConfig
+	secretKeyring   secretKeyringCmdConfig
 	status          statusCmdConfig
 	update          updateCmdConfig
 	verify          verifyCmdConfig
@@ -784,6 +785,7 @@ func (c *Config) newRootCmd() (*cobra.Command, error) {
 		c.newMergeCmd,
 		c.newPurgeCmd,
 		c.newRemoveCmd,
+		c.newSecretCmd,
 		c.newSourcePathCmd,
 		c.newStateCmd,
 		c.newStatusCmd,
--- a/chezmoi2/cmd/secretcmd.go
+++ b/chezmoi2/cmd/secretcmd.go
@@ -1,0 +1,17 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+func (c *Config) newSecretCmd() *cobra.Command {
+	secretCmd := &cobra.Command{
+		Use:     "secret",
+		Args:    cobra.NoArgs,
+		Short:   "Interact with a secret manager",
+		Long:    mustLongHelp("secret"),
+		Example: example("secret"),
+	}
+
+	secretCmd.AddCommand(c.newSecretKeyringCmd())
+
+	return secretCmd
+}
--- a/chezmoi2/cmd/secretkeyringcmd.go
+++ b/chezmoi2/cmd/secretkeyringcmd.go
@@ -22,6 +22,17 @@ func (c *Config) newSecretKeyringCmd() *cobra.Command {
 		Short: "Interact with keyring",
 	}
 
+	persistentFlags := keyringCmd.PersistentFlags()
+	persistentFlags.StringVar(&c.secretKeyring.service, "service", "", "service")
+	if err := keyringCmd.MarkPersistentFlagRequired("service"); err != nil {
+		panic(err)
+	}
+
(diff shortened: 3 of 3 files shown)


zb140:
I added the `--service` and `--user` args back in, and it's working great now.  @twpayne I couldn't find another example of required args in chezmoi2.  Is this (https://github.com/twpayne/chezmoi/blob/secret-keyring/chezmoi2/cmd/secretkeyringcmd.go#L27) an appropriate pattern?  I pretty much just copied it from chezmoi1, except that `panicIfError` doesn't exist anymore.

twpayne:
That's great, thanks! I'll add a helper for catching errors from `MarkPersistentFlagRequired` in a separate PR as it's a common pattern.

zb140:
@twpayne Excellent!

For the TTY stuff, console apps on Windows can be... complicated :smile:  (Come to think of it, so many parts of the system are called "console" that it's kind of complicated to even _talk_ about...)  When a console app is launched, the OS will create a console handle (if the app didn't inherit one from the command line that launched it) and connect it to stdin.  The app is then free to call Console APIs if it wants.

If the primary goal is just to enforce that prompts should only go to a real terminal (unless overridden by the new flag you mentioned), then I think it should be sufficient to make `getTTY` do something like this:
```go
if runtime.GOOS == "windows" {
	if stdin, ok := c.stdin.(*os.File); !ok || !term.IsTerminal(int(stdin.Fd())) {
		return nil, nil, fmt.Errorf("can't read from non-terminal")
	}
	c.ttyReader = bufio.NewReader(c.stdin)
	c.ttyWriter = c.stdout
}
```
`term.IsTerminal` does return the correct result using the Windows Console APIs.

If it's more complicated than that, then it will probably be necessary to go directly to the Console APIs.  Let's cross that bridge if we come to it :smile:

twpayne:
Thanks for all the info on the terminal stuff. https://github.com/twpayne/chezmoi/pull/990/commits/0887458a637be1d10e0c85aa028e168b473e938f should fix it. It uses `golang.org/x/term` which (fingers crossed) also includes Windows terminal support. With this change, chezmoi2 will try to read input from a terminal by default. If you want to read from stdin (as in your `chezmoi secret keyring get | chezmoi apply` case) then you'll need to pass the `--no-tty` flag to `chezmoi apply`.

zb140:
Yep, that mostly works.  I tried both the terminal and `--no-tty` variants and was able to enter the password and have it work.

The one issue I encountered was that my computer doesn't actually print the prompt ("Insert password to unlock...") even though it does listen for and accept input.  After some tracing through the `x/term` code, I eventually tracked it down to `x/term`'s call to `Write` on the terminal's `ReadWriter` was just not doing anything.  I replaced the stdin file with a `ReadWriter` composite of stdin and stdout (like this: https://github.com/zb140/chezmoi/commit/4f14e882044a755a5bf98bc2582e9313cf59119b) and I get a working prompt and working input.  I also ran all the tests and they passed.  But there still may be some other consequence of the change I haven't thought of yet.

Side note: I have a hunch that this behavior of the prompt not working might actually be machine-dependent rather than simply a blanket "it doesn't work on Windows".  I didn't have time to pull on that thread tonight but I should be able to do some digging tomorrow.

twpayne:
https://github.com/zb140/chezmoi/commit/4f14e882044a755a5bf98bc2582e9313cf59119b looks good to me - I was going to suggest the same.

PR #995, opened 2021-01-20
chezmoi2: Fix terminal prompt on Windows

zb140:
* Make sure the new terminal has a Writer and a Reader


zb140:
Oops, I didn't see this was already taken care of in #993 before submitting it.

PR #996, opened 2021-01-20
chezmoi2: Update the edit testscript for Windows

zb140:
* Also add some script tests


twpayne:
Brilliant, thank you!

PR #997, opened 2021-01-20
docs(REFERENCE): clarify .chezmoiignore patterns

9999years:
Docs incorrectly state that the pattern

    backups/**

in a `.chezmoiignore` file would ignore `backups` and all its contents recursively; the pattern ignores all the contents of `backups` but not the directory itself. (As a consequence, `chezmoi add` would add `backups` with a `.keep` file in it, but not any of the files or folders in `backups` itself.)

This behavior comes from the `doublestar` library, which has [a test ensuring that `**` patterns work this way](https://github.com/bmatcuk/doublestar/blob/v1.3.4/doublestar_test.go#L87), so this is presumably the intended behavior.


twpayne:
Thanks very much @9999years -- much appreciated!

PR #1008, opened 2021-01-24
Fix: Repair devcontainer file

JustinGrote:
Fixes the devcontainer and removes unnecessary items


twpayne:
Thanks for this. At the moment this PR breaks a few things:

1. GitHub Codespaces's support for umask is not correct, which in turn causes chezmoi's tests to fail. A workaround is needed, and this PR removes some of those workaround steps (notably installing the `acl` package). See [this thread](https://github.community/t/bug-umask-does-not-seem-to-be-respected/129638) for background information. Note that the ACL fix needs to be applied to the temporary directory where chezmoi creates its temporary test filesysystems, it's not sufficient to apply the fix to the current directory only.
2. The PR removes the installation of a few VSCode plugins helpful to developers.

I've been holding off supporting CodeSpaces properly until GitHub fix point 1.

JustinGrote:
@twpayne I didn't remove any of the extensions, they are still there.
image

and I didn't remove the acl
image


Maybe double-check the diff for the devcontainer dockerfile?
https://github.com/twpayne/chezmoi/pull/1008/files#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6

twpayne:
My apologies @JustinGrote - you're right and I was wrong - the extensions and acl package installation are still there. Sorry about that.

PR #1009, opened 2021-01-24
CI: Add artifact capture "nightlies" to Github Actions

JustinGrote:
This PR will capture the binaries to artifacts so that every commit has an available downloadable build to test, allowing rapid testing of new features without needing a development environment.

Currently capturing both Chezmoi1 and Chezmoi2

https://github.com/twpayne/chezmoi/pull/1009/checks?check_run_id=1756221264#step:9:15
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,12 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         name: chezmoi2-${{ matrix.os }}
-        path: dist/chezmoi*
+        path: dist/chezmoi2
+    - name: Artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: chezmoi-${{ matrix.os }}
+        path: dist/chezmoi
   test-release:
     runs-on: ubuntu-18.04
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,12 +49,16 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         name: chezmoi2-${{ matrix.os }}
-        path: dist/chezmoi2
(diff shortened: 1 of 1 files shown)


JustinGrote:
@twpayne should be good to go.

twpayne:
Thanks very much for this. The GitHub Action in this PR currently distributes the binaries built as part of the `test (ubuntu-18.04)` step. Could you instead update the PR to upload artifacts from the `test-release` step? You'll find all the files in the `dist/` subdirectory once the step is completed. You can simulate this locally by installing [goreleaser](https://goreleaser.com/) and running `make test-release`.

The binaries built this way have a number of advantages:
* More accurate version information (so `chezmoi --version` is meaningful, critical for debugging user-reported problems).
* More OSes and architectures supported.
* musl libc support (needed for musl-based Linux distributions like Alpine).

The binaries themselves end up in `dist/$BUILDID_$GOOS_$GOARCH/`. It's probably worth creating separate build artifacts per directory. The full `dist/` directory weighs in at about half a gig zipped, which is a bit much when all you want is a 30MB binary :)

JustinGrote:
The reason I chose those builds was because they can take advantage of the matrix build and be available within a minute. Since the goreleaser build takes about 10 minutes it takes much longer for artifacts to become available, and the intention of these was for nightly-style "fast tests" of new functionality, rather than more proper prerelease or rc candidates.

Regardless I have adjusted as asked. Since the sheer number of artifact types, creating a matrix of uploads would slow down the pipeline drastically, so I narrowed it to just the most popular binaries that would be typically used for rapid testing.
image

@twpayne ready for re-review

twpayne:
Thank you very much @JustinGrote!

PR #1012, opened 2021-01-25
Fix: State Version Check Fails with Artifacts

JustinGrote:
Current snapshot artifacts fail with:
`chezmoi: source state requires version 0.0.0 or later, chezmoi is version 0.0.0-SNAPSHOT-3bc8688`

Removing -snapshot will generate normal versions.


twpayne:
Thanks!

JustinGrote:
@twpayne I was wrong, it doesn't generate normal versions if the commit isn't tagged, I'll recreate this as an issue.

PR #1013, opened 2021-01-25
chezmoi2: Build cleanups

twpayne:
cc @JustinGrote as you might be modifying the same files at the moment.


JustinGrote:
No active changes from me right now, probably safe to merge.

PR #1014, opened 2021-01-25
Only check minimum version in non-snapshot builds

twpayne:
@JustinGrote what do you think of this approach?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,7 @@ jobs:
       uses: goreleaser/goreleaser-action@v2
       with:
         version: latest
-        args: release --skip-publish
+        args: release --skip-publish --snapshot
     - name: Test release
       run: |
         # verify that version information is embedded correctly
--- a/chezmoi2/cmd/config.go
+++ b/chezmoi2/cmd/config.go
@@ -1146,7 +1146,7 @@ func (c *Config) sourceState() (*chezmoi.SourceState, error) {
 		return nil, err
 	}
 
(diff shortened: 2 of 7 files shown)


JustinGrote:
I'll test the binary to see if it works, but I think even with a snapshot build its still technically a semver because it has a tag. And tagged semvers are considered less than the "release" version with no tags which is why the state check isn't working.

twpayne:
Note that this the failure observed:

    chezmoi: source state requires version 0.0.0 or later, chezmoi is version 0.0.0-SNAPSHOT-3bc8688

is chezmoi's own internal version check (https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md#chezmoiversion).

JustinGrote:
Tested as working on windows, LGTM.
image

twpayne:
The latest commit in this PR is `b275e0a740a3e9c07dbdedd2416c6bb774cbe984`. The version printed in your output begins with `5745fa3`. Are you sure you're testing the correct version?

JustinGrote:
Yep, it's the merge of your PR and master done by the checkout action:
https://github.com/twpayne/chezmoi/pull/1014/checks?check_run_id=1764455798#step:5:462

twpayne:
Thanks for the testing @JustinGrote. I missed that the version number `0.0.0-SNAPSHOT-hex` would have non-zero metadata and so was not equal to the zero [semver.Version](https://pkg.go.dev/github.com/coreos/go-semver/semver#Version) value. This is fixed now. Thank you for testing and please do report further problems!

PR #1016, opened 2021-01-28
chezmoi2: Address more feedback

twpayne:


twpayne:
Thanks to @JustinGrote and @fwSmit for their feedback here - very much appreciated!

PR #1019, opened 2021-02-02
Add BitwardenAttachment template support

benmezger:
This "mimics" `onepasswordDocument` for Bitwarden. I found an "issue" with Bitwarden while implementing this feature. Basically Bitwarden's file URL is build upon the `ID` + `itemid`, which makes anyone who exposes both `ids` to expose their files. 

Here is an example of the payload (with dummy values), pay attention to `id` and `id` in `attachments`, then the `url`:

```
[
  ...
  {
    "object": "item",
    "id": "98ebec3b-3a9f-4220-9ae2-",
    "organizationId": null,
    "folderId": "C6BE41BF-A86F-455D-B9EB-E36F6E2D37E2",
    "type": 2,
    "name": ".ssh/id_rsa.pub",
    "notes": null,
    "favorite": false,
    "secureNote": {
      "type": 0
    },
    "collectionIds": [],
    "attachments": [
      {
        "id": "arandom-id",
        "fileName": "id_rsa.pub",
        "size": "625",
        "sizeName": "625 Bytes",
        "url": "https://cdn.bitwarden.net/attachments/C6BE41BF-A86F-455D-B9EB-E36F6E2D37E2/arandomid"
      }
    ],
    "revisionDate": "2021-02-02T15:28:04.683Z"
  }
]
```

This can be avoided **IF** the user actually reads the docs... But I am still a bit afraid someone uses something like the following:

```
    {{- (bitwardenAttachment "C6BE41BF-A86F-455D-B9EB-E36F6E2D37E2" "random-id") -}}
```
Which exposes the file. We would for example, check whether the user is using two ids instead of a `filename` and return an error/warning. But I would like to review any other options. 

I've commited the docs and the code altogether, which is easier when reverting, but I could separate them if required.

Regarding `chezmoi2`, I was unsure whether I should move this to `chezmoi2` or wait a bit more.

### DIFF
--- a/cmd/secretbitwarden.go
+++ b/cmd/secretbitwarden.go
@@ -29,6 +29,7 @@ func init() {
 	config.Bitwarden.Command = "bw"
 	config.addTemplateFunc("bitwarden", config.bitwardenFunc)
 	config.addTemplateFunc("bitwardenFields", config.bitwardenFieldsFunc)
+	config.addTemplateFunc("bitwardenAttachment", config.bitwardenAttachmentFunc)
 
 	secretCmd.AddCommand(bitwardenCmd)
 }
@@ -80,4 +81,8 @@ func (c *Config) bitwardenFieldsFunc(args ...string) map[string]interface{} {
 		}
 	}
 	return result
+}
+
+func (c *Config) bitwardenAttachmentFunc(name, itemID string) string {
+	return string(c.bitwardenOutput([]string{"get", "attachment", name, "--itemid", itemID, "--raw"}))
 }
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -1079,10 +1079,7 @@ using the [Bitwarden CLI](https://bitwarden.com/help/article/cli/) (`bw`). _file
 and _itemid_ is passed to `bw get attachment <filename> --itemid <itemid>`
 and the output from `bw` is returned. The output from `bw` is cached so calling
 `bitwardenAttachment` multiple times with the same _filename_ and _itemid_ will only
-invoke `bw` once. _Beware_ that using the `id` instead of a filename, will expose your
-file publicly, as the `id` + `itemid` is built into the following URL in Bitwarden:
-`"https://cdn.bitwarden.net/attachments/<id>/<itemid>"`, which allows anyone to download
-your files.
+invoke `bw` once.
 
 #### `bitwardenAttachment` examples
 
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -69,6 +69,7 @@ Manage your dotfiles securely across multiple machines.
 * [Template functions](#template-functions)
   * [`bitwarden` [*args*]](#bitwarden-args)
   * [`bitwardenFields` [*args*]](#bitwardenfields-args)
+  * [`bitwardenAttachment` [*args*]](#bitwardenattachment-args)
   * [`gopass` *gopass-name*](#gopass-gopass-name)
   * [`include` *filename*](#include-filename)
   * [`ioreg`](#ioreg)
@@ -1070,6 +1071,22 @@ same arguments will only invoke `bw` once.
 #### `bitwardenFields` examples
 
     {{ (bitwardenFields "item" "example.com").token.value }}
+
+### `bitwardenAttachment` _filename_ _itemid_
+
+`bitwardenAttachment` returns a document from [Bitwarden](https://bitwarden.com/)
+using the [Bitwarden CLI](https://bitwarden.com/help/article/cli/) (`bw`). _filename_
+and _itemid_ is passed to `bw get attachment <filename> --itemid <itemid>`
+and the output from `bw` is returned. The output from `bw` is cached so calling
+`bitwardenAttachment` multiple times with the same _filename_ and _itemid_ will only
+invoke `bw` once. _Beware_ that using the `id` instead of a filename, will expose your
+file publicly, as the `id` + `itemid` is built into the following URL in Bitwarden:
+`"https://cdn.bitwarden.net/attachments/<id>/<itemid>"`, which allows anyone to download
+your files.
+
+#### `bitwardenAttachment` examples
+
+    {{- (bitwardenAttachment "<filename>" "<itemid>") -}}
 
 ### `gopass` *gopass-name*
 
--- a/testdata/scripts/bitwarden.txt
+++ b/testdata/scripts/bitwarden.txt
@@ -9,6 +9,11 @@ stdout password-value
 chezmoi execute-template '{{ (bitwardenFields "item" "example.com").Hidden.value }}'
 stdout hidden-value
 
+[short] stop
+
+chezmoi execute-template '{{ (bitwardenAttachment "filename" "item-id") }}'
+stdout hidden-file-value
+
 -- bin/bw --
 #!/bin/sh
 
@@ -45,6 +50,11 @@ case "$*" in
   "collectionIds": [],
   "revisionDate": "2020-10-28T00:21:02.690Z"
 }
+EOF
+   ;;
+"get attachment filename --itemid item-id --raw")
+    cat <<EOF
+hidden-file-value
 EOF
     ;;
 *)
@@ -85,6 +95,8 @@ IF "%*" == "get item example.com" (
     echo.  "collectionIds": [],
     echo.  "revisionDate": "2020-10-28T00:21:02.690Z"
     echo.}
+) ELSE IF "%*" == "get attachment filename --itemid item-id --raw" (
+    echo. hidden-file-value
 ) ELSE (
     echo Invalid command: $*
     echo "See --help for a list of available commands."
--- a/cmd/docs.gen.go
+++ b/cmd/docs.gen.go
@@ -2009,6 +2009,7 @@ func init() {
 		"* [Template functions](#template-functions)\n" +
 		"  * [`bitwarden` [*args*]](#bitwarden-args)\n" +
 		"  * [`bitwardenFields` [*args*]](#bitwardenfields-args)\n" +
+		"  * [`bitwardenAttachment` [*args*]](#bitwardenattachment-args)\n" +
 		"  * [`gopass` *gopass-name*](#gopass-gopass-name)\n" +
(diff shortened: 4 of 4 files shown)


cmd/secretbitwarden.go
```diff
@@ -81,3 +82,13 @@ func (c *Config) bitwardenFieldsFunc(args ...string) map[string]interface{} {
 	}
 	return result
 }
+
+func (c *Config) bitwardenAttachmentFunc(args ...string) string {
```
twpayne, line 86: You can use

```go
func (c *Config) bitwardenAttachmentFunc(name, itemID string) string
```

here to avoid having to parse `args`.

docs/REFERENCE.md
```diff
@@ -3,141 +3,143 @@
...
-  * [`onepasswordDetailsFields` *uuid* [*vault-uuid*]](#onepassworddetailsfields-uuid-vault-uuid)
-  * [`pass` *pass-name*](#pass-pass-name)
-  * [`promptBool` *prompt*](#promptbool-prompt)
-  * [`promptInt` *prompt*](#promptint-prompt)
-  * [`promptString` *prompt*](#promptstring-prompt)
-  * [`secret` [*args*]](#secret-args)
-  * [`secretJSON` [*args*]](#secretjson-args)
-  * [`stat` *name*](#stat-name)
-  * [`vault` *key*](#vault-key)
+
```
twpayne, line 6: It looks like something reformatted the entire `REFERENCE.md` file. Could you update the commit to only include the relevant changes?

twpayne (changes requested):
Thanks for this! It's OK to put this in chezmoi1 for now, although I'd also accept a PR for chezmoi2 :)

Could you also please add a test to [`scripts/testdata/bitwarden.txt`](https://github.com/twpayne/chezmoi/blob/master/testdata/scripts/bitwarden.txt)?

docs/REFERENCE.md
benmezger: Sure, I found it odd too, but I thought that was the normal behaviour. Will update accordingly.

benmezger:
@twpayne Another problem I've found is that if you have 2 attachments in one, you will **need** to use both of the ids, as for some reason `bw` does not find it when passing the filename =(.

docs/REFERENCE.md
benmezger: Auch, that was my editor ;')

twpayne:
Couple more questions on this as I don't use Bitwarden myself:

It looks like [Bitwarden attachments](https://bitwarden.com/help/article/attachments/) are a paid-only feature, so I currently can't test any of this myself. I'll ask them for a test account.

> I found an "issue" with Bitwarden while implementing this feature. Basically Bitwarden's file URL is build upon the `ID` + `itemid`, which makes anyone who exposes both `ids` to expose their files.

Just to check that I understand this correctly, is the following statement true? "Anyone who knows the `ID` and `itemid` of a Bitwarden attachment can download the attachment without authentication, i.e. `https://cdn.bitwarden.net/attachments/<ID>/<itemid>` is unprotected." 😱 

How about we change the API so that these two IDs are not exposed? I see in the output of `bw get --help` the example:

```
bw get attachment photo.jpg --itemid 99ee88d2-6046-4ea7-92c2-acac464b1412 --raw
```

...which looks like we can use a filename instead, which also gives a friendlier API to the user. Can we use `item` and `filename` as the arguments to `bitwardendAttachment` instead?

benmezger:
> Just to check that I understand this correctly, is the following statement true? "Anyone who knows the `ID` and `itemid` of a Bitwarden attachment can download the attachment without authentication, i.e. `https://cdn.bitwarden.net/attachments/<ID>/<itemid>` is unprotected." 😱

That is correct (I've tested this on an private tab, but I've created a dummy file, and I am sharing with you the payload I get back from `bw`. If you are able to download this file (see url in attachments), my assumption is correct, whoever has **both** the ids (the item id + the attachment id) they are able to build the URL with this data. 

> How about we change the API so that these two IDs are not exposed? I see in the output of bw get --help the example:

This approach **only** works when you have 1 file in your item, if you have more, for some reason is it not able to return the attachment, instead, it returns you the attachment IDs to `stderr`. I've tested this, because I had an `bw` item with 2 attachments: public/private ssh key. I had to create two items, one for each.

```
[
  {
    "object": "item",
    "id": "11a87565-3cdc-4c7b-b852-acc3012a0b61",
    "organizationId": null,
    "folderId": null,
    "type": 2,
    "name": "Test note",
    "notes": null,
    "favorite": false,
    "secureNote": {
      "type": 0
    },
    "collectionIds": [],
    "attachments": [
      {
        "id": "75acyihsmg1yufi8dvphyj3rnbgoa7uu",
        "fileName": "hello.txt",
        "size": "97",
        "sizeName": "97 Bytes",
        "url": "https://cdn.bitwarden.net/attachments/11a87565-3cdc-4c7b-b852-acc3012a0b61/75acyihsmg1yufi8dvphyj3rnbgoa7uu"
      }
    ],
    "revisionDate": "2021-02-02T18:05:08.796Z"
  }
]
```

twpayne:
> That is correct (I've tested this on an private tab, but I've created a dummy file, and I am sharing with you the payload I get back from `bw`. If you are able to download this file (see url in attachments), my assumption is correct, whoever has **both** the ids (the item id + the attachment id) they are able to build the URL with this data.

OK, I can download the file but the contents are binary (probably encrypted):

```
$ wget https://cdn.bitwarden.net/attachments/11a87565-3cdc-4c7b-b852-acc3012a0b61/75acyihsmg1yufi8dvphyj3rnbgoa7uu
--2021-02-02 19:11:30--  https://cdn.bitwarden.net/attachments/11a87565-3cdc-4c7b-b852-acc3012a0b61/75acyihsmg1yufi8dvphyj3rnbgoa7uu
Resolving cdn.bitwarden.net (cdn.bitwarden.net)... 152.199.21.175
Connecting to cdn.bitwarden.net (cdn.bitwarden.net)|152.199.21.175|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 97 [application/octet-stream]
Saving to: ‘75acyihsmg1yufi8dvphyj3rnbgoa7uu’

75acyihsmg1yufi8dvphyj3rnbgoa7uu             100%[============================================================================================>]      97  --.-KB/s    in 0s

2021-02-02 19:11:31 (2.72 MB/s) - ‘75acyihsmg1yufi8dvphyj3rnbgoa7uu’ saved [97/97]

$ ls -l 75acyihsmg1yufi8dvphyj3rnbgoa7uu
-rw-r--r--  1 twp  staff  97 Feb  2 19:05 75acyihsmg1yufi8dvphyj3rnbgoa7uu

$ xxd 75acyihsmg1yufi8dvphyj3rnbgoa7uu
00000000: 029c 70db a695 2de4 999f 7a01 1052 5914  ..p...-...z..RY.
00000010: 36a4 cd53 5310 6a59 27cf b9ea 9a72 5e68  6..SS.jY'....r^h
00000020: dc2a 6afc f477 ad64 a4e1 a176 06ac 99aa  .*j..w.d...v....
00000030: 1915 9814 c2b1 2cdd 3e53 d2cb d3f6 dafb  ......,.>S......
00000040: 1c48 e103 ff2d f86f 88e6 e190 f798 b088  .H...-.o........
00000050: 23c2 bf56 3f70 ebe2 833b 87fa faa6 4c3d  #..V?p...;....L=
00000060: 50                                       P
```

> This approach **only** works when you have 1 file in your item, if you have more, for some reason is it not able to return the attachment, instead, it returns you the attachment IDs to `stderr`. I've tested this, because I had an `bw` item with 2 attachments: public/private ssh key. I had to create two items, one for each.

My gut feeling is that we should let Bitwarden fix this before adding the `bitwardenAttachment` function to chezmoi. Using a filename is such a nicer interface than having to extract IDs. Your thoughts?

twpayne:
> My gut feeling is that we should let Bitwarden fix this before adding the `bitwardenAttachment` function to chezmoi. Using a filename is such a nicer interface than having to extract IDs. Your thoughts?

Actually, I suppose the function as-is is OK: the passed `name` can either be a filename (if there's only one file) or an ID (if there's more than one file), so no changes are needed to chezmoi when/if Bitwarden fix the find-by-filename bug.

testdata/scripts/bitwarden.txt
```diff
@@ -45,6 +50,11 @@ case "$*" in
   "collectionIds": [],
   "revisionDate": "2020-10-28T00:21:02.690Z"
 }
+EOF
+   ;;
+"get attachment filename --itemid item-id --raw")
+    cat <<EOF
+hidden-file-value
```
twpayne, line 57: A similar change will be needed to `bin/bw.cmd` below in the same file for Windows. Sorry.

benmezger:
> OK, I can download the file but the contents are binary (probably encrypted):

Good, the must have been my cache. 

> My gut feeling is that we should let Bitwarden fix this before adding the bitwardenAttachment function to chezmoi. Using a filename is such a nicer interface than having to extract IDs. Your thoughts?

I will open an issue at their repository asking if there is any possibilities of handling this when multiple attachments are set. Because at the moment, if we loop over these attachments, we would have to curl or like to download the file itself (due to the URL), as a `--raw` won't work in this case. 

> Actually, I suppose the function as-is is OK: the passed name can either be a filename (if there's only one file) or an ID (if there's more than one file), so no changes are needed to chezmoi when/if Bitwarden fix the find-by-filename bug.

Yes, I share the same opinion. I have updated the test files, and pushing the updated docs now. If you are interested, I would like to make the same changes to `chezmoi2`.

benmezger:
Also, let me know if you want these commits all fixedup.

benmezger:
@twpayne I figured out the issue regarding multiple files. It does actually work having multiple attachments in one item, however, my setup fails, because I have two filenames with the following names: `id_rsa` and `id_rsa.pub`, it seems that Bitwarden gets confused on which one to return.

```
bw get attachment "id_rsa.pub" --itemid <item-id>--raw  # this works
....
bw get attachment "id_rsa" --itemid <item-id>--raw  # does not work, returns:
More than one result was found. Try getting a specific object by `id` instead. The following objects were found:
<id-one>
<id-two>
```

twpayne:
Many thanks @benmezger!

PR #1021, opened 2021-02-03
Tweak bitwarden template functions and documentation

twpayne:


chezmoi2/cmd/bitwardentemplatefuncs.go
```diff
@@ -66,6 +66,6 @@ func (c *Config) bitwardenTemplateFunc(args ...string) map[string]interface{} {
 	return data
 }
 
-func (c *Config) bitwardenAttachmentTemplateFunc(name, itemID string) string {
-	return string(c.bitwardenOutput([]string{"attachment", name, "--itemid", itemID, "--raw"}))
```
twpayne, line 70: @benmezger Apologies for changing this back - I know I said call the variable `itemID` in my review of #1019.

The reason for this change is for the case of the variable name to match the lowercase used by Bitwarden, rather the idiomatic `itemID` used in Go code. When you have two different domains with different conventions, it's always a judgement call where to change from one convention to another. In the specific case of chezmoi's password manager integration, I decided to follow the password manager's convention instead of Go's convention on the assumption that the users of password managers would be more familiar with the password manager than chezmoi.

benmezger: @twpayne no apologies needed! Thanks for the changes! keeping the password convention makes sense in this context. I find easy to read when reading the command line while coding the integration, as you can easily follow along without swapping cases around when comparing and/or copying.

Issue #854, opened 2020-08-17
Add `--no-scripts` and `--re-run-scripts` to `chezmoi apply`


felipecrs:
I searched and I could not find anything about this.

## Describe the solution you'd like

To add `--no-scripts` and `--re-run-scripts` to `chezmoi apply`, which would respectively `apply` but not run scripts, and `apply` and re-run the scripts already marked as ran.

It would be also good if we could have those options available on `chezmoi init --apply` as well.

## Describe alternatives you've considered

There is no current way to prevent the scripts to execute I think. To make a script already marked as ran to run again I have to change its contents, so `chezmoi` will run it. (I'm talking about the scripts `run_once` and not just `run`.

twpayne:
This is a great suggestion, thanks.

`--no-scripts` is implemented in the `v2` branch (#654) which adds an `--include` flag that allows you to specify what types of entry to apply. I hope to finish `v2` in the next few weeks.

There's currently no `--re-run-scripts` flag, but you can simulate this by deleting chezmoi's state which is by default stored in `~/.config/chezmoi/chezmoistate.boltdb` before running `chezmoi apply`.

Hope this helps in the short term.

felipecassiors:
> There's currently no `--re-run-scripts` flag, but you can simulate this by deleting chezmoi's state which is by default stored in `~/.config/chezmoi/chezmoistate.boltdb` before running `chezmoi apply`.
> 
> Hope this helps in the short term.

It does, thank you so much!

PR #1026, opened 2021-02-06
Feat: Add noscript tag

benmezger:
This adds the `no-script` tag, which as far I understood @twpayne did implement it for `v2` (?), but I was not able to find it. If this PR makes sense to merge, I would be happy to migrate this to `chezmoi2`.

I have a question regarding the tests, how would one test this flag?

This is related to #854


twpayne:
Thanks very much for this @benmezger and sorry for the slow response.

chezmoi v2 is really close now (#1036) and includes this functionality. The mechanism used is the more general [`--include` flag](https://github.com/twpayne/chezmoi/blob/master/chezmoi2/docs/REFERENCE.md#--include-types) which extends `noscript` to apply to more types.

I hope to merge #1036 in the next few days (the code and testing is done, it's mostly documentation that needs updating) and this PR will not be mergeable when that is done, so I'll close this PR. Thank you again, and very happy to continue discussion :)

Issue #1046, opened 2021-03-01
Is it possible to ignore a single line in a target file?


Daniel157:
## What exactly are you trying to do?

I'm assembling my config file for kubectl (```$HOME/.kube/config```) using chezmoi. The config file consists of credentials for different Kubernetes clusters. Here you can see an example, how a config file for kubectl typically looks like: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/

In this config file, there is one line called ```current-context```. This line indicates the cluster, that I'm currently interacting with. This line changes all the time (depending on which cluster I'm currently working with).
I would like chezmoi to simply ignore this line. It should not overwrite/remove this line each time I call ```chezmoi apply```.

Is it possible to achieve this somehow?

## What have you tried so far?

I would imagine that it should be possible to do something like this in my template file:

```
current-context: {{ ignore }}
```

However, I couldn't find something like this.

## Where else have you checked for solutions?

* [x] I have read [chezmoi's how-to guide](https://github.com/twpayne/chezmoi/blob/master/docs/HOWTO.md), and not found the answer.
* [x] I have read [chezmoi's FAQ](https://github.com/twpayne/chezmoi/blob/master/docs/FAQ.md), and not found the answer.
* [x] I have searched [chezmoi's reference guide](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md), and not found the answer.
* [ ] Other, please give details.

## Output of `chezmoi doctor`

```
$ chezmoi doctor
     ok: version 1.8.11, commit 2e5a76c23de034344f7a9fa89cd3088cd9b52ac7, built at 2021-02-05T00:46:35Z, built by goreleaser
     ok: runtime.GOOS linux, runtime.GOARCH amd64
     ok: /home/daniel/.local/share/chezmoi (source directory, perm 700)
     ok: /home/daniel (destination directory, perm 755)
     ok: /home/daniel/.config/chezmoi/chezmoi.toml (configuration file)
     ok: /usr/bin/zsh (shell)
     ok: /usr/bin/nano (editor)
     ok: /usr/bin/vimdiff (merge command)
     ok: /usr/bin/git (source VCS command, version 2.25.1)
     ok: /usr/bin/gpg (GnuPG, version 2.2.19)
warning: op (1Password CLI, not found)
warning: bw (Bitwarden CLI, not found)
warning: gopass (gopass CLI, not found)
[...]

PR #1047, opened 2021-03-01
chezmoi2: Add output template function

twpayne:
Refs #1046.


twpayne:
cc @Daniel157 for review

PR #1061, opened 2021-03-13
Initialize the `keyring` cache

halostatue:
- Initialize the keyring cache to prevent errors when using keyring template functions.


twpayne:
Thank you very much @halostatue :)

PR #1072, opened 2021-03-18
Add new attributes to chattr reference documentation.

gotgenes:
Documents the ability to use `after`, `before`, and `once` with `chattr` command.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -670,10 +670,13 @@ attributes and their abbreviations are:
 
 | Attribute    | Abbreviation |
 | ------------ | ------------ |
+| `after`      | `a`          |
+| `before`     | `b`          |
 | `empty`      | `e`          |
 | `encrypted`  | *none*       |
 | `exact`      | *none*       |
 | `executable` | `x`          |
+| `once`       | `o`
 | `private`    | `p`          |
 | `template`   | `t`          |
 


gotgenes:
I saw these choices [present in the source code](https://github.com/twpayne/chezmoi/blob/master/cmd/chattrcmd.go#L43-L53) of `chattrcmd.go`, but undocumented by the command (which gets the help documentation from `REFERENCES.md`). I thought I'd be helpful by documenting them.

However, upon further investigation, I think the following are true:

1.  These attributes of `after_`, `before_`, and `once_` only apply to `run_` and perhaps `modify_` scripts, which are source-only files.

2.  The `chattr` command [only accepts tracked target files as arguments](https://github.com/twpayne/chezmoi/blob/master/cmd/chattrcmd.go#L96); it will not accept a  source-only file. Attempting to give a source-only file causes an error like

    ```
    $ chezmoi chattr before run_example.sh
    chezmoi: run_example.sh: not in source state
    ```

3.  When used on a target file that is tracked by chezmoi, the result of providing these attributes is no-operation, because the code will only apply these changes to . The file will be unmodified.

So maybe the better thing is not to document these at all.

It's not yet clear to me why these attributes are included in the map, given [the logic to apply them in the case statement](https://github.com/twpayne/chezmoi/blob/master/cmd/chattrcmd.go#L262-L268) can never be reached by the main method. The only case that can be possible is `chezmoi.SourceStateFile`.

It would be neat if `chattr` could take these other source file types in the future.

twpayne:
Thanks for this @gotgenes - you're completely on the right track, just missing one thing that I haven't documented.

Targets are specified as the name of the target file, which is the name of the source file after all the attributes have been processed. For a script, this means it's the name without `run_`, `once_`, `before_`, and `after_`. So, for example, if you have a script called `run_example.sh` in your source state then you can change its attributes with:

    chezmoi chattr before ~/example.sh

even though `~/example.sh` will never appear as a file in your home directory.

The reference manual really needs a section describing targets.

gotgenes:
A ha! Okay, I follow this, now, seeing the way to specify the target that will map to the source-only file. Thanks!

PR #1089, opened 2021-03-22
Only update state if target was actually changed

twpayne:


twpayne:
@fwSmit thanks for spotting that chezmoi v2 was [slower than expected](https://github.com/twpayne/chezmoi/issues/987#issuecomment-762788712).

It turns out that `chezmoi apply` in v2 was re-writing the state of every target, even if the target was not changed. This PR fixes that, giving a significant performance improvement (2x-ish on Linux, much more on macOS).

PR #1093, opened 2021-03-23
Update howto to use main instead of master

evilhamsterman:
Just a simple update to the howto docs to reference main as the branch rather than master since github and git are changing the default. This also aligns with your quickstart


twpayne:
Thank you :)

PR #1100, opened 2021-03-25
Use the first existing parent directory for a script's working directory

### LINKED ISSUE #1097: Is it possible to "organize" `before_` scripts?

Since before v2, I've had a folder of setup scripts that would install software and do some configuration actions, mostly written in Powershell (so they can be shared between OSes), plus a separate bootstrap script that would install my password manager and (on Linux) install Powershell.  Last night I decided to try to migrate the bootstrap script to a `once_before_` script.

I created `run_once_before_bootstrap-linux.sh` inside the same `setup` directory where my other scripts are stored.  The issue I ran into is that [`TargetStateScript.Apply`](https://github.com/twpayne/chezmoi/blob/master/internal/chezmoi/targetstateentry.go#L195) wants to set the CWD for the script to `~/setup`, which (of course) doesn't exist, since the whole point of `before_` is that it's before the main files are applied. 

## Additional context

Putting my `before_` script directly in the source directory works, of course, but I'd prefer not to have to do that, especially in the event that I end up with several of them.  Maybe it would be possible for `Apply` or `RunScript` to detect that the desired working directory doesn't exist and just use the destination dir instead?  I tried playing around with that a bit myself, but I couldn't quickly figure out how to determine what the destination was from inside `Apply`.  Of course, I'd love to hear any other thoughts!  I had a crazy idea to abuse the templating system by storing the setup code in `.chezmoitemplates/setup` and just templating it into a single top-level script, but that just feels differently dirty, rather than actually any better :smile:

zb140:
This allows putting `before_` scripts in subdirectories, which naturally won't exist during the `before` phase.

Fixes #1097.

### DIFF
--- a/internal/chezmoi/realsystem.go
+++ b/internal/chezmoi/realsystem.go
@@ -100,10 +100,26 @@ func (s *RealSystem) RunScript(scriptname RelPath, dir AbsPath, data []byte) (er
 	// Run the temporary script file.
 	//nolint:gosec
 	cmd := exec.Command(f.Name())
-	cmd.Dir = string(dir)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+
+	// The requested working dir may not actually exist yet, if this is a
+	// `before_` script.  In that case, look through the parent hierarchy till
+	// we find one.
+	for workDir := dir; string(dir) != "."; workDir = workDir.Dir() {
+		if _, err = s.Stat(workDir); err != nil {
+			if !os.IsNotExist(err) {
+				// something bad happened, but it wasn't that the dir didn't exist
+				return
+			}
+		} else {
+			// If there was no error, workDir exists and can be used as the CWD.
+			cmd.Dir = string(workDir)
+			break
+		}
+	}
+
 	err = s.RunCmd(cmd)
 	return
 }
--- a/testdata/scripts/scriptsubdir_unix.txt
+++ b/testdata/scripts/scriptsubdir_unix.txt
@@ -14,10 +14,13 @@ exec tar -tzf archive.tar.gz
 cmp stdout golden/archive
 
 -- golden/apply --
+$HOME
 $HOME/dir
 -- golden/archive --
+otherdir/script
 dir/
 dir/script
+otherdir/
 -- golden/dump --
 {
   "dir": {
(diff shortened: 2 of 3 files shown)


internal/chezmoi/realsystem.go
```diff
@@ -100,10 +100,20 @@ func (s *RealSystem) RunScript(scriptname RelPath, dir AbsPath, data []byte) (er
...
 	cmd := exec.Command(f.Name())
-	cmd.Dir = string(dir)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+
+	// The requested working dir may not actually exist yet, if this is a
+	// `before_` script.  In that case, look through the parent hierarchy till
+	// we find one.
+	for workDir := dir; string(dir) != "."; workDir = workDir.Dir() {
```
zb140, line 110: Is this an idiomatic way to do this in Go?  I'm usually a `{Java,Type}Script` person, so sometimes I'm not sure :wink:

twpayne: Yup, this totally works. Also, as `dir` is not used again after this loop, it would be equally good to mutate the `dir` variable in a while loop.

internal/chezmoi/realsystem.go
```diff
@@ -100,10 +100,20 @@ func (s *RealSystem) RunScript(scriptname RelPath, dir AbsPath, data []byte) (er
...
+		if _, err = os.Stat(string(workDir)); !os.IsNotExist(err) {
```
twpayne, line 111: This silently ignores all non-`os.IsNotExist` errors. Non-`nil` errors that are not `os.IsNotExist` errors should be returned. This also uses `os.Stat` whereas it should use `s.Stat`.

testdata/scripts/scriptsubdir_unix.txt
```diff
@@ -14,10 +14,13 @@ exec tar -tzf archive.tar.gz
 cmp stdout golden/archive
 
 -- golden/apply --
+$HOME
 $HOME/dir
 -- golden/archive --
+otherdir/script
 dir/
 dir/script
+otherdir/
```
twpayne, line 23: I'm a bit surprised that this test passes as chezmoi normally sorts filenames alphabetically to give repeatable output.

internal/chezmoi/realsystem.go
zb140: Oops, I totally should have picked up on that. Thanks!

testdata/scripts/scriptsubdir_unix.txt
zb140: Yeah, I'm glad you mentioned that, I was surprised too.  I eventually convinced myself it was `tar` that wasn't sorting them, and that's why it was in this order.  But I could be wrong about that.

twpayne: Actually, I just realized that what happens is that chezmoi emits the diff in the order that the changes are applied, which is normally in alphabetical order. However, this is a `before_` script, so it's run before the destination directory is updated, which is why `otherdir/script` appears first.

PR #1112, opened 2021-03-29
Fix issue in example for trying to use existing boolean vars first in…

bionade24:
… How To section: First arg of get doesn't have to be in quotes.
http://masterminds.github.io/sprig/dicts.html


twpayne:
Great catch, thanks!

Issue #1075, opened 2021-03-18
Should try symlink templates files when files are pointing to absolute well known directories.


3v1n0:
## Is your feature request related to a problem? Please describe.

This is a bit related to https://github.com/twpayne/chezmoi/issues/167#issuecomment-802005513 so, when a symlink is added to chezmoi, it should instead try to create a symlink template if the symlink is pointing to a path in `{{ .chezmoi.homedir }}` or `{{ .chezmoi.sourceDir }}`  (it's actually the main purpose of my scripts as documented in https://github.com/twpayne/chezmoi/issues/726#issuecomment-636518384)

This will make such links to be more portable, without having to care at local level to use relative paths.

## Describe the solution you'd like

When a symlink is detected:
 - If it's an absolute path and it's pointing to a sub-dir of the home or chezmoi dir, it should create a symlink template instead.

## Describe alternatives you've considered

Basically having scripts doing this for me

twpayne:
Thanks to @Flygrounder for implementing this in #1124. See #1139 for a couple of minor tweaks.

PR #1124, opened 2021-04-01
Added --template-symlinks-home and --template-symlinks-source options

Flygrounder:
This is not a finished work, opening a PR to discuss changes. 
I am trying to add functionality, described in #1075. I've added a basic implementation of templating symlinks with targets inside home, that replaces `s.destDirAbsPath` with `{{ .chezmoi.homedir }}`.
To test this functionality, you can run `chezmoi add --template-symlinks-home <your_symlink>`

 I am struggling with adding a test for this functionality. The problem is when I add an absolute symlink in test, it points to `/tmp/go-vfs-vfst3713338411/home/user/.dir/subdir/file`, not `/home/user/.dir/subdir/file`. Are there any workarounds?


twpayne:
Thanks very much for this!

For the tests, it's probably easiest to use a [`testscript`](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript) test for this, like the existing tests in [`testdata/scripts`](https://github.com/twpayne/chezmoi/tree/master/testdata/scripts).  Within these tests, the variable `HOME` points to the temporary home directory. This means that you'll probably want something like the following in `testdata/scripts/symlinks_templatesymlinkshome.txt`:

```
mkhome

# test that chezmoi add --template-symlinks-home creates an absolute symlink
symlink .symlink -> .dir/subdir/file
chezmoi add --template-symlinks-home $HOME${/}.symlink
chezmoi cat $HOME${/}.symlink
stdout ${HOME@R}/.dir/subdir/file
```

Warning: I haven't tested this, so it might need a couple of tweaks.

Flygrounder:
At this point, this PR should close #1075

twpayne:
Thanks very much for this. Would you mind squashing all the commits into a single commit?

Flygrounder:
No problems, you can squash them if you find this appropriate

twpayne:
Thank you very much @Flygrounder :)

PR #1132, opened 2021-04-03
Follow Systemd FHS spec

EvanCarroll:
The directory `~/bin` is an antiquated 20 year old social convention that came out of need and a failure on the part of the Linux Filesystem Hierarchy Standard (FHS) to state anything on the matter. The new replacement for that is standardized by the Systemd File Hierarchy Specification; see `man file-hierarchy` for more information. It states `~/.local/bin/` is the right place for user-installed binary files.

### DIFF
--- a/assets/scripts/install.sh
+++ b/assets/scripts/install.sh
@@ -7,7 +7,7 @@
 
 set -e
 
-BINDIR=${BINDIR:-./bin}
+BINDIR=${BINDIR:-./.local/bin}
 TAGARG=latest
 LOG_LEVEL=2
 EXECARGS=
--- a/internal/cmd/generate-install.sh/install.sh.tmpl
+++ b/internal/cmd/generate-install.sh/install.sh.tmpl
@@ -7,7 +7,7 @@
 
 set -e
 
-BINDIR=${BINDIR:-./bin}
+BINDIR=${BINDIR:-./.local/bin}
 TAGARG=latest
 LOG_LEVEL=2
 EXECARGS=


twpayne:
Thanks for the PR. Sadly this will break everybody who relies on the existing behavior. You can use the `-b` option to the install script to set the installation directory.

EvanCarroll:
Two things to consider,

* Any machine supporting systemd should already have their path configured as such. I don't have to do anything to make it work. Whereas I do have to add `~/bin` to make the default work. Likewise if the installation file moved the only thing that should break on systemd machines is when the path is hard set to `~/bin` in scripts.
* Your docs already show this in places. Specifically in [here the default is overridden to `~/.local/bin`](https://github.com/twpayne/chezmoi/blob/ae7dda15971bd6d572fd74948592b7afc7a30109/docs/HOWTO.md#use-chezmoi-with-github-codespaces-visual-studio-codespaces-visual-studio-code-remote---containers). But yet the [One-line binary install
](https://github.com/twpayne/chezmoi/blob/master/docs/INSTALL.md#one-line-binary-install) use the default and install it to a different location.

The only area I would forsee even minor confusion is on BSD they may have to update their path.

twpayne:
chezmoi runs on more machines than systemd. It runs on multiple variants of Linux, macOS, Windows, FreeBSD, and can easily be compiled for OpenBSD, NetBSD, Plan 9, and Solaris, none of which use systemd and don't support the FHS.

With hindsight, I perhaps would have made the default installation directory `.local/bin`, but even then I'm not sure as `.local` is hidden by default, which is potentially quite confusing. Anyhow, it is too late now as changing the default will break all scripts that rely on the existing behavior (which was maintained from the previous [godownloader](https://github.com/goreleaser/godownloader) script).

EvanCarroll:
Suffice it to say on closure, I would strongly suggest at least on machines that are systemd to install to the proper place, and not ask users to add new directories to their path. Even if backwards compat is an issue, before the users can pull down with chezmoi they'll have to know where the binary is (to provide the shell with an abs path) or add it to their `$PATH`. It seems somewhat likely that _some_ of them will hack a bashrc or profile to always add ~/bin to their path, which just isn't right and nothing uses it, at least on my machine my only holdover is `chezmoi` at that location.

So I would go with 

* Instructions and documentation that push for Linux users to using `~/.local/bin`
* Detecting in the script to default intelligently for Linux, which you already do here https://github.com/EvanCarroll/chezmoi/blob/c5fc6ecea767274fda27d06989b18c0407fd6a59/assets/scripts/install.sh#L50 (if interested I'll provide a patch).

The docs themselves seem to make it sound like this is core business domain,

> and/or different operating systems (e.g. Linux, macOS, and Windows) in sync, while still being able to easily cope with differences from machine to machine.

I'm also kind of curious, if osx users run this script they're ok with the install in `~/bin`?

twpayne:
I think most macOS users will install chezmoi with homebrew and not use the install script. Note that you can use the install script to install chezmoi in `~/.local/bin` with:

    sh -c "$(curl -fsSL https://git.io/chezmoi)" -- -b .local/bin

PR #1139, opened 2021-04-05
Replace --template-symlinks-home and --template-symlinks-source with --template-symlinks=value

twpayne:
This tidies up a few things from #1124:

* Replace `--template-symlinks-home` and `--template-symlinks-source` with `--template-symlinks=value` where `value` can be either `home` or `source`.
* Add documentation.
* Tidy up tests and logic.


3v1n0:
This looks nice, thanks.

However, ideally the selection between `source` and `home` would be automatic depending on the prefix, can be an `auto` mode be added?

twpayne:
> However, ideally the selection between `source` and `home` would be automatic depending on the prefix, can be an `auto` mode be added?

#1143 should implement this.

Issue #1066, opened 2021-03-16
chezmoi thinks that all files have been modified


twpayne:
## Describe the bug

This was reported by @fwSmit in #1039, specifically in [this comment](https://github.com/twpayne/chezmoi/issues/1039#issuecomment-798515391):

> I ran into this bug when running `chezmoi2 apply --source-path ...`. It gave an error that no tty was found, suggesting that some user input was needed. It turns out, some entirely different file was changed according to chezmoi2, even though I didn't change it.
> 
> ```shell
> ❯ chezmoi2 apply
> .XCompose has changed since chezmoi last wrote it [diff,overwrite,all-overwite,skip,quit]? diff
> diff --git a/.XCompose b/.XCompose
> .XCompose has changed since chezmoi last wrote it [diff,overwrite,all-overwite,skip,quit]? quit
> ❯ diff ./.XCompose ~/.local/share/chezmoi/dot_XCompose
> ❯ ls -l .XCompose
> -rw-r--r-- 1 friso friso 3413 Sep 19 08:57 .XCompose
> ❯ ls -l ~/.local/share/chezmoi/dot_XCompose
> -rw-r--r-- 1 friso friso 3413 Sep 19 08:57 ~/.local/share/chezmoi/dot_XCompose
> ```
> 
> I don't see any changes between the source file and the destination file, so maybe the internal cache is corrupted or something.
> 
> ```
> .XCompose has changed since chezmoi last wrote it [diff,overwrite,all-overwite,skip,quit]? skip
> .config/autostart has changed since chezmoi last wrote it [overwrite,all-overwite,skip,quit]? skip
> ...
> .config/git/config has changed since chezmoi last wrote it [diff,overwrite,all-overwite,skip,quit]? skip
> .config/gtk-3.0/colors.css has changed since chezmoi last wrote it [diff,overwrite,all-overwite,skip,quit]? quit
> ```
> 
> It seems like all files "have been changed since chezmoi last wrote it".
> I consider this to be two bugs:
> 
> * Chezmoi incorrectly says the file has changed.
> * `--source-path` should ignore all other files

Following conversation:

> Thanks for reporting this @fwSmit.
> 
> > * `--source-path` should ignore all other files
> 
> I think this is already the case, I tried to reproduce the problem with [ebdd6f4](https://github.com/twpayne/chezmoi/commit/ebdd6f4eb239c5b99fc22a8f9720ae8635ed9a62), but was unable to do so.
> 
[...]

Issue #1091, opened 2021-03-22
2.0.3 wants to change most of my file permissions


thernstig:
## Describe the bug
I installed 2.0.3 on a fresh Ubuntu install and then did `chezmoi init git@myrepo`. After that chezmoi reported things such as:

```sh
~> chezmoi diff
diff --git a/.bashrc b/.bashrc
old mode 100644
new mode 100664
```

I have never seen that before. My umask is set to 002 in fish config. If I just type `umask` I get `0002`.

## To reproduce

No idea.

## Expected behavior

Since my umask is `002` I did not expect chezmoi to change the group permission to writeable.

## Output of command with `--verbose`

```
~> chezmoi command --verbose
chezmoi: unknown command "command" for "chezmoi"
```

## Output of `chezmoi doctor`

```
RESULT    CHECK                MESSAGE
ok        version              v2.0.3, commit c500b8a6fee41e66309fcf8ad2f5e94d73e5bed7, built at 2021-03-22T00:40:59Z, built by goreleaser
ok        os-arch              linux/amd64
ok        config-file          /home/user/.config/chezmoi/chezmoi.toml is a file
ok        source-dir           /home/user/.local/share/chezmoi is a directory
ok        suspicious-entries   no suspicious entries found in /home/user/.local/share/chezmoi
ok        dest-dir             /home/user is a directory
ok        shell                found /usr/bin/fish
ok        editor               found /mnt/c/Users/user/AppData/Local/Programs/Microsoft VS Code/bin/code
ok        git-cli              found /usr/bin/git, version 2.25.1
ok        merge-cli            found /usr/bin/vimdiff
ok        gnupg-cli            found /usr/bin/gpg, version 2.2.19
warning   1password-cli        op not found in $PATH
warning   bitwarden-cli        bw not found in $PATH
warning   gopass-cli           gopass not found in $PATH
warning   keepassxc-cli        keepassxc-cli not found in $PATH
warning   keepassxc-db         not set
warning   lastpass-cli         lpass not found in $PATH
warning   pass-cli             pass not found in $PATH
warning   vault-cli            vault not found in $PATH
warning   secret-cli           not set
```

## Additional context

I am using wsl 2 with Ubuntu 20.04 installed.

twpayne:
What's the output of

    ls -l ~/.bashrc

before you run `chezmoi diff`?

[...]

PR #1149, opened 2021-04-08
Store actual umasks in state file

twpayne:
This is a work-in-progress fix for #1066, not ready for merging yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,14 @@ jobs:
         [Environment]::SetEnvironmentVariable("Path", $env:PATH, "Machine")
         choco install --yes gpg4win
         echo "C:\Program Files (x86)\GnuPG\bin" >> $env:GITHUB_PATH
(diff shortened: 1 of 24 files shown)


fwSmit:
Are you sure that it's a umask thing? I didn't change it (as far as I know), but the bug still occurred.

twpayne:
I'm pretty sure that this is a bug in chezmoi, and actually the same as #1066, #1091, and #1155.

The tl;dr is that there are actually two bugs:
1. Due to incorrect umask handling, chezmoi is storing the last-written permissions in its persistent state incorrectly, so it keeps thinking that a file has changed when actually it hasn't.
2. The diff code is displaying an empty diff when it should be displaying that difference in permissions that chezmoi thinks exists (but actually doesn't).

This PR isn't complete yet. It's missing:
1. A proper set of tests.
2. A migration for existing chezmoi users. If the above theory is correct, then chezmoi2 has stored permissions incorrectly for many users, and needs to fix itself.

PR #1162, opened 2021-04-16
Do not follow symlinks if the file will be ignored anyway

### LINKED ISSUE #1161: Files starting with .# in source are not actually ignored (chezmoi version v2.0.4)
## Describe the bug

Chezmoi is not actually ignoring files that start with a . that are found in source. From the [reference page](https://www.chezmoi.io/docs/reference/):

> chezmoi ignores all files and directories in the source directory that begin with a . with the exception of files and directories that begin with .chezmoi.

For me, this is particularly annoying when editing my source dotfiles from Emacs since it creates lockfiles called `.#filename.ext` which are symlinks.

## To reproduce

  1. Opened source directory file in emacs. `emacs ~/.local/share/chezmoi/dot_aliases` 
  2. Made edit. 
  3. Did not save buffer. 
  4. `chezmoi diff` &rarr; `chezmoi: stat /home/user/.local/share/chezmoi/.#dot_aliases: no such file or directory`

## Expected behavior

Chezmoi should fully ignore (i.e. not touch) all files starting with `.`, including Emacs lockfiles.

```
$ chezmoi diff --verbose
chezmoi: stat /home/user/.local/share/chezmoi/.#dot_aliases: no such file or directory
```

## Output of `chezmoi doctor`

```
$ chezmoi doctor
```

## Additional context

I've written about a solution from the Emacs side [here](https://github.com/tuh8888/chezmoi.el#lock-files) which disables lockfile creation. However, these lockfiles serve a purpose, and I am not sure why chezmoi needs to be touching these files if it is supposed to be ignoring them.

twpayne:
Fixes #1161.


twpayne:
Cirrus CI is failing due to what looks like an incorrect ACL from Google:

```
$ curl https://storage.googleapis.com/golang/go1.16.3.freebsd-amd64.tar.gz
<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object.</Details></Error>%
```

PR #1177, opened 2021-04-29
Correct `and` template example in How To guide

jesstelford:
I'm new to Go's templates, but based on [the docs](https://golang.org/pkg/text/template/#hdr-Functions), it looks like there were two erroneous `)` in the example here.


twpayne:
Thanks for spotting this and for the fix :)

PR #1191, opened 2021-05-05
Update rpm zsh completion directory to match recent Fedora versions.

tomprince:
Looking at the source for zsh[1], it appears that the default directory for
extra functions like completions is `../site-functions` not `../functions`. This
matches the behavior on Fedora.

[1] https://github.com/zsh-users/zsh/blob/c1f932d66896753eb118003b7c1b1139bc6c8725/configure.ac#L308


twpayne:
Thanks!

PR #1199, opened 2021-05-09
Validate generate config before writing it in init command

### LINKED ISSUE #1197: Validate configuration before writing to file

I've been tinkering with my chezmoi config quite a bit (hi, issues!) and have a pattern of accidentally writing an incorrect config file from my template. 

Once the config file has been written, chezmoi can no longer read it, so I have to manually fix it before re-applying my file.

In my most recent example, I accidentally added this to my toml config:
```
[diff]
exclude: ["scripts"]
```
Which is not valid toml, but was still written to the template.

I'd suggest a flow during `chezmoi init` where chezmoi 
* writes the configuration to a temporary buffer or similar
* verifies it is valid
* if valid, writes to config file
* if not
    * writes error to stderr
    * exits with error code 

* Being better myself (unlikely)
* Attempting to read partial config - this seems complicated to implement and likely more fragile, with unpredictable side effects

## Additional context
I don't think there is any - but I'd be happy to take a crack at this if interested.

twpayne:
Fixes #1197.

@sheeley is this what you had in mind?

### DIFF
--- a/cmd/initcmd.go
+++ b/cmd/initcmd.go
@@ -175,6 +175,27 @@ func (c *Config) runInitCmd(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+
+		// Validate the config.
+		v := viper.New()
+		v.SetConfigType(ext)
+		if err := v.ReadConfig(bytes.NewBuffer(configFileContents)); err != nil {
+			return err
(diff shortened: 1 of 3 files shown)


sheeley:
Holy cow, yes! You're incredible.

Issue #1074, opened 2021-03-18
Chezmoi just fails when encryption is enabled and running from console


3v1n0:
## Describe the bug

When running chezmoi (configured to use gpg) in an headless setup it will just fail instead of providing a text-only based UI for decrypting.

Once I connect a display to the setup, the gpg key is requested via polkit dialog, so all can be unlocked.
So it's quite critical if you're using chezmoi in a remote server or a display-less setup

## To reproduce

Connect via SSH to an installation with chezmoi
 > chezmoi status

chezmoi: exit status 2

## Expected behavior

The gpg textual dialog is prompted to insert your gpg key.

## Output of command with `--verbose`

Same as above, no info.

However strace shows something more

```
[pid 10440] write(2, "gpg: public key decryption faile"..., 65 <unfinished ...>
[pid 10424] <... epoll_pwait resumed>[{EPOLLIN, {u32=1076465536, u64=140536701357952}}], 128, 1, NULL, 7) = 1
[pid 10440] <... write resumed>)        = 65
[pid 10424] futex(0x1e883d8, FUTEX_WAKE_PRIVATE, 1 <unfinished ...>
[pid 10440] write(2, "\n", 1 <unfinished ...>
[pid 10424] <... futex resumed>)        = 1
[pid 10429] <... futex resumed>)        = 0
[pid 10424] read(10,  <unfinished ...>
[pid 10440] <... write resumed>)        = 1
[pid 10424] <... read resumed>"gpg: public key decryption faile"..., 32768) = 66
[pid 10440] write(2, "gpg: decryption failed: No secre"..., 37 <unfinished ...>
[pid 10424] read(10,  <unfinished ...>
[pid 10440] <... write resumed>)        = 37
[pid 10424] <... read resumed>"gpg: decryption failed: No secre"..., 32768) = 37
[pid 10440] write(2, "\n", 1 <unfinished ...>
[pid 10424] read(10,  <unfinished ...>

```

## Output of `chezmoi doctor`

```
$ chezmoi doctor
RESULT    CHECK                MESSAGE
ok        version              v2.0.0, commit e0143cd0c7ec8b7bbe5f6a04d42a5dd5ed823b75, built at 2021-03-18T01:38:35Z, built by goreleaser
ok        os-arch              linux/amd64
ok        config-file          /home/marco/.config/chezmoi/chezmoi.toml is a file
ok        source-dir           /home/marco/.local/share/chezmoi is a directory
[...]

PR #1229, opened 2021-05-24
Use temporary files, instead of stdin/stdout, when calling gpg

twpayne:
Refs #1074.

@3v1n0 would you be able to test if this PR fixes the problems you reported in #1074?

It changes the way that chezmoi reads and writes data from gpg back to how it was done prior to chezmoi v2, i.e it reverts back to using temporary files rather than gpg's standard input and output.


twpayne:
Merging this as there's a small performance impact but it should fix stuff.

Issue #751, opened 2020-05-13
Add `.chezmoidata.<format>` for template data


twpayne:
## Is your feature request related to a problem? Please describe.

This was first [suggested](https://github.com/twpayne/chezmoi/issues/436#issuecomment-535484492) by @martin2250 and then came up again in #749 opened by @cljoly. 

I want to be able have structured data stored in my source state and available to templates.

## Describe the solution you'd like

chezmoi should read structured data from a `.chezmoidata.<format>` file in the source state and make it available as template data. `<format>` can be either `json`, `toml`, or `yaml`.

Template data should be read in the following order, with newer data overwriting older data:

1. Default data (e.g. `.chezmoi.hostname`).
2. Data read from `.chezmoidata.<format>` in the source state.
3. Data read from the `data` section of the config file.

If there are multiple `.chezmoidata.<format>` files then they should be read in alphabetical order with later formats overriding earlier ones.

## Describe alternatives you've considered

Using only the existing config file for data means that changes have to be manually propagated from machine to machine, as @martin2250 points out.

Making `.chezmoidata.<format>` a template itself starts to become very complicated.

## Additional context

Thanks to @martin2250 for the idea - I didn't realize it was such a good idea at the time.

twpayne:
@martin2250 for info this is now working in the `v2` branch (#654).

While developing it, I hit an interesting corner case:

`.chezmoidata.<format>` contains template data, and that template data can be used in other control files that are interpreted as templates, like `.chezmoiignore` and `.chezmoiremove`. chezmoi walks through the filesystem in alphabetical order and, luckily, `data` occurs before `ignore` and `remove` in the English dictionary so we don't have to do anything special to ensure that data from `.chezmoidata.<format>` is available in `.chezmoiignore` and `.chezmoiremove`. Phew!

If `.chezmoidata` was interpreted as a template itself it would become strangely self-referential, but of course we could work around this by defining the template data used when `.chezmoidata` is interpreted to be the data accumulated so far before the `.chezmoidata` template itself is executed.

twpayne:
Oh, and unluckily for us, `.chezmoitemplates` is read after `.chezmoiignore` and `.chezmoiremove`, so you don't get to use partial templates in `.chezmoiignore` and `.chezmoiremove` unless I add a hack to change the filesystem traversal order. I don't think anyone will need or complain about this, but this is the Internet.

cljoly:
[...]

PR #1233, opened 2021-05-25
Reference .chezmoidata in template data documentation

cljoly:
This was introduced in #751


twpayne:
This is great, thank you!

PR #1246, opened 2021-06-06
Fix typo in REFERENCE.md

daanzu:
I confess I didn't test building the site after this change. BTW, thanks for the great tool!


twpayne:
Thank you very much!

PR #1251, opened 2021-06-07
Fix a minor typo in HOWTO.md

eugenesvk:
remove any whitespace before and after→
remove and


twpayne:
Thank you very much @eugenesvk for both this and #1252 :)

PR #1280, opened 2021-06-30
boom, adding support for account

### LINKED ISSUE #1278: support account choice for 1password

I use two accounts on 1P, one personal and one from work. When I do `export AWS_ACCESS_KEY_ID="{{ (onepassword "aws credentials").details."access key id" }}"`, chezmoi (and op) don't know where to find that. If I and randomly lastly signed in the right account, it works, but if not, it doesn't.

On the terminal I can do `op get item "aws credentials" --account my` and it will show that there is no item like that. Then doing `op get item "aws credentials" --account team_account` shows me the item.

I would like a way to do something like `{{ (onepassword "aws credential" "team_account").details.whatever }}` and get the right item.

The problem is that now there is a way to add a second argument and it is used for `vault`. I wonder how we could add a way to select an account. I'm happy to implement it myself, but maybe @twpayne you have a good idea for the "API" here?

pvinis:

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -102,9 +102,9 @@ Manage your dotfiles securely across multiple machines.
   * [`lastpass` *id*](#lastpass-id)
   * [`lastpassRaw` *id*](#lastpassraw-id)
   * [`lookPath` *file*](#lookpath-file)
-  * [`onepassword` *uuid* [*vault-uuid*]](#onepassword-uuid-vault-uuid)
-  * [`onepasswordDocument` *uuid* [*vault-uuid*]](#onepassworddocument-uuid-vault-uuid)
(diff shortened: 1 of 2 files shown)


twpayne:
Thanks, good looks good! Please can you also line wrap the documentation changes to match the existing style.

Please can you squash the commits into one and change the commit message to match chezmoi's style, e.g.

    Add optional account arg to onepassword template functions

The commit messages are used in the auto-generated release notes so they need to be relevant :)

pvinis:
Do you do the wrapping manually or is there a script or something?

twpayne:
I do the wrapping in my editor (VS Code) and have a hot key that wraps paragraphs of text.

pvinis:
Squashed and wrapped 👍

twpayne:
Awesome, thank you @pvinis :)

PR #1296, opened 2021-07-12
Make modify scripts always overwrite the target without prompting

twpayne:
As [identified by @lcrockett](https://github.com/twpayne/chezmoi/issues/1046#issuecomment-864343405) in #1046.

Before this change, chezmoi would always warn if the target of a modify script had changed since chezmoi last wrote it, but this is counter to the concept of a modify script: we expect the target of modify scripts to have changed since chezmoi last wrote it.

This PR changes the behavior of chezmoi to always overwrite the target of modify scripts, even if `--force` is not specified.

@lcrockett would you be able to test this? No worries if not :)


lcrockett:
Works as intended, thank you for the change !

```bash
$ chezmoi apply
.k9s/config.yml has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]? skip
$ ./chezmoi apply
$ chezmoi diff
$ 
```

As far as I can tell, this is an all-or-nothing context change. That works fine for me, not sure if at some point a selective approach (only a selection of modify_ 'files' would be automatically overwritten) would be more practical. I reckon time will tell. Cheers !

twpayne:
Great, thanks for the testing @lcrockett - much appreciated.

PR #1303, opened 2021-07-14
Allow source directory to be a symlink

### LINKED ISSUE #1299: Error if source directory is symlinked
## Describe the bug
If the source directory (such as the default one, `~/.local/share/chezmoi`) is a symlink - which points to a valid directory - then chezmoi will error with, ie:

```
chezmoi: C:/Users/<user>/.local/share/chezmoi: not a directory
```

## To reproduce

I verified the bug on windows - to do so, set up chezmoi like normal, then do (from a cmd shell):

```
move %USERPROFILE%\.local\share\chezmoi %USERPROFILE%\.local\share\chezmoi_real
mklink /J  %USERPROFILE%\.local\share\chezmoi %USERPROFILE%\.local\share\chezmoi_real
```

I haven't tested on other OSes, but from looking at the code, I think it should be reproducible there too, ie:

```
mv ~/.local/share/chezmoi ~/.local/share/chezmoi_real
ln -s ~\.local\share\chezmoi_real ~\.local\share\chezmoi
```

## Expected behavior

Since the symlink is valid, I would expect chezmoi to work normally.

```
> chezmoi apply --verbose
chezmoi: C:/Users/<user>/.local/share/chezmoi: not a directory
```

## Output of `chezmoi doctor`

```
> chezmoi doctor
ok        version              v2.1.1, commit 4b428e8f2a80c64562fb0fecee60cf25c0b91db1, built at 2021-07-11T15:12:03Z, built by goreleaser
ok        config-file          C:/Users/paulm/.config/chezmoi/chezmoi.toml is a file
ok        source-dir           C:/Users/paulm/.local/share/chezmoi is a directory
ok        suspicious-entries   no suspicious entries found in C:/Users/paulm/.local/share/chezmoi
ok        dest-dir             C:/Users/paulm is a directory
ok        shell                found C:\WINDOWS\system32\cmd.exe
ok        editor               found C:\WINDOWS\system32\notepad.exe
ok        git-cli              found C:\Program Files\Git\cmd\git.exe, version 2.17.1
warning   merge-cli            vimdiff not found in $PATH
ok        age-cli              age not found in $PATH
ok        gnupg-cli            gpg not found in $PATH
info      1password-cli        op not found in $PATH
info      bitwarden-cli        bw not found in $PATH
info      gopass-cli           gopass not found in $PATH
info      keepassxc-cli        keepassxc-cli not found in $PATH
info      lastpass-cli         lpass not found in $PATH
info      pass-cli             pass not found in $PATH
info      vault-cli            vault not found in $PATH
info      secret-cli           not set
```

## Additional context

I think the error is being raised by this line:
https://github.com/twpayne/chezmoi/blob/1a55a00825d804039c67c5e0a159c817090b830e/internal/chezmoi/sourcestate.go#L550

I think the problem is that `LStat` is used here:
https://github.com/twpayne/chezmoi/blob/1a55a00825d804039c67c5e0a159c817090b830e/internal/chezmoi/sourcestate.go#L543

...when I think just a normal `Stat` is what would be desired.  By using `LStat` on a symlink to a directory, when we subsequently call `IsDir()`, it returns false, because it's a symlink.

twpayne:
Fixes #1299.

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -540,8 +540,7 @@ func (s *SourceState) MustEntry(targetRelPath RelPath) SourceStateEntry {
 
 // Read reads the source state from the source directory.
 func (s *SourceState) Read() error {
-	info, err := s.system.Lstat(s.sourceDirAbsPath)
-	switch {
+	switch info, err := s.system.Stat(s.sourceDirAbsPath); {
 	case errors.Is(err, fs.ErrNotExist):
 		return nil
 	case err != nil:
@@ -552,7 +551,7 @@ func (s *SourceState) Read() error {
 
 	// Read all source entries.
 	allSourceStateEntries := make(map[RelPath][]SourceStateEntry)
-	if err := Walk(s.system, s.sourceDirAbsPath, func(sourceAbsPath AbsPath, info fs.FileInfo, err error) error {
+	if err := WalkDir(s.system, s.sourceDirAbsPath, func(sourceAbsPath AbsPath, info fs.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


pmolodo:
Awesome, thanks!

PR #1304, opened 2021-07-14
Allow install of chezmoi from a powershell where StrictMode is on

pmolodo:
Currently, if, ie, the currently running context has called:

```
Set-StrictMode -version latest
```

...then `install.ps1` will error when, ie,

`[Runtime.InteropServices.RuntimeInformation]::OSArchitecture`
 in `get_goarch` is fetched from .NET 4.0.

This change won't affect the calling context.

### DIFF
--- a/assets/scripts/install.ps1
+++ b/assets/scripts/install.ps1
@@ -247,6 +247,10 @@ function Install-Chezmoi {
         [string[]]$ExecArgs
     )
 
+    # some sub-functions (ie, get_goarch, likely others) require fetching of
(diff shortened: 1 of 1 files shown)


twpayne:
Thank you very much!

cc @zb140 for review, but I assume that this is a safe change, so merging by default.

zb140:
Yeah, the script does assume that accessing non-existing properties will not fail. Sorry about that. 👍 on the change. Thanks!

twpayne:
Thank you very much @zb140 😄

PR #1314, opened 2021-07-20
Fix bare-git and add VCSH to comparison chart

alerque:
- Normalize cell length in Markdown table
- Set the record straight on some bare-git features
- Add VCSH to comparison chart


alerque:
Full disclosure: I've (just in the past year) taken over maintenance of `vcsh` from it's original author, hence I have a vested interest. Looking around at "competitors" chezmoi seems to have a lot going for it, and I give it full props for that.

twpayne:
Thank you very much!

PR #1315, opened 2021-07-21
allow entering username/password when using builtin git clone

pmolodo:
I was unable to clone a private chezmoi repo when bootstrapping on a system that did not have git installed
(https://github.com/twpayne/chezmoi/discussions/1310)

This commit allows this, by allowing the user to enter their username/
password at the command prompt.

### DIFF
--- a/internal/cmd/initcmd.go
+++ b/internal/cmd/initcmd.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"errors"
@@ -11,10 +12,13 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 	"text/template"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"golang.org/x/term"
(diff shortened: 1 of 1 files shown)


pmolodo:
Note that I'm not familiar with go, so apologies if there are any obvious mistakes / style issues / etc!

pmolodo:
The linter seems to be failing with:

```
Error: unnecessary conversion (unconvert)
```

However, there's no line number, so I'm not sure precisely what it's flagging.  I tried running `golangci-lint run` locally, but it seems to note be doing the same thing as the github check, because it didn't mention anything about an unnecessary conversion (but did return a bunch of other, presumably "ok", issues).

Also - let me know if you'd like me to collapse the commits before merging. Not sure what your standard protocol is...

pmolodo:
For reference, this is what I get when I run `golangci-lint run` from the chezmoi root:

```
$ golangci-lint run --out-format=github-actions
::error file=internal\cmd\util_windows.go,line=43,col=3::error is not nil (line 42) but it returns nil (nilerr)
::error file=internal\cmds\generate-install.sh\main.go,line=139,col=3::use of `fmt.Println` forbidden by pattern `^fmt\.Print.*$` (forbidigo)
::error file=internal\cmds\generate-chezmoi.io-content-docs\main.go,line=29,col=2::use of `fmt.Printf` forbidden by pattern `^fmt\.Print.*$` (forbidigo)
::error file=internal\cmds\generate-chezmoi.io-content-docs\main.go,line=45,col=4::use of `fmt.Printf` forbidden by pattern `^fmt\.Print.*$` (forbidigo)
::error file=internal\cmds\generate-chezmoi.io-content-docs\main.go,line=67,col=4::use of `fmt.Println` forbidden by pattern `^fmt\.Print.*$` (forbidigo)
::error file=internal\cmds\generate-chezmoi.io-content-docs\main.go,line=75,col=3::use of `fmt.Println` forbidden by pattern `^fmt\.Print.*$` (forbidigo)
::error file=internal\cmd\config.go,line=477,col=2::directive `//nolint:ifshort` is unused for linter "ifshort" (nolintlint)
```

Any info on how to run in the exact same manner as it is during gitlab ci check would be helpful!

pmolodo:
Cleaned up commit history, and redid commit message (capitalized initial verb) to match guidelines in https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md

twpayne:
Thank you for this @pmolodo!

The lint error is coming from `int(syscall.Stdin)`. As `syscall.Stdin` is already an `int`, the conversion is not required. To discover this, I ran `golangci-lint` locally. You can also run`make lint` to do this.

chezmoi already has code to ready strings (usernames) and passwords, so the body of the `if` block can be simplified to use chezmoi's existing functions. Rather than review all the details here, I'll merge this PR as-is so you get the contributor credit and open a follow-up PR.

pmolodo:
Weird - I definitely ran `golangci-lint` locally, as I mentioned above, and got totally different results.  I checked the go version, and it was the same.  Maybe something to do with windows vs. linux? Though I don't know why that would make such a difference.  Unless it's a conditional compilation fork kinda thing?

Anyway, thanks for figuring that out, and the merge!

Issue #885, opened 2020-09-14
Add pre-defined templates for config files of popular applications


MurzNN:
At now, when we want to add configuration of some new app to chezmoi, we must manually search needed config files and add it to chezmoi. I have found similar project "mackup", that have cool feature: pre-defined ready-to-use templates for popular applications, here is list of them: https://github.com/lra/mackup#supported-applications

Will be good to have same feature in chezmoi too, for simply type something like `chezmoi add-app blender`, instead of search for each configuration file/folder and add it manually to chezmoi.

What do you think about this idea?

twpayne:
Thank you, this is a really interesting idea. A couple of observations:

1. mackup has a really nice [configuration directory containing language-neutral configuration files](https://github.com/lra/mackup/tree/master/mackup/applications). It would be easy to import this data into chezmoi.
2. mackup is [licensed under the GPL-3](https://github.com/lra/mackup/blob/master/LICENSE) whereas chezmoi [has a MIT license](https://github.com/twpayne/chezmoi/blob/master/LICENSE). It's not clear if we can combine the two. I'm not a lawyer but an obvious question is whether mackup's configuration files are part of its source code.
3. We need to think carefully about what should happen when only some of the configuration files associated with an application are present when you run `chezmoi add-app`. Should the absent configuration files be forced to be empty, be ignored, or something else?

Thank you again for proposing this: this is potentially a very nice improvment.

twpayne:
A possible solution to point 2 is for chezmoi to read mackup's configuration files from an existing mackup installation (e.g. if you have already installed mackup with `brew install mackup`). This means that the two tools are distributed independently and should avoid licensing problems. The downside is that you have to install mackup before you can use the `add-app` command.

MurzNN:
About point 2 - maybe you can talk directly with author of mackup about "joining forces" for manage shared independent catalog of `.cfg` files with app config files info, using current file format, on separate MIT license, that will use `mackup` and `chezmoi` together (and maybe other apps in future too)?

> 3. We need to think carefully about what should happen when only some of the configuration files associated with an application are present when you run `chezmoi add-app`. Should the absent configuration files be forced to be empty, be ignored, or something else?

About missing files - I think, that good solution will be store them in `chezmoi` as some "waiting for source" mode, so check file availability on each run, and if file is absent - nothing happens, if appeared - add it to repository and start to manage, like other files.

twpayne:
[...]

PR #1325, opened 2021-08-01
Add mackup command

twpayne:
Refs #885 and https://github.com/lra/mackup/issues/1733.

First steps. Not ready for merge yet.


twpayne:
@MurzNN this PR adds

```console
$ chezmoi mackup add $APPLICATION
```

where `$APPLICATION` is the basename of the file that you'll find in `/opt/homebrew/Cellar/mackup/0.8.32/libexec/lib/python3.9/site-packages/mackup/applications`, for example `curl` or `vscode`. This assumes that you have installed mackup on your machine with `brew` and that your Python version is 3.9.something.

Right now this is now an undocumented command in chezmoi. It should work, but what's needed is:

1. Agreement from the mackup developers that it's OK to use their config files in this way.
2. Fewer assumptions about where mackup stores its config flies. Right now, this PR assumes the use of Python 3.9, but it's not a robust assumption. Ideally, the mackup developers would add a command that prints the directory name of where it stores its config files.
3. (lower priority) Deeper integration with mackup, e.g. `list` and `unmanage` commands.

twpayne:
@ecogit @drod3763 @realdimas you all 👍 'd the original issue (#885). Do you have any feedback on this implementation?

drod3763:
I don't use mackup, I just thought the templates were a good idea.

PR #1330, opened 2021-08-06
Add .chezmoi.rawHomeDir template variable using raw OS path to home dir

### LINKED ISSUE #1329: Slashes and Backslashes on Windows make symlinks a problem
## Describe the bug

Generated paths (via template) and used paths for symlinks on windows differ in `/` and `\`, leading to constant "modified" state.

## To reproduce

Create symlink to directory, e.g. 
Filename: `.\AppData\Local\symlink_nvim.tmpl`
Content:
```
{{ joinPath .chezmoi.homeDir ".config/nvim" }}
```

## Expected behavior

- ✅ A symlink is created 
- ✅ The symlink to the directory works
- ❌ Chezmoi state delievers consistent result, not complaining about changes (after applying it once)

## How does it show?
- Running `chezmoi status` ALWAYS results in chages:
```
MM AppData/Local/nvim
```
After applying this stays that way. So let's look into the "changes":

- `chezmoi apply`
```
AppData/Local/nvim has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]?
```

Let's see what actually changed, using diff:

```diff
+ C:/Users/z003hemr/.config/nvim
- C:\Users\z003hemr\.config\nvim
```
Oh no! I could change the `.config/nvim` part of it, but not the used `.chezmoi.homeDir` variable of it.

## Background / What do I want to achieve / Workarounds

Some applications have different config file paths on windows. I therefore want them to point to the very same paths as on non-windows systems using symlinks. This does work for the applications I use (in this example: nvim).
Workarounds like including different files/paths do not work for all applications.

twpayne:
This PR adds a `.chezmoi.rawHomeDir` template variable that uses the OS-specific path separator.

@derphilipp I think this should allow the creation of platform-appropriate symlinks. Something like

```
{{ joinPath .chezmoi.rawHomeDir ".config" "nvim" }}
```

should work.

Fixes #1329.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -1426,12 +1426,13 @@ chezmoi provides the following automatically-populated variables:
 | `.chezmoi.arch`         | Architecture, e.g. `amd64`, `arm`, etc. as returned by [runtime.GOARCH](https://pkg.go.dev/runtime?tab=doc#pkg-constants).      |
 | `.chezmoi.fqdnHostname` | The fully-qualified domain name hostname of the machine chezmoi is running on.                                                  |
 | `.chezmoi.group`        | The group of the user running chezmoi.                                                                                          |
-| `.chezmoi.homeDir`      | The home directory of the user running chezmoi.                                                                                 |
+| `.chezmoi.homeDir`      | The normalized home directory of the user running chezmoi.                                                                      |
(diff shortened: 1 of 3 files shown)


twpayne:
Hmm, further investigation reveals that this is not the correct fix.

derphilipp:
Thank you  very much, since 2.1.5 it works perfectly!

PR #1339, opened 2021-08-11
Fix the FreeBSD installation method

0mp:
The "ports" method of software installation
on FreeBSD is `make -C /usr/ports/sysutils/chezmoi install`.

It's better to call the method described in the documentation as
"pkg".


0mp:
Thanks!

PR #1343, opened 2021-08-13
Comparison: availability in major Linux repos

GreyCat:
Added availability of compared tools in major Linux software repositories (Debian, Ubuntu, RedHat/Fedora).


twpayne:
Thank you very much for this investigation and PR :)

There are a couple of factors here:
* Although the updates to the table are superficially accurate, some important details are not expressed. For example, installing `vcsh` on a minimal Debian Jessie install requires a 100MB download of multiple packages and gets you a version of `vcsh` dated 2014-01-06. This version of `vcsh` is more than seven years old. For this entry in the comparison table to be accurate, the feature comparison would also have to express the functionality of `vcsh` as it was seven years ago.
* Getting packages into mainstream Linux distributions takes a lot of time. chezmoi is a relatively new project, and has a one-line install on many operating systems, including non-Linux systems (e.g Windows and FreeBSD) not mentioned in this PR. Therefore, I don't think that this PR represents a fair comparison of dotfile managers in their current state.

That said, thank you again for your work and analysis :)

Issue #1357, opened 2021-08-21
Update Offers Same File Every Time


fanuch:
## What exactly are you trying to do?

Run `chezmoi update`  only with valid changes.
Update keeps showing me an empty "diff" on all systems for my vimrc file

## What have you tried so far?

`git pull; git status; git diff`

## Where else have you checked for solutions?

* [x] I have read [chezmoi's how-to guide](https://github.com/twpayne/chezmoi/blob/master/docs/HOWTO.md), and not found the answer.
* [x] I have read [chezmoi's FAQ](https://github.com/twpayne/chezmoi/blob/master/docs/FAQ.md), and not found the answer.
* [x] I have searched [chezmoi's reference guide](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md), and not found the answer.
* [x] Other, please give details. Stack Overflow

## Output of any commands you've tried with `--verbose` flag

```console
$ chezmoi update --verbose
Already up to date.
Current branch master is up to date.
.vimrc has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]? overwrite
# Rerun the same command instantly
$ chezmoi update --verbose
Already up to date.
Current branch master is up to date.
.vimrc has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]? overwrite
```

## Output of `chezmoi doctor`

```console
$ chezmoi doctor
RESULT   CHECK                MESSAGE
ok       version              v2.0.7, commit 1ce19442a526747d53317a4cd5764f4e4a60a216, built at 2021-03-31T00:25:36Z, built by goreleaser
ok       os-arch              linux/amd64
ok       config-file          /home/ari/.config/chezmoi/chezmoi.toml is a file
ok       source-dir           /home/ari/.local/share/chezmoi is a directory
ok       suspicious-entries   no suspicious entries found in /home/ari/.local/share/chezmoi
ok       dest-dir             /home/ari is a directory
ok       shell                found /usr/bin/zsh
ok       editor               found /usr/bin/vim
ok       git-cli              found /usr/bin/git, version 2.25.1
ok       merge-cli            found /usr/bin/vimdiff
ok       age-cli              age not found in $PATH
ok       gnupg-cli            found /usr/bin/gpg, version 2.2.19
info     1password-cli        op not found in $PATH
info     bitwarden-cli        bw not found in $PATH
info     gopass-cli           gopass not found in $PATH
[...]

PR #1386, opened 2021-09-03
Fix install from source documentation

twpayne:


twpayne:
Many thanks to @fanuch who [spotted](https://github.com/twpayne/chezmoi/issues/1357#issuecomment-912256607) that the current docs were completely broken.

PR #1388, opened 2021-09-05
Add onepasswordItemFields template function

### LINKED ISSUE #1098: 1password op get item --fields support

Current `op` command has a `get item` with `--fields` flag which makes it a lot simpler to extract values from details or **section fields**.

The `output` function can be used with this:

```
{{- output "op" "get" "item" "some-id" "--fields" "some-field" }}
```

But it might be nicer to have a built-in _onepassworditemFields_ function (or something) to make this a bit nicer to integrate?

twpayne:
Fixes #1098.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -109,6 +109,7 @@ Manage your dotfiles across multiple machines, securely.
   * [`onepassword` *uuid* [*vault-uuid* [*account-name*]]](#onepassword-uuid-vault-uuid-account-name)
   * [`onepasswordDocument` *uuid* [*vault-uuid* [*account-name*]]](#onepassworddocument-uuid-vault-uuid-account-name)
   * [`onepasswordDetailsFields` *uuid* [*vault-uuid* [*account-name*]]](#onepassworddetailsfields-uuid-vault-uuid-account-name)
+  * [`onepasswordItemFields` *uuid* [*vault-uuid* [*account-name*]]](#onepassworditemfields-uuid-vault-uuid-account-name)
   * [`output` *name* [*arg*...]](#output-name-arg)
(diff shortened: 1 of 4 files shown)


bric3:
Neat !

docs/REFERENCE.md
```diff
@@ -2116,6 +2117,57 @@ accounts).
...
+        "k": "string",
+        "n": "D4328E0846D2461E8E455D7A07B93397",
+        "t": "exampleLabel",
+        "v": "exampleValue"
+    }
+}
+```
+
+---
+
```
bric3, line 2170: I think the readme should provide the equivalent commands

```shell
op get item abcdefghijklmnopqrstuvwxyz --fields "exampleLabel"
exampleValue
```

```go-template
{{ (onepasswordItemFields "abcdefghijklmnopqrstuvwxyz").exampleLabel.v }}
```

bric3:
Hi,

After looking at the code I wonder if this approach is future proof in the sens that 1password could change their structure payload. Why not using the command flag `--fields` (of course it means chezmoi will require a minimum version of the `op` tool).

docs/REFERENCE.md
twpayne: Added in a79ba2f0699ddbc1245715da12975daec18db513.

twpayne:
If 1password change the structure of the return value they will break a lot of things for other people, so I don't think they would do it. The reason for using the JSON output rather than the command flag is that the JSON output has a well-defined format and multiple requests for different aspects of the same item only result in one underlying call to `op`.

bric3:
Indeed this is a remote possibility. Note that I don't actually have preference in term of usage.

PR #1406, opened 2021-09-12
Change gitlab link for other people's dotfiles

Grimler91:
The previous link (https://gitlab.com/search?search=chezmoi) for some
reason only turns up one result (at least for me).  The new link
(https://gitlab.com/explore/projects?topic=chezmoi) instead searches
for repositories that are tagged "chezmoi", which finds six gitlab
repositories for me.


Grimler91:
I believe https://gitlab.com/search?search=chezmoi used to find more repositories, so maybe gitlab has changed their search algorithm

twpayne:
Thanks very much for this. I also think that GitLab's search behavior has changed. The new link is nicer as well because it seems to show repos that have been explicitly tagged with `chezmoi`.

PR #1422, opened 2021-09-17
Vagrant CI fixes

twpayne:


twpayne:
cc @tklauser this is the fix for the Vagrant Cloud rate limiting. Each cached Vagrant Box is ~1GB but GitHub's actions/cache gives you 5GB of cache by default.

PR #1444, opened 2021-09-22
Support paths containing spaces in VIM integration

Dekker1:
Paths of files managed by chezmoi can contain spaces, such as configuration files in `~/Library/Application Support` on MacOS. The current Vim integration command would fail as spaces in the path where not escaped. The usage of quotation marks should avoid this problem.


twpayne:
Awesome, thanks!

Issue #891, opened 2020-09-24
Support a root directory other than the repository's root


felipecrs:
## The problem

I use my dotfiles repository extensively, not only for dotfiles, but other things that relates to my machines management. Despite `chezmoi` supports ignoring files through the `.chezmoiignore`, I would like to sort my repository by moving all my real dotfiles to a folder called `home` in the repository (just like `homeshick`), so I would not need to manage the `.chezmoiignore` anymore.

## Solution

A way to set the default root for the source dotfiles, other than the repository root. Probably a new key in the `.chezmoi.toml.tmpl`

## Alternatives

I though on setting `sourceFolder` to the sub-folder that I want, but I wonder if this won't break any other `chezmoi` functionality such as `chezmoi update` since the `sourceFolder` would not be the git root anymore.

twpayne:
Would it work to put all your non-home directory stuff in a directory called `stuff` and then add `stuff` to `.chezmoiignore`?

felipecrs:
Unfortunately not, many of the files that I'm referring to are supposed to stay in the repository root. Such as:

```
README.md
LICENSE
install # Installer for Codespaces
.gitpod.*
.vscode/
.devcontainer/
```
https://github.com/felipecrs/dotfiles/blob/master/.chezmoiignore

felipecrs:
This would also help newcomers from homeshick and homesick to migrate.

explorador:
I actually tried setting the `sourceFolder` to a sub-directory and it works fine for me, even `chezmoi update`

felipecrs:
Ok, so I implemented your suggestion, @explorador at https://github.com/felipecrs/dotfiles/pull/32. It's working so far. 

@twpayne:

The only issue that I can think of is with `chezmoi init` for initially installing the dotfiles. The current syntax does not seem to support using a subfolder of the git repo:

https://www.chezmoi.io/docs/reference/#init-repo

felipecrs:
Actually, there is a major issue with this approach:

Every time I call `chezmoi init`, chezmoi creates the `~/.dotfiles/home/.git` folder. Ideally, chezmoi should detect that its home folder is already in a git project.

felipecrs:
Moving from https://github.com/twpayne/chezmoi/issues/1423#issuecomment-923408966.

@twpayne I can try to propose a PR to fix it but unfortunately not today, most likely tomorrow. Please don't wait for me if you are willing to try my suggestion, as it should be very simple anyway.

[...]

PR #1448, opened 2021-09-22
Detect when the source directory is already in a git working copy

twpayne:
Refs #891.

cc @felipecrs 

I don't think this fixes everything, but it should address [this comment about `chezmoi init`](https://github.com/twpayne/chezmoi/issues/891#issuecomment-922379179).

### DIFF
--- a/internal/cmd/initcmd.go
+++ b/internal/cmd/initcmd.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/fs"
 	"os"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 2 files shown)


internal/cmd/initcmd.go
```diff
@@ -133,9 +133,22 @@ func (c *Config) runInitCmd(cmd *cobra.Command, args []string) error {
 		c.init.purgeBinary = true
 	}
 
-	// If the source repo does not exist then init or clone it.
-	switch _, err := c.baseSystem.Stat(c.SourceDirAbsPath.Join(".git")); {
-	case errors.Is(err, fs.ErrNotExist):
+	// Find the working copy.
+	workingCopyDirAbsPath := c.SourceDirAbsPath
+FOR:
+	for workingCopyDirAbsPath != "/" {
```
felipecrs, line 139: This reminds me of https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders. It should work!

I would just have delegated the work to `git` instead, as it is already a dependency of chezmoi anyways.

twpayne: chezmoi actually doesn't have a hard dependency on git, see the [`--use-builtin-git` flag](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md#--use-builtin-git-value).

felipecrs: Hm... what I understand by `--user-builtin-git` is that chezmoi ships a git binary inside of it. This mean git is always available, even when not installed in the computer. No?

felipecrs:
This is awesome, @twpayne. If you were waiting for me for release 2.5.0, please go ahead!

internal/cmd/initcmd.go
twpayne: It embeds [github.com/go-git/go-git](https://github.com/go-git/go-git), which is an implementation of git in a Go library. It means that limited git functionality (`init` and `update`) is available, even if git is not installed on the machine. The main reason for adding this was to avoid requiring the user to install git on ephemeral machines before being able to install their dotfiles.

felipecrs: Oh okay! Thank you for the explanation! Now I understand.

PR #1458, opened 2021-09-24
chore: Remove unused and undocumented mackup command

### LINKED ISSUE #885: Add pre-defined templates for config files of popular applications

twpayne:
Fixes #885.


nkgm:
Really wish I'd come across this a couple of months earlier. Would this still work if merged in the current version?

twpayne:
Yes, this would work. However, do consider:
1. The mackup developers are unresponsive and do not seem keen to collaborate.
2. The mackup data quality is OK but not brilliant. It's not complete for a number of applications.

nkgm:
Those are fair points. I guess it's easy enough adding the files manually, though I do wish we had file completion on `chezmoi re-add` with just the modified files (or simply the files chezmoi is already managing) as these app config file paths are easy to forget.

Thanks for your work on chezmoi btw. It's hard to imagine how I've managed without it for so long :)

twpayne:
Thank you for your kind words!

> I do wish we had file completion on `chezmoi re-add` with just the modified files (or simply the files chezmoi is already managing) as these app config file paths are easy to forget.

This is a great suggestion, although it might be a little tricky to implement. Could you please open a new issue to track this?

eugenesvk:
I've also just become aware of the mackup project and its app database, and it seems like a great idea (even though it's a bit late for me as I've had to muck ;) through those apps' config files manually to pick the ones that should be synced via chezmoi)

I know you mentioned
> It seems that nobody else needs or wants the feature

but maybe a big part of that is that "nobody" was aware of its existence? Maybe this could be restored to some documented experimental command to allow for people to know about it and see if incorporating it into their workflows improves config management?

twpayne:
The other problems are:

> * The mackup developers are unresponsive and do not seem keen to collaborate.
> * The mackup data quality is OK but not brilliant. It's not complete for a number of applications.

Overall, I don't think that mackup really adds that much. If enough (say five) different people chime in here to say they want this feature restored, then I would consider reverting this PR.

eugenesvk:
Would you then kindly reopen https://github.com/twpayne/chezmoi/issues/885 so that at least this issue is a bit more visible so people could chime in (fwiw it already had 4 👍  before it got locked)

> The mackup developers are unresponsive and do not seem keen to collaborate

That's unfortunate, chezmoi is a great and powerful tool to collaborate with :)!

> The mackup data quality is OK but not brilliant. It's not complete for a number of applications

That's still better than the baseline of having to manually pick specific config files in a sea of caches and data files and what not?

twpayne:
> Would you then kindly reopen #885 so that at least this issue is a bit more visible so people could chime in (fwiw it already had 4 👍 before it got locked)

Done. Note that, despite the 4 👍 s, and a direct request, no one tested the feature.

PR #1459, opened 2021-09-24
feat: Add working tree config option

### LINKED ISSUE #891: Support a root directory other than the repository's root

twpayne:
Fixes #891.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -22,6 +22,7 @@ Manage your dotfiles across multiple machines, securely.
   * [`--use-builtin-git` *value*](#--use-builtin-git-value)
   * [`-v`, `--verbose`](#-v---verbose)
   * [`--version`](#--version)
+  * [`-w`, `--working-tree` *directory*](#-w---working-tree-directory)
 * [Common command line flags](#common-command-line-flags)
   * [`-f`, `--format` `json`|`yaml`](#-f---format-jsonyaml)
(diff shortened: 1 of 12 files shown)


felipecrs:
This is interesting. So now, I should probably add a `workingTree` to my chezmoi config template?

twpayne:
It should be optional. chezmoi should automatically detect the working tree, but there could well be places where it does not work as expected. For sure, if `sourceDir` is a subdirectory of your working tree (e.g. `home`, as in your case) then you need to set `sourceDir` in your config file, but you /shouldn't/ need to set `workingTree`. Please tell me if you find any unexpected behavior.

felipecrs:
Ok, will do. Thank you!

Issue #1108, opened 2021-03-28
Having issues with getting values from `chezmoi` in templates


mhmdanas:
## What exactly are you trying to do?

Trying to write a template for a configuration file which has different locations on different systems. I have a file in `.chezmoitemplates`; let's call it `file.toml`. Here are the file's contents (I've removed all unnecessary details):
```toml
[foo]
bar = "{{ chezmoi.homeDir }}"
```
I then have the actual config file with the following content:
```
{{- template "file.toml" }}
```
The issue is that `{{ chezmoi.homeDir }}` doesn't work, giving me the following error:
```
chezmoi: template: file.toml:2:18: executing "file.toml" at <.chezmoi.homeDir>: nil data; no entry for key "chezmoi"
```

## What have you tried so far?

Running `chezmoi execute-template "{{ .chezmoi.homeDir }}"` works just fine.

## Where else have you checked for solutions?

* [x] I have read [chezmoi's how-to guide](https://github.com/twpayne/chezmoi/blob/master/docs/HOWTO.md), and not found the answer.
* [x] I have read [chezmoi's FAQ](https://github.com/twpayne/chezmoi/blob/master/docs/FAQ.md), and not found the answer.
* [x] I have searched [chezmoi's reference guide](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md), and not found the answer.
* [x] I searched in the issue tracker and found nothing similar.

## Output of any commands you've tried with `--verbose`

```
$ chezmoi apply --verbose
chezmoi: template: file.toml:2:18: executing "file.toml" at <.chezmoi.homeDir>: nil data; no entry for key "chezmoi"
```

## Output of `chezmoi doctor`

```
$ chezmoi doctor
RESULT    CHECK                MESSAGE
ok        version              v2.0.4, commit 9cc8184cc466a87c44006218e02365046e5d5f8b, built at 2021-03-24T18:55:30Z, built by goreleaser
ok        os-arch              linux/amd64
ok        config-file          /home/mhmdanas/.config/chezmoi/chezmoi.toml is a file
ok        source-dir           /home/mhmdanas/.local/share/chezmoi is a directory
ok        suspicious-entries   no suspicious entries found in /home/mhmdanas/.local/share/chezmoi
ok        dest-dir             /home/mhmdanas is a directory
ok        shell                found /home/mhmdanas/.cargo/bin/nu
ok        editor               found /usr/bin/nano
[...]

PR #1470, opened 2021-09-29
docs: fix error in HOWTO guide

mhmdanas:
The "different files locations on different systems" guide didn't pass data to the template file, which can result in errors. See https://github.com/twpayne/chezmoi/issues/1108 for an example of this problem.


twpayne:
Awesome, thank you for the fix!

Issue #1430, opened 2021-09-18
`chezmoi init` should advise when multiple configuration files are present


felipecrs:
## Is your feature request related to a problem? Please describe.

Recently I switched my `.chezmoi.toml.tmpl` to YAML format, and then I found this issue:

`chezmoi` does not delete the previously generated `.config/chezmoi/chezmoi.toml` when it inits `.config/chezmoi/chezmoi.yaml`. But I believe it should. The `yaml` is newer, and the user expects it to be effective immediately.

## Describe the solution you'd like

Maybe `chezmoi` should delete the previously generated configuration file (but only if it was generated by chezmoi itself? because users can have written data to it manually).

Maybe `chezmoi` should at least advise the user that a new configuration file was created but another is already present and must be removed.

## Describe alternatives you've considered

I always generate the chezmoi configuration automatically, I never touch it manually. So, I wrote:

```
# .chezmoiremove

{{ if stat (joinPath .chezmoi.homeDir ".config/chezmoi/chezmoi.yaml") -}}
.config/chezmoi/chezmoi.toml
{{ end -}}
```

And it does the job for removing the old configuration. But there are some caveats:

The running `chezmoi` apply will still be using the old loaded config in the memory. This means that I have to call `apply` two times to make everything is good. It would be simpler and more deterministic if this was solved at the `init` time.

twpayne:
Thanks for reporting this. chezmoi's support for config files in multiple formats (using github.com/spf13/viper) does cause a lot of ambiguity here. Furthermore, chezmoi supports the XDG Base Directory specification, which also allows multiple configuration directories. So, there could potentially be tens of config files. chezmoi's documentation states explicitly that it uses the first config file it finds.

Implementing a warning when there are multiple config files will require a scan of all possible config filenames in all possible config directories on every invocation. To be honest, it takes quite a special set of circumstances to trigger this, i.e. not many users will be converting their config file formats.

Instead, I propose to add the check for a unique config file to the `chezmoi doctor` command. Users should run this before reporting issues, which should allow them to catch this when it occurs.

felipecrs:
Agreed 100%!

felipecrs:
But there is one thing that probably would not be too risky:

chezmoi init could verify if the extension of the config template matches the extension of the actual config file that chezmoi readed.

And if it does not match a message could be raised.

But regardless, this is just an idea, I already have a working workaround and I'm probably the only one who faced this issue. So I'm completely fine with either way forward.

[...]

PR #1472, opened 2021-09-29
feat: Add --init option to update command

### LINKED ISSUE #1427: `chezmoi update --init`

I often update my dotfiles, but I have to call two commands always:

```
$ chezmoi update --apply=false

$ chezmoi init --apply --force
```

It would be much easier if I could simply call:

```
chezmoi update --init --force
```

I would not have to handle `--apply=false` in the first and `--apply` in the later.

The 2 commands approach still works.

## Additional context

None.

twpayne:
Fixes #1427.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -1593,6 +1593,11 @@ Pull changes from the source repo and apply any changes.
 
 Only update entries of type *types*.
 
+#### `--init`
+
+Regenerate and reread the config file from the config file template before
+applying any changes.
+
 #### `update` examples
 
 ```console
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -551,6 +551,101 @@ func (c *Config) colorAutoFunc() bool {
 	return false
 }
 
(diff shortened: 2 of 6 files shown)


twpayne:
@felipecrs A quick head's up on this. This PR fixes the immediate issue, but there's a more subtle underlying problem that you as an advanced user of chezmoi might encounter.

Very roughly, chezmoi's reading config code looks like this:
1. Set defaults.
2. Read config from user.
3. Compute extra config variables derived from user config (e.g. guess `workingTree` from `sourceDir` if `workingTree` is not explicitly set, refs #1459).

This `--init` flag to `chezmoi update` only performs step 2 above and does not perform step 3. This means that if you're changing directories or removing template variables (and maybe other things) then the config used by `chezmoi update --init` might differ from the config used by the next invocation of `chezmoi update`.

In practice, as long as the config changes are small this should not be a problem, but I know that you occasionally make big config changes (e.g. #1430), so I wanted to warn you.

felipecrs:
Interesting, I will have this in mind. Thanks a lot for this new feature!

Issue #769, opened 2020-05-26
Persist file permissions for group and other


thernstig:
First, as I have not said this yet, great tool! I was evaluating all kind of various dotfile tools and found this one to be the most well-thought implementation. Great job 😃 

## Is your feature request related to a problem? Please describe.

By example:

```sh
# In my home directory
~> ll testingprivexec2.txt
.rwx--x--x 7 userA groupA 24 May 10:56 testingprivexec2.txt*
~> chmod 711 testingprivexec2.txt
~> chezmoi add testingprivexec2.txt
~> chezmoi apply
~> ll testingprivexec2.txt
.rwxrwxrwx 7 userA groupA 24 May 10:56 testingprivexec2.txt*
```

If I set a specific `group` or `other` permission on a file in my home directory that is controlled by chezmoi, that permission is not persisted after a `chezmoi apply`. By experiment I've found it to use the umask, which does feel natural as that is how it works e.g. for `git clone`.

It would be nice though if all file permissions could be persisted. It is possible this is not in the scope of chezmoi.

## Describe the solution you'd like

I have not investigated this deeply, but here is an outlandish idea that at least works i git. Use [git notes](https://git-scm.com/docs/git-notes) e.g. like shown here https://stackoverflow.com/a/5960171. This could also be used to store the private and executable info, so files did not have to get renamed to `executable_` or `private_` etc. 

## Describe alternatives you've considered

No alternatives considered.

twpayne:
Thank you for your kind words about chezmoi :)

> It would be nice though if all file permissions could be persisted. It is possible this is not in the scope of chezmoi.

It's true that chezmoi has a very simple model of permissions: private/not private and executable/not executable. `chezmoi add` considers a file executable if any of the user/group/other permissions are executable, and considers a file private if neither group nor other have any permissions.

This does mean that permissions like `711` are not preserved precisely: chezmoi considers such a file to be not private but executable, which means that when you run `chezmoi apply` it will be given permissions `777 &^ umask`, usually `755` when `umask` is `022`.

At the same time, `711` permissions on a file are slightly strange, and I genuinely am not sure how different operating systems would treat such a file. Are you allowed to execute a file without reading its contents? Of course, it's different for directories where `r` means reading contents and `x` means that you are allowed to traverse it.

Could you give a concrete example of file permissions that chezmoi doesn't handle (like `711`) that you'd like to preserve and why you want to preserve them?

[...]

Issue #752, opened 2020-05-16
apply file changes on write with edit --apply


jmattheis:
My current workflow for changing my config is 
```
$ chezmoi edit --apply ~/.config/some/config
```
then I've to write quit my `$EDITOR` to apply the changes. It would be cool, if I didn't have to close my editor and only write the file (`:w` in vim). Chezmoi could listen on file changes (with something like inotify) and then apply them.

Usecases are:
* vim: edit config, write, source applied config from chezmoi
* i3: edit config, write, inplace reload i3 without killing existing windows

This is only a minor inconvenience so feel free to close the issue if you don't see the need for it.

Anyway, keep up the good work with chezmoi!

twpayne:
Interesting idea, and related to what @Zackhardtoname is [asking for](https://github.com/twpayne/chezmoi/issues/748#issuecomment-629262971).

This should be fairly straightforward to implement, but I'd like to get the v2 rework done first. I expect to add this though :)

Lilja:
So I randomly looked into this thread and felt a bit bored and decided to create my first vim plugin. 

It solves this problem by getting the current PID of the vim process, looks up the parent PID and sees if chezmoi was invoked(`chezmoi edit $dotfile`). If so, it will listen to buffer write and invoke the `chezmoi apply $dotfile` automatically.

Feel free to use it! https://github.com/lilja/vim-chezmoi

camgraff:
+1 for this! I'd love to get back the ability to make a change to my .vimrc, save, re-source, and test if my change results in the expected functionality all without ever heaving to leave the editor. 

@twpayne if you're too busy right now and this doesn't conflict with the v2 rework, I'd be happy to get a PR up for this. Id' just need a little direction on the design i.e. would you want this to become the default behavior or introduce a new flag with `chezmoi edit`?

twpayne:
@camgraff Thanks for your offer - it's not quite time to work on this yet.

[...]

PR #1481, opened 2021-10-01
feat: Make editors see target filenames

twpayne:
This PR makes `chezmoi edit $TARGET` invoke your editor with an unmangled filename matching `$TARGET` exactly, independent of chezmoi's file name mangling.

For example, say you have a `private_dot_bashrc` in your source state. When you run `chezmoi edit ~/.bashrc` then your editor will be opened with a file with the basename `.bashrc`, so all of your editor's syntax highlighting should work.

Under the hood it creates a hard link from a file in a temporary directory (e.g.`$TMPDIR/.bashrc`) to the actual file in your source directory (e.g. `~/.local/share/chezmoi/private_dot_bashrc`) so your editor thinks it's actually editing `$HOME/.bashrc`.

After you quit your editor the directory with the hard links is cleaned up.

Refs #752 #769
cc @alker0 @m-rey @thernstig

This is an experimental feature - I'd love to hear your feedback.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -72,8 +72,9 @@ you can run `chezmoi diff` to check what effect the changes would have, and run
 `chezmoi apply` if you're happy with them.
 
 `chezmoi edit` provides the following useful features:
-* It opens the correct file in the source state for you, so you don't have to
-  know anything about source state attributes.
+* It opens the correct file in the source state for you with a filename matching
+  the target filename, so your editor's syntax highlighting will work and you
+  don't have to know anything about source state attributes.
 * If the dotfile is encrypted in the source state, then `chezmoi edit` will
   decrypt it to a private directory, open that file in your `$EDITOR`, and then
   re-encrypt the file when you quit your editor. That makes encryption more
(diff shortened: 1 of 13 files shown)


thernstig:
Will this work with specifying an entire directory as the `$TARGET`. My current workflow is to open the entire `.local/share/chezmoi/` in VS Code to edit files. If I could use `chezoi edit .local/share/chezmoi/` for it to unmangle all files it would be cool. Not sure that would work though.

twpayne:
No, this only works on files, not directories. There's no real way that it can work with directories because:
* For this to work with directories, chezmoi would have to create a complete mirror tree structure of the directory containing hardlinks for each file, which would be expensive.
* Even if chezmoi were to do this, there are conflicts between files in the source state and files in the working tree (e.g. `dot_gitignore` in the source state conflicts with `.gitignore` in the working tree).

VIM users can use [github.com/alker0/chezmoi.vim](https://github.com/alker0/chezmoi.vim) to get syntax highlighting in the source directory. For VSCode, it's likely that an equivalent plugin for VSCode would be needed.

Issue #1495, opened 2021-10-05
"create" prefix causes a template to not be processed


Peterdes:
Problem: no file starting with create_ is being processed as a template.
Example: create a file named "create_dumb.tmpl" with contents
```
{{ .chezmoi.os }}
```
then run `chezmoi apply -nv`. The template is not processed.

It is because `newCreateTargetStateEntryFunc` in `internal/chezmoi/sourcestate.go` does not involve any template processing.

The behavior is fixed by the following change:
```
diff --git a/internal/chezmoi/sourcestate.go b/internal/chezmoi/sourcestate.go
index 15fc1e64..c78184ea 100644
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -1184,7 +1184,7 @@ func (s *SourceState) newSourceStateDir(sourceRelPath SourceRelPath, dirAttr Dir
 // newCreateTargetStateEntryFunc returns a targetStateEntryFunc that returns a
 // file with sourceLazyContents if the file does not already exist, or returns
 // the actual file's contents unchanged if the file already exists.
-func (s *SourceState) newCreateTargetStateEntryFunc(fileAttr FileAttr, sourceLazyContents *lazyContents) targetStateEntryFunc {
+func (s *SourceState) newCreateTargetStateEntryFunc(sourceRelPath SourceRelPath, fileAttr FileAttr, sourceLazyContents *lazyContents) targetStateEntryFunc {
 	return func(destSystem System, destAbsPath AbsPath) (TargetStateEntry, error) {
 		contents, err := destSystem.ReadFile(destAbsPath)
 		switch {
@@ -1194,6 +1194,12 @@ func (s *SourceState) newCreateTargetStateEntryFunc(fileAttr FileAttr, sourceLaz
 			if err != nil {
 				return nil, err
 			}
+			if fileAttr.Template {
+				contents, err = s.ExecuteTemplateData(sourceRelPath.String(), contents)
+				if err != nil {
+					return nil, err
+				}
+			}
 		default:
 			return nil, err
 		}
@@ -1386,7 +1392,7 @@ func (s *SourceState) newSourceStateFile(sourceRelPath SourceRelPath, fileAttr F
 	var targetStateEntryFunc targetStateEntryFunc
 	switch fileAttr.Type {
 	case SourceFileTypeCreate:
-		targetStateEntryFunc = s.newCreateTargetStateEntryFunc(fileAttr, sourceLazyContents)
[...]

PR #1496, opened 2021-10-05
fix: Allow create_ entries to be templates

twpayne:
Refs #1495.


twpayne:
Co-authored by @Peterdes.

PR #1521, opened 2021-10-11
feat: Add decrypt and encrypt template functions

twpayne:
cc @azmodude the `decrypt` function will help you [only encrypt the sensitive parts of files](https://twitter.com/azmodude/status/1446934800577748996).

I don't think the `encrypt` function will be used much, but it's there for completeness.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -98,6 +98,8 @@ Manage your dotfiles across multiple machines, securely.
   * [`bitwarden` [*arg*...]](#bitwarden-arg)
   * [`bitwardenAttachment` *filename* *itemid*](#bitwardenattachment-filename-itemid)
   * [`bitwardenFields` [*arg*...]](#bitwardenfields-arg)
+  * [`decrypt` *ciphertext*](#decrypt-ciphertext)
+  * [`encrypt` *plaintext*](#encrypt-plaintext)
   * [`gitHubKeys` *user*](#githubkeys-user)
   * [`gopass` *gopass-name*](#gopass-gopass-name)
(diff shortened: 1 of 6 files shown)


azmodude:
You, Sir, are an absolute legend! Thanks for this feature originating from a Tweet I did not even mention you properly in. :)

Did some tests just now, works beautifully by dropping encrypted include files in my source state, including and decrypting them via templates and ignoring them in my .chezmoiignore file at the same time to never leave source state.
This also opens up a lot of possibilities for me having an encrypted include to my .chezmoi.toml.tmpl and setting some private variables there to be later inserted into files.

Thanks again!

twpayne:
Thanks for the kind words!

> This also opens up a lot of possibilities for me having an encrypted include to my .chezmoi.toml.tmpl and setting some private variables there to be later inserted into files.

Note that there might be a chicken-and-egg problem here: when running `chezmoi init` on a new machine, encryption will not initially be configured. Encryption will only be available once the config file has been written and re-read by chezmoi.

azmodude:
Most deserved!

> Note that there might be a chicken-and-egg problem here: when running chezmoi init on a new machine, encryption will not initially be configured. Encryption will only be available once the config file has been written and re-read by chezmoi.

Yup, stumbled upon that issue. What I am doing for now is I am writing a very basic `chezmoi.toml` in my bootstrapping script that sets everything up (fetch GPG key, install op, install chezmoi, ...) initially to include my encryption method and key. The subsequent 'real' `chezmoi init` is then able to pick that minimal config up and replace it with the full decrypted version of the file. Works quite well for me.

PR #1603, opened 2021-11-09
feat: Add .chezmoiroot to allow source state to be subdir of source dir

### LINKED ISSUE #1519: Support `.chezmoiroot` file to redirect `init`'s source dir to subfolder

This is a follow-up of https://github.com/twpayne/chezmoi/issues/891#issuecomment-933742656. The problem of having to run

```
sh <(curl -fsSL https://git.io/chezmoi) -- init --apply --working-tree=~/.local/share/chezmoi --source=~/.local/share/chezmoi/home felipecrs
```

Is that it's a long command, difficult to remember, and breaks the fact that `chezmoi init` would not work by default on all the repositories.

A new file named `.chezmoiroot` (the name is debatable), which I would put in the root of my dotfiles repository and would have contents like so:

```
$ cat .chezmoiroot
home
```

It would tell chezmoi that, when initializing the repository, it should adjust the `sourceDir` to be the `home` subfolder automatically.

This would allow me to remove some workarounds that I have to circumvent this, such as [this](https://github.com/felipecrs/dotfiles/commit/9cafef4d988ce652c0fcf9ad1ea9fc24b3f58c46#diff-1e142e6277b12b7e1110478a24caee8f006a9349e86970c890203d6266209463R35).

I think this is the only thing missing to declare first-class support for subfolders in the dotfiles directory as the source dir. With it in place, the following would work:

```console
sh <(curl -fsSL https://git.io/chezmoi) -- init --apply felipecrs
```

Described earlier.

## Additional context

I'm quite happy with my dotfiles being in a subdirectory. Previously, I oftenly created wrong files in my home dir because I forgot to add to .chezmoiignore, now I don't have to remember this step anymore.

twpayne:
Fixes #1519.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -54,6 +54,7 @@
   * [Run a PowerShell script as admin on Windows](#run-a-powershell-script-as-admin-on-windows)
 * [Use chezmoi with GitHub Codespaces, Visual Studio Codespaces, or Visual Studio Code Remote - Containers](#use-chezmoi-with-github-codespaces-visual-studio-codespaces-or-visual-studio-code-remote---containers)
 * [Customize chezmoi](#customize-chezmoi)
+  * [Use a subdirectory of your dotfiles repo as the root of the source state](#use-a-subdirectory-of-your-dotfiles-repo-as-the-root-of-the-source-state)
   * [Don't show scripts in the diff output](#dont-show-scripts-in-the-diff-output)
   * [Customize the diff pager](#customize-the-diff-pager)
   * [Use a custom diff tool](#use-a-custom-diff-tool)
@@ -1599,6 +1600,37 @@ sudo apt install -y vim-gtk
 ---
 
 ## Customize chezmoi
+
+---
+
+### Use a subdirectory of your dotfiles repo as the root of the source state
+
+By default, chezmoi uses the root of your dotfiles repo as the root of the
+source state. If your source state contains many entries in its root, then your
+target directory (usually your home directory) will in turn be filled with many
+entries in its root as well. You can reduce the number of entries by keeping
+`.chezmoiignore` up to date, but this can become tiresome.
+
+Instead, you can specify that chezmoi should read the source state from a
+subdirectory of the source directory instead by creating a file called
+`.chezmoiroot` containing the relative path to this subdirectory.
+
+For example, if `.chezmoiroot` contains:
+
+```
+home
+```
+
+Then chezmoi will read the source state from the `home` subdirectory of your
+source directory, for example the desired state of `~/.gitconfig` will be read
+from `~/.local/share/chezmoi/home/dot_gitconfig` (instead of
+`~/.local/share/chezmoi/dot_gitconfig`).
+
+When migrating an existing chezmoi dotfiles repo to use `.chezmoiroot` you will
+need to move the relevant files in to the new root subdirectory manually. You do
+not need to move files that are ignored by chezmoi in all cases (i.e. are listed
+in `.chezmoiignore` when executed as a template on all machines), and you can
+afterwards remove their entries from `home/.chezmoiignore`.
 
 ---
 
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -1284,6 +1284,16 @@ func (c *Config) newRootCmd() (*cobra.Command, error) {
 // newSourceState returns a new SourceState with options.
 func (c *Config) newSourceState(ctx context.Context, options ...chezmoi.SourceStateOption) (*chezmoi.SourceState, error) {
 	sourceStateLogger := c.logger.With().Str(logComponentKey, logComponentValueSourceState).Logger()
+
+	sourceDirAbsPath := c.SourceDirAbsPath
+	switch data, err := c.sourceSystem.ReadFile(c.SourceDirAbsPath.JoinString(chezmoi.RootName)); {
+	case errors.Is(err, fs.ErrNotExist):
+	case err != nil:
+		return nil, err
+	default:
+		sourceDirAbsPath = c.SourceDirAbsPath.JoinString(string(bytes.TrimSpace(data)))
+	}
+
 	s := chezmoi.NewSourceState(append([]chezmoi.SourceStateOption{
 		chezmoi.WithBaseSystem(c.baseSystem),
 		chezmoi.WithCacheDir(c.CacheDirAbsPath),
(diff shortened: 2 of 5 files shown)


docs/HOWTO.md
```diff
@@ -1601,6 +1602,36 @@ sudo apt install -y vim-gtk
...
 
+### Use a subdirectory of your dotfiles repo as the root of the source state
+
+By default, chezmoi uses the root of your dotfiles repo as the root of the
+source state. If your source state contains a lot of entries in its root, then
+your dotfiles repo will in turn contain a lot of entries in its root.
+
+You can instead specify that chezmoi should read the source state from a
+subdirectory of the source directory instead by creating a file called
+`.chezmoiroot` containing the relative path the source state. This reduces the
```
felipecrs, line 1613:
```suggestion
`.chezmoiroot` containing the relative path to the source state. This reduces the
```

docs/HOWTO.md
```diff
@@ -1601,6 +1602,36 @@ sudo apt install -y vim-gtk
 
 ---
 
+### Use a subdirectory of your dotfiles repo as the root of the source state
+
+By default, chezmoi uses the root of your dotfiles repo as the root of the
+source state. If your source state contains a lot of entries in its root, then
+your dotfiles repo will in turn contain a lot of entries in its root.
```
felipecrs, line 1609:
```suggestion
your target folder (usually your home) will in turn be filled with a lot of entries in its root as well.
```

docs/HOWTO.md
```diff
@@ -1601,6 +1602,36 @@ sudo apt install -y vim-gtk
...
+
+You can instead specify that chezmoi should read the source state from a
+subdirectory of the source directory instead by creating a file called
+`.chezmoiroot` containing the relative path the source state. This reduces the
+number of files in the root of your dotfiles repo and reduces the number of
```
felipecrs, line 1614:
```suggestion
number of files in the root of your target folder and reduces the number of
```

```suggestion
number of
```

docs/HOWTO.md
```diff
@@ -1601,6 +1602,36 @@ sudo apt install -y vim-gtk
 
 ---
 
+### Use a subdirectory of your dotfiles repo as the root of the source state
+
+By default, chezmoi uses the root of your dotfiles repo as the root of the
+source state. If your source state contains a lot of entries in its root, then
+your dotfiles repo will in turn contain a lot of entries in its root.
+
```
felipecrs, line 1610: It's probably a good idea to first explain the first solution, which would be to keep the `.chezmoiignore` updated. And also to mention why or where it would be a better idea to use `.chezmoiroot`. E.g.: with `.chezmoiroot`, if you add a new file to the root of the repository, you don't need to remember to manually add it to the `.chezmoiignore`.

docs/HOWTO.md
```diff
@@ -1601,6 +1602,36 @@ sudo apt install -y vim-gtk
...
+You can instead specify that chezmoi should read the source state from a
+subdirectory of the source directory instead by creating a file called
+`.chezmoiroot` containing the relative path the source state. This reduces the
+number of files in the root of your dotfiles repo and reduces the number of
+files that you need to explicitly exclude with `.chezmoiignore`.
```
felipecrs, line 1615: But if you add that paragraph about explaining chezmoignore's approach, this last sentence may not be needed anymore.

docs/HOWTO.md
```diff
@@ -1601,6 +1602,36 @@ sudo apt install -y vim-gtk
...
+
+For example, if `.chezmoiroot` contains:
+
+```
+home
+```
+
+Then chezmoi will read the source state from the `home` subdirectory of your
+source directory, for example the desired state of `~/.gitconfig` will be read
+from `~/.local/share/chezmoi/home/dot_gitconfig`.
```
felipecrs, line 1625: Just to make it even more clear.

```suggestion
from `~/.local/share/chezmoi/home/dot_gitconfig` (instead of `~/.local/share/chezmoi/dot_gitconfig`).
```

internal/cmd/config.go
```diff
@@ -1278,6 +1278,16 @@ func (c *Config) newRootCmd() (*cobra.Command, error) {
...
 	sourceStateLogger := c.logger.With().Str(logComponentKey, logComponentValueSourceState).Logger()
+
+	sourceDirAbsPath := c.SourceDirAbsPath
+	switch data, err := c.sourceSystem.ReadFile(sourceDirAbsPath.JoinString(chezmoi.RootName)); {
+	case errors.Is(err, fs.ErrNotExist):
+	case err != nil:
+		return nil, err
+	default:
+		sourceDirAbsPath = sourceDirAbsPath.JoinString(string(bytes.TrimSpace(data)))
+	}
```
felipecrs, line 1289: I'm wondering... shouldn't `workingTree` be used as base here rather than `sourceDir` itself? I don't know. Did you think about it? The answer is probably no, but it's better to make sure we thought on everything.

internal/cmd/testdata/scripts/root.txt
```diff
@@ -0,0 +1,17 @@
...
+cmp $CHEZMOISOURCEDIR/home/symlink_dot_symlink golden/symlink_dot_symlink
+
+-- golden/.file --
+# contents of .file
+-- golden/symlink_dot_symlink --
+.file
+-- home/user/.local/share/chezmoi/.chezmoiroot --
+home
+-- home/user/.local/share/chezmoi/home/dot_file --
+# contents of .file
```
felipecrs, line 17: perhaps it worth it to add a test for `init` with a config template as well?

felipecrs (commented):
It's probably good to go. None of my suggestions are actually required, they are just to fine-tune.

internal/cmd/testdata/scripts/root.txt
twpayne: Done (in my working copy, not pushed yet).

docs/HOWTO.md
twpayne: Done (in working copy).

docs/HOWTO.md
twpayne: Done (in working copy).

internal/cmd/config.go
twpayne: This is an extremely good point, and I think you're right. I've updated the code and the docs to interpret `.chezmoiroot` as relative to the working tree instead of the source directory.

With these three directories in play (source dir, working tree, and chezmoi root) then it's likely that I've missed some corner cases. Currently the tests pass, but the tests are not exhaustive. If there are some tests that demonstrate problems with the working dir approach then I would be very happy to add them :)

twpayne:
PR updated, please take another look.

internal/cmd/config.go
felipecrs: My only remaining concern is: if the user set the `sourceDir` in the configuration file, it will take precedence. Right?

twpayne: On further thought, I think `.chezmoiroot` has to be read from the source directory, otherwise the `--source` option will be ignored if `.chezmoiroot` is present in the working tree. Unfortunately with the flag parsing library that chezmoi uses (github.com/spf13/cobra) I don't think it's possible to distinguish between "the user did not specify the flag so it's set to the default value" and "the user did specify the flag and passed the default value".

felipecrs: I agree.

> I don't think it's possible to distinguish between "the user did not specify the flag so it's set to the default value" and "the user did specify the flag and passed the default value".

I suppose there will not be a need for that if chezmoi reads chezmoiroot from source directory rather than from working tree.

PR #1617, opened 2021-11-12
chore: Update Makefile to match conventional standards

### LINKED ISSUE #1610: Add Packaging usecase to Makefile
The current Makefile is intended to be used by the CI to produce downloadable binaries. This is counter productive for most Operating system packaging workflows. As concrete example. In OpenIndiana (the illumos reference distribution) we build software by checking out the git sources of oi-userland and downloading a source tarball into a specific directory into that tree. We then run the Makefile with build and install targets (as is common) and set the output directory to a custom created directory tree where the Makefile is supposed to install the binary. This tree is then used to create the Package manifest which is the textfile containing which files and metadata is going to be installed. To make this process work, we need to patch the Makefile to Hardcode the version number and disable cross-compilation. see https://github.com/OpenIndiana/oi-userland/blob/5629024fa7e6411597f32f6b40374e720f4e3d13/components/sysutils/chezmoi/patches/01-only-build-illumos.patch 

The Source tarball should contain a Makefile suitable for packaging. This means the Makefile follows the following conventions.
- The Makefile does not call git to find what version it is building.
- The Version can be overridden via Environment variable.
- The `install` target supports `DESTDIR` variable
- The default target only builds the source for the current Host os (no cross compilation) or if the Makefile has a documented build target for that purpose. 

Additionally the Source Tarball should contain a Packaging specific document giving the following instructions to the reader
- What additional software must be installed for this software to build
- What additional software can be installed to enable additional features and what flags are needed to compile that additional software
- Which commands to use to build for the target Host os and install the binary in a directory tree of their choosing for example 
```bash
make build
DESTDIR=${PROTO_DIR} make install
```

twpayne:
Fixes #1610.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,14 +8,14 @@ default: run build-all test lint format
 install:
 	go install -ldflags "-X main.version=$(shell git describe --abbrev=0 --tags) \
 		-X main.commit=$(shell git rev-parse HEAD) \
-		-X main.date=$(shell date -u +%Y-%m-%dT%H:%M:%SZ) \
+		-X main.date=$(shell git show -s --format=%ct HEAD) \
 		-X main.builtBy=source"
 
 .PHONY: build
 build:
 	go build -ldflags "-X main.version=$(shell git describe --abbrev=0 --tags) \
 		-X main.commit=$(shell git rev-parse HEAD) \
-		-X main.date=$(shell date -u +%Y-%m-%dT%H:%M:%SZ) \
+		-X main.date=$(shell git show -s --format=%ct HEAD) \
 		-X main.builtBy=source"
 
 .PHONY: build-all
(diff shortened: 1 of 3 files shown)


Toasterson:
Sorry, I think I was not very clear on `DESTDIR`. `GOBIN` expects the fullpath whereas `DESTDIR` in autotools appends the final path onto `DESTDIR`. For example if we set `DESTDIR` to `/home/proto` then the final path would be `/home/proto/usr/bin/chezmoi` by convention. So the Logic that checks for the `DESTDIR` variable would need to append `/usr/bin` so `GOBIN` works nicely.

twpayne:
Thank you for spotting this. I've updated the PR to respect `$DESTDIR` as you describe, and also added `$PREFIX` option (default `/usr/local`) so the final installation location is `${DESTDIR}${PREFIX}/bin/chezmoi`.

twpayne:
For info the Windows CI failure is #1622 and is not related to this PR.

Toasterson:
Awesome LGTM

Toasterson (approved):

PR #1534, opened 2021-10-18
feat: Add .chezmoi.args template variable

twpayne:
Fixes #1532.

`.chezmoi.executable` is the requested `.chezmoi.binPath`.

`.chezmoi.args` combine the requested `.chezmoi.command` and `.chezmoi.options`.

I really don't think that this is a good idea, but here's a PR for you to experiment with :)

PR #1618, opened 2021-11-12
feat: Add .chezmoi.executable template variable

felipecrs:
So, it turns out that `.chezmoi.args[0]` is not the path for the chezmoi binary as I thought. It was during my test because I was indeed calling the chezmoi by providing the full path to its executable.

Refs https://github.com/twpayne/chezmoi/pull/1534

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -451,25 +451,25 @@ names.
 The following prefixes and suffixes are special, and are collectively referred
 to as "attributes":
 
-| Prefix       | Effect                                                                          |
-| ------------ | ------------------------------------------------------------------------------- |
-| `after_`     | Run script after updating the destination.                                      |
-| `before_`    | Run script before updating the destination.                                     |
-| `create_`    | Ensure that the file exists, and create it with contents if it does not.        |
-| `dot_`       | Rename to use a leading dot, e.g. `dot_foo` becomes `.foo`.                     |
-| `empty_`     | Ensure the file exists, even if is empty. By default, empty files are removed.  |
-| `encrypted_` | Encrypt the file in the source state.                                           |
-| `exact_`     | Remove anything not managed by chezmoi.                                         |
-| `executable_`| Add executable permissions to the target file.                                  |
-| `literal_`   | Stop parsing prefix attributes.                                                 |
-| `modify_`    | Treat the contents as a script that modifies an existing file.                  |
-| `once_`      | Only run the script if it has not been run before.                              |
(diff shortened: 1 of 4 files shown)


docs/REFERENCE.md
```diff
@@ -1853,6 +1853,7 @@ chezmoi provides the following automatically-populated variables:
 | -------------------------- | ---------- | ------------------------------------------------------------------------------------------------------------------------------- |
 | `.chezmoi.arch`            | `string`   | Architecture, e.g. `amd64`, `arm`, etc. as returned by [runtime.GOARCH](https://pkg.go.dev/runtime?tab=doc#pkg-constants).      |
 | `.chezmoi.args`            | `[]string` | The arguments passed to the `chezmoi` command, starting with the program command.                                               |
+| `.chezmoi.executable` | `string` | The path to the `chezmoi` executable, if available. |
```
twpayne, line 1856: Please align the `|`s here.

internal/cmd/testdata/scripts/templatevars.txt
```diff
@@ -9,6 +9,9 @@ chezmoi execute-template '{{ .chezmoi.arch }}'
 chezmoi execute-template '{{ index .chezmoi.args 1 }}'
 stdout execute-template
 
+chezmoi execute-template '{{ .chezmoi.executable }}'
+stdout ^.*[\\/]bin[\\/]chezmoi(.exe)?$
```
twpayne, line 13: The `^.*` at the start of the regexp doesn't have any effect: it matches any leading characters from the start of the string, which is the same as if `^.*` was not present. I'm not sure that it's guaranteed that the executable will be in a `bin` directory, so I'd remove the `[\\/]bin[\\/]` part. Overall I think the regexp should be `chezmoi(.exe)?$`.

twpayne (commented):
Thanks - just a couple of minor changes requested.

docs/REFERENCE.md
felipecrs: I just had to disable the auto generation of toc on save:

https://github.com/twpayne/chezmoi/pull/1618/files#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686R21-R22

And I also removed a deprecated configuration (the terminal).

twpayne (approved):
Perfect, thank you!

docs/REFERENCE.md
felipecrs: Ops, I initially pushed with `"markdown.extension.toc.updateOnSave": true` but in fact it was meant to be `false`, as they TOCs seems to be generated by the website itself.

PR #1636, opened 2021-11-17
fix: Return unparsed gopass output in gopassRaw template func

twpayne:
Refs #1635.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -2077,11 +2077,11 @@ once.
 
 ### `gopassRaw` *gopass-name*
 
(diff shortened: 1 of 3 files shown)


twpayne:
cc @davidolrik

davidolrik (changes requested):
You should remove the `--password` flag as that only returns the first line.

davidolrik (approved):
Functionality looks good!

twpayne:
Thanks for the review :)

davidolrik:
Thanks for the quick fix!

PR #1643, opened 2021-11-19
feat: Add conditional requests GitHub Client

### LINKED ISSUE #1642: Caching the results of the github api locally

Multiple calls to template functions of a short period of time , such as multiple `diff` operations , may make multiple calls to the `github api` , thus triggering [Rate Limit](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) .

store the response and `Etag` of the github api in [persistentState](https://github.com/twpayne/chezmoi/blob/434c29f49711f35a90bf07831c678f02b6136b95/internal/chezmoi/persistentstate.go#L18) like `githubApiStateBucket` to avoid repeated calls

n/a

## Additional context

Ref: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests

ps: I would like to try to implement this feature

MuXiu1997:
Fixes #1642.

This implementation seems too invasive of the `persistentState`, and I'm not sure it's the good way to go
Looking for your advice!

cc @twpayne

### DIFF
--- a/internal/cmd/githubtemplatefuncs.go
+++ b/internal/cmd/githubtemplatefuncs.go
@@ -1,13 +1,26 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
+	"io"
+	"net/http"
 
 	"github.com/google/go-github/v40/github"
+
+	"github.com/twpayne/chezmoi/v2/internal/chezmoi"
 )
+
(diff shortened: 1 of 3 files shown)


twpayne (commented):
Thanks for this. Let's discuss a bit more in #1642 first.

PR #1647, opened 2021-11-20
feat: Add cache HTTP client

### LINKED ISSUE #1646: Add caching http client

Add a caching http client for [`getExternalDataRaw`](https://github.com/twpayne/chezmoi/blob/8e1d6478d2c7f076dcd72bca1fb90b0acccf0516/internal/chezmoi/sourcestate.go#L1094) and `GitHub API`

Replace `http.DefaultClient` with [gregjones/httpcache](https://github.com/gregjones/httpcache)

n/a

## Additional context

_Originally posted by @MuXiu1997 in https://github.com/twpayne/chezmoi/issues/1642#issuecomment-974609464_

MuXiu1997:
Fixes #1646.

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -77,6 +77,7 @@ type SourceState struct {
 	encryption              Encryption
 	ignore                  *patternSet
 	interpreters            map[string]*Interpreter
+	httpClient              *http.Client
 	logger                  *zerolog.Logger
 	minVersion              semver.Version
 	mode                    Mode
@@ -119,6 +120,13 @@ func WithDestDir(destDirAbsPath AbsPath) SourceStateOption {
 func WithEncryption(encryption Encryption) SourceStateOption {
 	return func(s *SourceState) {
 		s.encryption = encryption
+	}
+}
+
+// WithHTTPClient sets the HTTP client.
+func WithHTTPClient(httpClient *http.Client) SourceStateOption {
+	return func(s *SourceState) {
+		s.httpClient = httpClient
 	}
 }
 
@@ -202,6 +210,7 @@ func NewSourceState(options ...SourceStateOption) *SourceState {
 		umask:                Umask,
 		encryption:           NoEncryption{},
 		ignore:               newPatternSet(),
+		httpClient:           http.DefaultClient,
 		logger:               &log.Logger,
 		readTemplateData:     true,
 		priorityTemplateData: make(map[string]interface{}),
@@ -1092,9 +1101,6 @@ func (s *SourceState) executeTemplate(templateAbsPath AbsPath) ([]byte, error) {
 // getExternalDataRaw returns the raw data for external at externalRelPath,
 // possibly from the external cache.
 func (s *SourceState) getExternalDataRaw(ctx context.Context, externalRelPath RelPath, external External, options *ReadOptions) ([]byte, error) {
-	// FIXME be more intelligent about HTTP caching, e.g. following RFC 7234,
-	// rather than blindly re-downloading each time
-
 	var now time.Time
 	if options != nil && options.TimeNow != nil {
 		now = options.TimeNow()
@@ -1122,7 +1128,7 @@ func (s *SourceState) getExternalDataRaw(ctx context.Context, externalRelPath Re
 	if err != nil {
 		return nil, err
 	}
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := s.httpClient.Do(req)
 	s.logger.Err(err).
 		Str("method", req.Method).
 		Int("statusCode", resp.StatusCode).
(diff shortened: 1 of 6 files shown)


internal/cmd/config.go
```diff
@@ -1296,13 +1297,16 @@ func (c *Config) newSourceState(ctx context.Context, options ...chezmoi.SourceSt
 		sourceDirAbsPath = c.SourceDirAbsPath.JoinString(string(bytes.TrimSpace(data)))
 	}
 
+	httpClient := newCacheHTTPClient(c.CacheDirAbsPath.Join(httpCacheDirRelPath).String())
```
twpayne, line 1300: Let's store the `HTTPClient` in the config (`c`) so we can re-use it elsewhere.

internal/cmd/config.go
```diff
@@ -1296,13 +1297,16 @@ func (c *Config) newSourceState(ctx context.Context, options ...chezmoi.SourceSt
...
+
 	s := chezmoi.NewSourceState(append([]chezmoi.SourceStateOption{
 		chezmoi.WithBaseSystem(c.baseSystem),
 		chezmoi.WithCacheDir(c.CacheDirAbsPath),
 		chezmoi.WithDefaultTemplateDataFunc(c.defaultTemplateData),
 		chezmoi.WithDestDir(c.DestDirAbsPath),
 		chezmoi.WithEncryption(c.encryption),
 		chezmoi.WithInterpreters(c.Interpreters),
+		chezmoi.WithHTTPClient(httpClient),
```
twpayne, line 1309: Please keep the `chezmoi.With` functions sorted alphabetically.

internal/cmd/upgradecmd.go
```diff
@@ -240,7 +240,8 @@ func (c *Config) downloadURL(ctx context.Context, url string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	resp, err := http.DefaultClient.Do(req)
+	httpClient := newCacheHTTPClient(c.CacheDirAbsPath.Join(httpCacheDirRelPath).String())
```
twpayne, line 243: Here we can use the `*HTTPClient` that we store `c` (see earlier comment).

internal/cmd/util.go
```diff
@@ -72,6 +74,12 @@ func firstNonEmptyString(ss ...string) string {
 	return ""
 }
 
+// newCacheHTTPClient returns a new http.Client that will cache responses
+// according to the HTTP RFC.
+func newCacheHTTPClient(cacheDirAbsPath string) *http.Client {
```
twpayne, line 79: nit: Through the code variables with the type `chezmoi.AbsPath` the name `*AbsPath`. Please change the type of the argument from `string` to `chezmoi.AbsPath`. The reason for using `AbsPath` instead of `string` is to improve type safety through the code. There's a bit more explanation in the [architecture guide](https://github.com/twpayne/chezmoi/blob/master/docs/ARCHITECTURE.md#path-handling).

twpayne (commented):
Thank you - this looks fantastic. I've left a few very minor review comments.

internal/chezmoi/sourcestate.go
```diff
@@ -115,6 +116,13 @@ func WithDestDir(destDirAbsPath AbsPath) SourceStateOption {
 	}
 }
 
+// WithHTTPClient sets the HTTP client.
```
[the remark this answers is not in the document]
MuXiu1997: Oh no, I noticed this at first and then forgot about it

internal/cmd/config.go
MuXiu1997: I agree, I think so too
At first I wanted to do this, but I was worried that you might not like to change the `config`

internal/chezmoi/sourcestate.go
```diff
@@ -77,6 +77,7 @@ type SourceState struct {
 	encryption              Encryption
 	ignore                  *patternSet
 	interpreters            map[string]*Interpreter
+	httpClient              *http.Client
 	logger                  *zerolog.Logger
 	minVersion              semver.Version
 	mode                    Mode
```
MuXiu1997, line 83: @twpayne I'm not sure it's the right place for `httpClient`

internal/chezmoi/sourcestate.go
```diff
@@ -202,6 +210,7 @@ func NewSourceState(options ...SourceStateOption) *SourceState {
 		umask:                Umask,
 		encryption:           NoEncryption{},
 		ignore:               newPatternSet(),
+		httpClient:           http.DefaultClient,
 		logger:               &log.Logger,
 		readTemplateData:     true,
 		priorityTemplateData: make(map[string]interface{}),
```
MuXiu1997, line 216: @twpayne I'm not sure it's the right place for `httpClient`

twpayne: Nor am I, but it's a perfectly reasonable place. The main reason I tend to order things alphabetically is that it means I no longer have to make a decision about where to add something, or which way to scroll to find something. In `struct`s though, where I generally want to keep related fields together, it's less clear.

Issue #1650, opened 2021-11-20
`toYaml` template function


felipecrs:
## Is your feature request related to a problem? Please describe.

I'm now handling some YAML files, and a `toYaml` function would come very in handy.

## Describe the solution you'd like

The same `toYaml` function which Helm provides to templates.

## Describe alternatives you've considered

To not use YAML, but instead JSON.

## Additional context

https://github.dev/helm/helm/blob/dd0177aed71dd3c5b044e25c55870cd66f0f05fa/pkg/engine/funcs.go#L88

It may be worth to add some more functions that helm adds too, like `toToml`, `fromYaml`, etc.

felipecrs:
My specific use case, in case you wonder, is:

`home/private_dot_config/private_chezmoi/private_root.config.yaml.tmpl`:

```yaml
sourceDir: {{ joinPath .chezmoi.sourceDir "root" | quote }}
destDir: "/"

{{ $chezmoiData := unset . "chezmoi" }}

data:
  {{ .chezmoi.data | toYaml | indent 2 }}
```

But of course it shouldn't be limited to that.

PR #1658, opened 2021-11-23
feat: Add fromYaml template function

felipecrs:
Refs #1654
Refs #1650

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -105,6 +105,7 @@ Manage your dotfiles across multiple machines, securely.
   * [`bitwardenFields` [*arg*...]](#bitwardenfields-arg)
   * [`decrypt` *ciphertext*](#decrypt-ciphertext)
   * [`encrypt` *plaintext*](#encrypt-plaintext)
+  * [`fromYaml` *yamltext*](#fromyaml-yamltext)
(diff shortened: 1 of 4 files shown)


twpayne (commented):
There's no need to limit this to maps. `fromYaml` should accept any value.

docs/REFERENCE.md
```diff
@@ -2023,6 +2024,18 @@ the same arguments will only invoke `bw get` once.
 
 ---
 
+### `fromYaml` *yamltext*
+
+`fromYaml` returns a dict representation of *yamltext*.
```
twpayne, line 2029: This comment needs updating: the parsed value is returned.

internal/cmd/templatefuncs.go
```diff
@@ -19,6 +19,15 @@ type ioregData struct {
 	value map[string]interface{}
 }
 
+func (c *Config) fromYamlTemplateFunc(str string) interface{} {
+	var result interface{}
+
+	if err := chezmoi.FormatYAML.Unmarshal([]byte(str), &result); err != nil {
+		returnTemplateError(err)
```
twpayne, line 26: There should be a `return nil` after this line.

twpayne (approved):
Looks great, thanks!

PR #1659, opened 2021-11-23
feat: Add execTemplate template function

felipecrs:
Fixes #1619

PR #1661, opened 2021-11-23
feat: Add execTemplate template function

### LINKED ISSUE #1619: New template function for capturing the output of the `template` function

I have:

```console
❯ cat ~/.dotfiles/home/.chezmoitemplates/get-latest-github-release
{{- $url :=  printf "https://github.com/%s/releases/latest" . -}}
{{- output "curl" "-fsSL" "-o" "/dev/null" "-w" "%{url_effective}" $url | trim | base -}}
```

And I would like to use it like so:

```console
❯ cat ~/.dotfiles/home/.chezmoiexternal.yaml
{{ $denoVersion := template "get-latest-github-release" "denoland/deno" -}}
".deno/bin/deno":
  type: file
  url: "https://github.com/denoland/deno/releases/download/{{ $denoVersion }}/deno-x86_64-unknown-linux-gnu.zip"
  executable: true
  filter:
    command: zcat
```

But it's not possible, because:

```console
$ chezmoi execute-template '{{ $denoVersion := template "get-latest-github-release" "denoland/deno" }}{{ $denoVersion }}'
chezmoi: template: arg1:1: unexpected <template> in command
```

EDIT 1: I previously posted with the wrong error message above, which turns to be valid but posterior to the error originally reported, as pointed by @twpayne at [here](https://github.com/twpayne/chezmoi/issues/1619#issuecomment-967649423).

EDIT 2: In the variable assignment, it was missing `:` before `=` as pointed by @twpayne at [here](https://github.com/twpayne/chezmoi/issues/1619#issuecomment-967688720). But that wasn't the cause of the issue, anyway.

A function which would let me do that, as pointed in https://stackoverflow.com/a/40170999/12156188. So that I could do:

```console
$ chezmoi execute-template '{{ $denoVersion := execTemplate "get-latest-github-release" "denoland/deno" }}{{ $denoVersion }}'
v1.16.1
```

I could not find any.

## Additional context

My dotfiles is currently bloated with a lot of code duplication, and now I'm trying to clean it up by using some templates.

MuXiu1997:
Fixes #1619
This PR is a continuation of #1659, made by @felipecrs

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -105,6 +105,7 @@ Manage your dotfiles across multiple machines, securely.
   * [`bitwardenFields` [*arg*...]](#bitwardenfields-arg)
   * [`decrypt` *ciphertext*](#decrypt-ciphertext)
   * [`encrypt` *plaintext*](#encrypt-plaintext)
+  * [`execTemplate` *string* *data*](#exectemplate-string-data)
   * [`gitHubKeys` *user*](#githubkeys-user)
   * [`gitHubLatestRelease` *user-repo*](#githublatestrelease-user-repo)
   * [`gopass` *gopass-name*](#gopass-gopass-name)
@@ -2020,6 +2021,21 @@ the same arguments will only invoke `bw get` once.
 ### `encrypt` *plaintext*
 
 `encrypt` encrypts *plaintext* using chezmoi's configured encryption method.
+
+---
+
+### `execTemplate` *string* *data*
+
+`execTemplate` executes the defined *string* template using *data* as parameter
+and returns a string with the result. This allows the result of a template to be
+captured in a variable, or piped to another function (which is not possible by
+using the built-in `template` directive).
+
+#### `execTemplate` examples
+
+```
+{{ define "test" }}{{ . }}{{ end }}{{ $var := execTemplate "test" "123" }}{{ $var }}
+```
 
 ---
 
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -86,6 +86,7 @@ type SourceState struct {
 	userTemplateData        map[string]interface{}
 	priorityTemplateData    map[string]interface{}
 	templateData            map[string]interface{}
+	templateFuncBuilders    map[string]func(tmpl *template.Template) interface{}
 	templateFuncs           template.FuncMap
 	templateOptions         []string
 	templates               map[string]*template.Template
@@ -183,6 +184,15 @@ func WithSystem(system System) SourceStateOption {
 func WithDefaultTemplateDataFunc(defaultTemplateDataFunc func() map[string]interface{}) SourceStateOption {
 	return func(s *SourceState) {
 		s.defaultTemplateDataFunc = defaultTemplateDataFunc
+	}
+}
+
+// WithTemplateFuncBuilders sets the template function builders.
+func WithTemplateFuncBuilders(
+	templateFuncBuilders map[string]func(tmpl *template.Template) interface{},
+) SourceStateOption {
+	return func(s *SourceState) {
+		s.templateFuncBuilders = templateFuncBuilders
 	}
 }
 
@@ -636,12 +646,24 @@ func (s *SourceState) Encryption() Encryption {
 	return s.encryption
 }
 
+// newTemplate returns the parsed template.Template.
+func (s *SourceState) newTemplate(name, text string) (*template.Template, error) {
+	tmpl := template.New(name)
+	for name, templateFuncBuilder := range s.templateFuncBuilders {
+		s.templateFuncs[name] = templateFuncBuilder(tmpl)
+	}
+	tmpl, err := tmpl.Option(s.templateOptions...).
+		Funcs(s.templateFuncs).
+		Parse(text)
+	if err != nil {
+		return nil, err
+	}
+	return tmpl, nil
+}
+
 // ExecuteTemplateData returns the result of executing template data.
 func (s *SourceState) ExecuteTemplateData(name string, data []byte) ([]byte, error) {
-	tmpl, err := template.New(name).
-		Option(s.templateOptions...).
-		Funcs(s.templateFuncs).
-		Parse(string(data))
+	tmpl, err := s.newTemplate(name, string(data))
 	if err != nil {
 		return nil, err
 	}
@@ -1072,7 +1094,7 @@ func (s *SourceState) addTemplatesDir(templatesDirAbsPath AbsPath) error {
 			}
 			templateRelPath := templateAbsPath.MustTrimDirPrefix(templatesDirAbsPath)
 			name := templateRelPath.String()
-			tmpl, err := template.New(name).Option(s.templateOptions...).Funcs(s.templateFuncs).Parse(string(contents))
+			tmpl, err := s.newTemplate(name, string(contents))
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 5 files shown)


MuXiu1997:
Fixes #1619

felipecrs:
You did all the hard work. I woul take years to figure this all out.

Before merging I suggest you squash the 2 commits in one, and add me as co-author in the commit message:

Co-Authored-By: Felipe Santos <felipecassiors@gmail.com>

MuXiu1997:
> Before merging I suggest you squash the 2 commits in one, and add me as co-author in the commit message:
> 
> Co-Authored-By: Felipe Santos [felipecassiors@gmail.com](mailto:felipecassiors@gmail.com)
Great suggestion! First time I've done this, I'll try it

MuXiu1997:
@felipecrs I tried to do it, but I don't know if I did it right

felipecrs:
> @felipecrs I tried to do it, but I don't know if I did it right

image

Yeap, you did it. But you can remove your name, that's a no-op. Github will already count you as you are the author of the commit itself.

felipecrs:
Oh, actually it should probably be like this:

```diff
- Co-Authored-By: Felipe Santos felipecassiors@gmail.com
+ Co-authored-by: Felipe Santos <felipecassiors@gmail.com>
```

Here is the official reference: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

MuXiu1997:
Maybe my name will be missing after the merger because I kept the author as you
I'd better add it just in case

felipecrs:
> Maybe my name will be missing after the merger because I kept the author as you I'd better add it just in case

It won't be... what you did was to add me as "co-author". The author can be seen with `git log`, which will show your name. Also, this little icon with both my avatar and yours proves that it works:

image

twpayne:
Thank you for your work here, but this fundamentally goes against the design of `text/template`. See https://github.com/twpayne/chezmoi/issues/1662#issuecomment-977486224 for more details.

felipecrs:
@twpayne there is a lot of other functions to aid in templating, like `output`, I wonder why not to add one more. Are you concerned about maintenance burden?

The problem of creating an external script to do the job is that it would not be cross-platform, and that `text/template` is almost good enough (it only misses a function like this).

felipecrs:
I think my bottom-line is: I know that `text/template` is not a programming language, but it is used like one for many different reasons. This is what powers chezmoi the most: to allow advanced techniques in a dotfiles repository, directly from within a source file.

I'm sure no one will ever use `chezmoi` to build a webserver (as reference to "become PHP 3"), but if we can provide a template function (this `execTemplate` one is nothing but just one more in the existing set of template functions) that would help dotfiles maintainers to reduce the code duplication in their repository, without adding any drawback, I can't see why not.

MuXiu1997:
In fact, I am more agree with @twpayne 's point of view
People who want to simply use chezmoi can still simply use
Some of the complex features we want can be achieved with some external tricks, and the existing features of chezmoi are enough to achieve what we discussed

felipecrs:
I am really not sure if it's correct to count the infinite things that `output` can do as "existing features of chezmoi". If it were, the `gitHubLatestRelease` wasn't supposed to be needed too, as it could be simply implemented externally.

MuXiu1997:
You have a point there, but I didn't realize at first that outputs could be used that way, especially with the addition of `toYaml`, which is why I suggested adding `gitHubLatestRelease`

felipecrs:
Also, as my (probable) last argument to have this new `execTemplate` function, I say that `helm` provides such functionality in their `include` function, which is a bit different than chezmoi's one.

https://helm.sh/docs/chart_template_guide/#the-include-function

https://stackoverflow.com/a/52417369

So, we would not be the first ones to implement such a thing on top of `text/template`, especially considering that helm is a well stablished and big project.

Perhaps we should just tweak chezmoi's include function to behave more like the helm's include function? As of now, I think chezmoi's include function is a subset of helm's as helm's include function can also be used to read external files.

felipecrs:
> which is why I suggested adding `gitHubLatestRelease`

Do not get me wrong please, I'm totally up for adding new functions to chezmoi if they can simplify the life of the dotfiles maintainers.

MuXiu1997:
> Also, as my (probable) last argument to have this new execTemplate function, I say that helm provides such functionality in their include function, which is a bit different than chezmoi's one.

This is a powerful point, but the final decision is still in the hands of @twpayne

twpayne:
> @twpayne there is a lot of other functions to aid in templating, like `output`, I wonder why not to add one more. Are you concerned about maintenance burden?

My concern is that adding functions like `execTemplate` will result in complex and fragile systems with poorly-defined behavior, and that poorly-defined behavior will have to be maintained in perpetuity for backwards compatibility.

> Also, as my (probable) last argument to have this new `execTemplate` function, I say that `helm` provides such functionality in their `include` function, which is a bit different than chezmoi's one.

I don't think Helm is a good example to follow. Just because one *can* add functions like `execTemplate`, I don't think one *should*. Fundamentally, text templates are a poor way to generate structured data, which is why Helm is such a complex mess. Instead, structured data should be generated, e.g. using [CUE](https://cuelang.org/), a custom Python script, or something else. Fundamentally, structured data should be generated by something that understands the structure of the data, not from plain text.

twpayne:
I should also add: for sophisticated automation, I would prefer that people write scripts *around* chezmoi, rather than *in* chezmoi.

felipecrs:
> and that poorly-defined behavior will have to be maintained in perpetuity for backwards compatibility.

In my perspective, that's just a function that executes `template` and allows the output to be captured as a _string_. There is nothing complicated involved, nothing probable to change or stuff like that. I'm supposing it's the kind of function we can add and forget, as there isn't anything to be improved on it.

twpayne:
What should be the output of the following?

```
{{ define "foo" }}
  {{ execTemplate "foo" . }}
{{ end }}
{{ execTemplate "bar" }}
```

What about:

```
{{ define "foo" }}
  {{ execTemplate "bar" . }}
{{ end }}
{{ define "bar" }}
  {{ execTemplate "foo" . }}
{{ end }}
{{ execTemplate "foo" "baz" }}
```

felipecrs:
I suppose the same as:

```
{{ define "foo" }}
  {{ template "foo" . }}
{{ end }}
{{ template "bar" }}
```

```
{{ define "foo" }}
  {{ template "bar" . }}
{{ end }}
{{ define "bar" }}
  {{ template "foo" . }}
{{ end }}
{{ template "foo" "baz" }}
```

felipecrs:
Currently:

```console
❯ ./chezmoi execute-template '{{ define "foo" }}
  {{ template "foo" . }}
{{ end }}
{{ template "foo" "bar" }}'
chezmoi: template: arg1:2:14: executing "foo" at <{{template "foo" .}}>: exceeded maximum template depth (100000)

❯ ./chezmoi execute-template '{{ define "foo" }}
  {{ execTemplate "foo" . }}
{{ end }}
{{ execTemplate "foo" "bar" }}'
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc0215fe370 stack=[0xc0215fe000, 0xc0415fe000]
fatal error: stack overflow
```

twpayne:
Indeed, this is the current behavior. At no point was the desired behavior of `execTemplate` defined.

felipecrs:
> Indeed, this is the current behavior. At no point was the desired behavior of `execTemplate` defined.

But is this really a situation which needs to be handled? Isn't the current behavior good enough? Both outputs an error, so that the dotfiles owner is advised of the problem and can fix it.

PR #1669, opened 2021-11-26
chore: Update Makefile

MuXiu1997:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: ;
+
 GO?=go
 GOLANGCI_LINT_VERSION=$(shell grep GOLANGCI_LINT_VERSION: .github/workflows/main.yml | awk '{ print $$2 }')
 ifdef VERSION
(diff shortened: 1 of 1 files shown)


twpayne (commented):
What does this change do and how does it work? From my reading of the GNU Make Reference Manual this change makes all targets in the Makefile not PHONY, which is undesirable.

MuXiu1997:
@twpayne 
```diff
+ .PHONY: ;
+
```
it makes all targets in the Makefile PHONY
Just what we need

twpayne:
> @twpayne
> 
> ```diff
> + .PHONY: ;
> +
> ```
> 
> it makes all targets in the Makefile PHONY Just what we need

Could you point to the part of the GNU Make Reference Manual that describes this behavior? I couldn't find it.

MuXiu1997:
I was wrong, I learned this from elsewhere, but didn't verify that it would work
Indeed, it doesn't work, sorry

MuXiu1997:
ref: https://github.com/photoprism/photoprism/blob/7940786ce95faa0692e1834ed9123e40a8cf62b7/Makefile#L1-L5

PR #1672, opened 2021-11-26
feat: Make upgrade command work without calling GitHub API

felipecrs:
I converted my dotfiles to install chezmoi using the binary directly rather than using Homebrew. In the moment I applied these changes in my work environment, I noticed that chezmoi upgrade command fails in the very beginning.

At my work environment we do not use GitHub, and therefore it's not desirable to ask my colleagues to create a GitHub account just for running chezmoi upgrade.

image

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -1750,16 +1750,9 @@ $ chezmoi update
 ### `upgrade`
 
 Upgrade chezmoi by downloading and installing the latest released version. This
-will call the GitHub API to determine if there is a new version of chezmoi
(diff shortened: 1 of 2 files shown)


felipecrs:
WIP

twpayne:
Before you spend any more time on this, please be aware that I won't accept this. I'm aware of this way of getting around GitHub's API rate limiting without using a GitHub API token, but I want to keep chezmoi as a "good citizen".

felipecrs:
@twpayne I totally understand. I sent a message to GitHub support asking about this, I wonder what they think about it. I'll close the PR for now.

PR #1674, opened 2021-11-26
feat: Add exit template function

twpayne:
Refs #1670.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -107,6 +107,7 @@ Manage your dotfiles across multiple machines, securely.
   * [`bitwardenFields` [*arg*...]](#bitwardenfields-arg)
   * [`decrypt` *ciphertext*](#decrypt-ciphertext)
   * [`encrypt` *plaintext*](#encrypt-plaintext)
+  * [`exit` *code*](#exit-code)
   * [`fromYaml` *yamltext*](#fromyaml-yamltext)
   * [`gitHubKeys` *user*](#githubkeys-user)
   * [`gitHubLatestRelease` *user-repo*](#githublatestrelease-user-repo)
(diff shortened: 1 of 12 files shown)


twpayne:
I think the test failures are with Go 1.16, which doesn't seem to wrap template errors. The tests pass on Go 1.17.

As chezmoi supports the last two versions of Go (currently 1.16 and 1.17), we can only merge this once Go 1.18 is released. Converting to draft.

docs/REFERENCE.md
```diff
@@ -2606,6 +2607,12 @@ used.
 
 ---
 
+### `stop`
+
+`stop` stops template execution, without returning an error.
```
felipecrs, line 2612: Ops, in fact I did need to error (making the chezmoi call return a non-zero code). Perhaps this function can accept a parameter like `exitCode`, which defaults to 0 and can be changed on a need basis?

felipecrs (commented):
Sounds amazing, except that for my use case I would also need to make chezmoi to exit with a non-zero code. I suggested one approach.

docs/REFERENCE.md
twpayne: OK, PR updated. PTAL.

felipecrs:
This is brilliant! Just what I needed, thank you so much!

felipecrs:
Regarding Go 1.16, just for curiosity, is there any place using it to build chezmoi? Like external package managers or something like that? It's really just for curiosity, I'm not asking to change anything.

twpayne:
> Regarding Go 1.16, just for curiosity, is there any place using it to build chezmoi?

Yeah, from memory Arch Linux and Illumnos are having difficulty moving on from Go 1.16.

felipecrs:
> Yeah, from memory Arch Linux and Illumnos are having difficulty moving on from Go 1.16.

Understood. Thank you.

twpayne:
I've updated the code to only include the `exit` template function when chezmoi is built with Go 1.17 or later so this can be merged now.

felipecrs:
That's awesome, thanks a lot!

PR #1685, opened 2021-11-29
feat: Add .chezmoi.workingTree template variable

felipecrs:
Use case:

```console
> cat ~/.dotfiles/home/repos/symlink_dotfiles.tmpl
{{- .chezmoi.workingTree -}}

```


twpayne:
Thank you!

PR #1694, opened 2021-11-30
docs: Make comparison table more accessible

glibsm:
Since there is particular reason those specific emojis, pick
ones that are much easier to tell apart without reliance on
color. Red/orange/green can be quite difficult to tell apart
for some people.


glibsm:
@twpayne I wasn't really sure what else to use instead of `⁉️`, open to suggestions

twpayne:
Thank you very much for this - really appreciated.

> @twpayne I wasn't really sure what else to use instead of `⁉️`, open to suggestions

We could keep the existing 🟠 as it is visually distinct to ✅ and ❌ (circle, square, cross), but the current ⁉️ is also fine.

PR #1703, opened 2021-12-02
chore: Add GitHub Action to lock closed issues

twpayne:


twpayne:
Instead pushed directly to `master` as there does not seem to be a way to test this without actually running it.

twpayne:
Note that the objective of the PR is to lock issues that have been resolved after seven days of inactivity as discussion on closed issues is easily lost. The default message should ask people to open a new issue instead. Unresolved issues are left open (as they [should be](https://drewdevault.com/2021/10/26/stalebot.html)).

PR #1718, opened 2021-12-08
Fix delete-bucket command in HOWTO doc

wedaly:
The HOWTO document says to use "chezmoi delete-bucket" to clear
the state of the `run_once_` scripts, but when I try that I
get this error:

  chezmoi: unknown command "delete-bucket" for "chezmoi"

This commit updates the documentation to use
"chezmoi state delete-bucket" instead,
which I believe is the intended command.


twpayne:
Awesome, thank you :)

Issue #1723, opened 2021-12-09
Prevent `init --one-shot` from deleting source folder if not created by it


felipecrs:
## Is your feature request related to a problem? Please describe.

The exact same situation as https://github.com/twpayne/chezmoi/issues/1721. The difference is that, with `--one-shot`, chezmoi tries to delete the source folder after applying.

That's completely intended when using something like `chezmoi init felipecrs --one-shot`, where chezmoi would create the source folder, but I think that it should not be like that for when the source folder was created by something else than chezmoi.

## Describe the solution you'd like

chezmoi to not try to delete `source` when it wasn't created by it.

## Describe alternatives you've considered

```console
$ chezmoi init file:///tmp/dotfiles
```

But with the following caveats:

1. Chezmoi will copy these files to a temporary source folder, which is not optimal
2. Does not work unless I share the whole `.git` folder when building the container, which messes up with Docker's caching.

## Additional context

None.

felipecrs:
By the way, I would also suggest to mention the deletion of the cache and the source state in https://www.chezmoi.io/docs/reference/#--purge.

twpayne:
No, sorry. chezmoi is a dotfile manager for individuals and its a reasonable assumption that `chezmoi init --one-shot` both creates and removes the source directory.

> By the way, I would also suggest to mention the deletion of the cache and the source state in https://www.chezmoi.io/docs/reference/#--purge.

Done in fd690b402893edbbcad0a15f635c2895ce0f6eda.

felipecrs:
> its a reasonable assumption that chezmoi init --one-shot both creates and removes the source directory.

That's not correct. Chezmoi is not creating the source directory in my example. That's the reason why I think it should not delete it: because it's not creating it.

twpayne:
> That's not correct. Chezmoi is not creating the source directory in my example. That's the reason why I think it should not delete it: because it's not creating it.

I disagree here. Part of `chezmoi init`'s responsibilities is to create the source directory, either with a new git repository, or cloning it from an an existing repository. The only exception to this is that `chezmoi init` has also taken the role of re-generating the config file if the config file template has changed. With hindsight, regenerating the config file should have been a different command.

If the source directory already exists, then all of `chezmoi init`'s functionality can be implemented using existing commands. For example, I think the following commands are the equivalent of the desired behavior of `chezmoi init --one-shot` in this issue:

```bash
SOURCEDIR=/tmp/dotfiles
[...]

PR #1741, opened 2021-12-13
feat: Skip the purge of source if not created by chezmoi

felipecrs:
Closes https://github.com/twpayne/chezmoi/issues/1723

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -1491,7 +1491,7 @@ temporary environments (e.g. Docker containers).
 
 #### `--purge`
 
-Remove the source and config directories after applying.
+Remove the cache, config, and source directories after applying.
 
 #### `--purge-binary`
 
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -59,6 +59,7 @@ const (
 
 type purgeOptions struct {
 	binary bool
+	source bool
 }
 
 type templateConfig struct {
--- a/internal/cmd/initcmd.go
+++ b/internal/cmd/initcmd.go
@@ -115,8 +115,8 @@ func (c *Config) newInitCmd() *cobra.Command {
 	flags.IntVarP(&c.init.depth, "depth", "d", c.init.depth, "Create a shallow clone")
 	flags.VarP(c.init.exclude, "exclude", "x", "Exclude entry types")
 	flags.BoolVar(&c.init.oneShot, "one-shot", c.init.oneShot, "Run in one-shot mode")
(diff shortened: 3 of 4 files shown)


felipecrs:
@twpayne I created this PR just to illustrate what I had in mind. It's not yet ready because it at least is missing tests.

But I will totally understand if you do not agree, as you mentioned earlier in https://github.com/twpayne/chezmoi/issues/1723.

twpayne:
Thanks for your work here.

To be clear about the intention of the existing commands:
A. `chezmoi init` means "start using chezmoi for the first time by creating an empty source directory and local repo" or (sadly, overloaded, as already stated) "re-generate chezmoi's config file".
B. `chezmoi init --one-shot $REPO` means "install dotfiles from `$REPO`, and then remove all traces of chezmoi".
C. `chezmoi purge` means "remove all traces of chezmoi from the system".

As it stands, there are a few problems with this PR:
1. As can been seen from the failing tests, this breaks the `chezmoi purge` command, breaking intention C above.
2. It seems to prevent `chezmoi init --purge` from removing the source directory in the specific case of `chezmoi init` being called with no positional arguments, i.e intention A above, when `chezmoi init` calls `git init` to create a new git repo in the source directory if no such git repo already exists. This is different to the described intention in #1723 which only refers to the source directory, not the git repo within it. `chezmoi init --purge` also makes no sense: it means "initialize chezmoi for the first time, and then remove all traces of it".
3. I think the original intention of #1723 was to prevent `chezmoi init --one-shot $REPO` from deleting the source directory in the case that it already existed, but we see that this is contrary to intention B.

Happy to hear clarifications.

felipecrs:
1. This is not intended. I will fix it. Thanks for noting.
2. You are right. I see no clear way around this. But as you mentioned, there are no use-cases for `chezmoi init --purge`. Perhaps this compromise of "breaking" such usage can be taken?
3. No. It's not `chezmoi init --one-shot $REPO` but `chezmoi init --one-shot --source /my-repo`, which have no positional arguments. However, if I change to be effective on just `--one-shot` we will run into 2. again (but with `chezmoi init --one-shot` as opposed to `chezmoi init --purge`).

felipecrs:
Or, I can change the documentation to say that chezmoi will _attempt_ to delete the _sourceDir_. In my case, it will then fail because my file system is readonly, but chezmoi won't exit with an error because it was just an attempt.

twpayne:
Given that the desired functionality can be achieved using existing `chezmoi` commands I will close this.

PR #1761, opened 2021-12-22
refactor: use `[` instead of `test`

CodesOfRishi:
Just making 2 lines of code in `assets/scripts/install.sh` a little more readable, by using `[` instead of the `test` command.


twpayne:
You need to modify `internal/cmds/generate-install.sh/install.sh.tmpl` instead and run `go generate`. Please read the [contributing guide](https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md#generated-code).

PR #1765, opened 2021-12-22
Fix managed command example docs

avorima:
Running the examples using the short-hand `-i` flag listed in the output of `chezmoi managed --help` results in errors:

```
$ chezmoi managed -i d
chezmoi: invalid argument "d" for "-i, --include" flag: d: unknown entry type
$ chezmoi managed -i d,f
chezmoi: invalid argument "d,f" for "-i, --include" flag: d: unknown entry type
```

If this behaviour is actually a regression and thus a bug I can also look into how to fix it.


twpayne:
Thanks for catching this. Indeed, it looks like the short forms were never implemented.

PR #1768, opened 2021-12-23
Provide a way to get field value in pass entry.

xals:
Hello,

Following the guide in Browserpass "Organizing password store" (https://github.com/browserpass/browserpass-extension#organizing-password-store), we may have fields in an entry.

This PR allows to query this kind of field.

PS: this is my first time coding in Go, I checked and fixed with `make lint`, but my code may not be fully optimized.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -1129,6 +1129,24 @@ The first line of the output of `pass show <pass-name>` is available as the
 {{ pass "<pass-name>" }}
 ```
 
+If fields are passed to the function, the first one found in the `pass` entry
+will be returned. If no fields have been found, an error is raised.
+This is useful to get data from entries compatible with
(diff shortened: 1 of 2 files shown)


twpayne:
Thanks for this.

I prefer to solve this in a different way, as done in #1772, where I've added you as a co-author to the commit. Specifically:
* Rather than extending an existing template function with optional parameters, I've added a new template function to be consistent with the other password manager template functions.
* Rather than selecting the field in Go, the template function returns structured data so the field can be selected in the template.
* I've added tests.
* The documentation is added to the reference manual which documents all aspects of chezmoi, instead of the how-to.
* The commit messages follow chezmoi's [contributing guide](https://github.com/twpayne/chezmoi/blob/master/docs/CONTRIBUTING.md#contributing-changes).

Issue #1780, opened 2021-12-25
.chezmoiignore doesn't consider exclude when doing apply


glensc:
## Describe the bug

I have vim editor backups (`*~`) in `.chezmoiignore`:
```
➜ head -n1 .chezmoiignore
*~
```

but doing `chezmoi apply -v` still considers those files being somehow managed.

## To reproduce

```
# create pristine state
cd $(chezmoi source-path)
date > ~/.config/chez
chezmoi add ~/.config/chez
chezmoi apply -v
```

Reproducer 1

```
$ touch ~/.config/chez~ private_dot_config/chez~
$ chezmoi apply -v
diff --git a/.config/chez~ b/.config/chez~
deleted file mode 100644
index e69de29bb2d1d6434b8b29ae775ad8c2e48c5391..0000000000000000000000000000000000000000
--- a/.config/chez~
+++ /dev/null

$ ls -l ~/.config/chez~ private_dot_config/chez~
ls: cannot access '/Users/glen/.config/chez~': No such file or directory
-rw-r--r-- 1 glen staff 0 dets  25 11:09 private_dot_config/chez~
```

What is happening here?!?

Second reproducer:

```
$ date > ~/.config/chez~
$ chezmoi apply -v --dry-run
.config/chez~ has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]?
```

What the heck? `chez~` should not be managed!

## Expected behavior

`chezmoi` should behave like the `*~` doesn't exists anywhere.

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT   CHECK                MESSAGE
ok       version              v2.9.4, commit 7f0702503528fb708a5d9fc7b2a826decb777b26, built at 2021-12-19T21:10:21Z, built by Homebrew
ok       os-arch              darwin/amd64
ok       go-version           go1.17.5 (gc)
ok       executable           /usr/local/bin/chezmoi
ok       upgrade-method       replace-executable
ok       config-file          ~/.config/chezmoi/chezmoi.toml
ok       source-dir           ~/.local/share/chezmoi is a directory
ok       suspicious-entries   no suspicious entries
ok       working-tree         ~/.local/share/chezmoi is a directory
ok       dest-dir             ~ is a directory
ok       shell                found /usr/local/bin/bash
ok       edit-command         found /usr/local/bin/nvim
ok       umask                022
ok       git-command          found /usr/local/bin/git, version 2.34.1
[...]

PR #1781, opened 2021-12-25
Add collapsible details to bug report template

glensc:
It just looks better when the extra details are not visible by default:
- https://github.com/twpayne/chezmoi/issues/1780


twpayne:
Thanks!

PR #1783, opened 2021-12-26
onepassword: prompt to sign-in if the current session is missing or expired

kdomanski:
Currently when a user tries to run `chezmoi` against a repo which uses the 1Passoword template functions, it will work correctly with a session token already exported within the environment, but fail if e.g. the session is expired.

With this change, chezmoi will use the `op signin` function to verify the validity of current session and interactively prompt for master password if that fails. This prevents the user from having to repeatedly sign in and export their session manually.

Before:
```
$ chezmoi diff
chezmoi: template: xxx.sh.tmpl:5:30: executing "xxx.sh.tmpl" at <onepassword "xxx">: error calling onepassword: op get item xxx: exit status 1: [ERROR] 2021/12/26 01:28:33 You are not currently signed in. Please run `op signin --help` for instructions
```

After (in an interactive session):
```
$ chezmoi diff
Enter the password for user@example.com at my.1password.com:
```

After (without a TTY):
```
chezmoi: template: xxx.sh.tmpl:5:30: executing "xxx.sh.tmpl" at <onepassword "xxx">: error calling onepassword: op signin --raw: exit status 1: [ERROR] 2021/12/26 01:32:08 inappropriate ioctl for device
```

### DIFF
--- a/internal/cmd/onepasswordtemplatefuncs.go
+++ b/internal/cmd/onepasswordtemplatefuncs.go
@@ -13,6 +13,7 @@ import (
 
 type onepasswordConfig struct {
 	Command       string
+	Prompt        bool
 	outputCache   map[string][]byte
 	sessionTokens map[string]string
 }
@@ -129,7 +130,12 @@ func getOnepasswordArgs(baseArgs, args []string) []string {
 
 // refreshSession will return the current session token if the token within the environment is still valid.
 // Otherwise it will ask the user to sign in and get the new token.
+// If `sessioncheck` is disabled, it returns an empty string.
 func (c *Config) onepasswordGetOrRefreshSession(callerArgs []string) string {
+	if !c.Onepassword.Prompt {
+		return ""
+	}
+
 	var account string
 	if len(callerArgs) > 2 {
 		account = callerArgs[2]
--- a/internal/cmd/onepasswordtemplatefuncs.go
+++ b/internal/cmd/onepasswordtemplatefuncs.go
@@ -4,13 +4,17 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/twpayne/chezmoi/v2/internal/chezmoi"
 )
 
 type onepasswordConfig struct {
-	Command     string
-	outputCache map[string][]byte
+	Command       string
+	outputCache   map[string][]byte
+	sessionTokens map[string]string
 }
 
 type onePasswordItem struct {
@@ -23,8 +27,9 @@ type onePasswordItem struct {
 }
 
 func (c *Config) onepasswordItem(args ...string) *onePasswordItem {
+	sessionToken := c.onepasswordGetOrRefreshSession(args)
 	onepasswordArgs := getOnepasswordArgs([]string{"get", "item"}, args)
-	output := c.onepasswordOutput(onepasswordArgs)
+	output := c.onepasswordOutput(onepasswordArgs, sessionToken)
 	var onepasswordItem onePasswordItem
 	if err := json.Unmarshal(output, &onepasswordItem); err != nil {
 		returnTemplateError(fmt.Errorf("%s: %w\n%s", shellQuoteCommand(c.Onepassword.Command, onepasswordArgs), err, output))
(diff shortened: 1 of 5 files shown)


internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -58,19 +61,26 @@ func (c *Config) onepasswordItemFieldsTemplateFunc(args ...string) map[string]in
...
 
+	var secretArgs []string
+	if sessionToken != "" {
+		secretArgs = []string{"--session", sessionToken}
+	}
+
 	name := c.Onepassword.Command
-	cmd := exec.Command(name, args...)
+	// appending the session token here, so we don't log it by accident
+	cmd := exec.Command(name, append(secretArgs, args...)...)
```
twpayne, line 83: Instead of passing the session token on the command line (which also makes it visible to other users on the same machine via `ps`) is it possible to pass the session token in an environment variable instead?

internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -110,5 +121,79 @@ func getOnepasswordArgs(baseArgs, args []string) []string {
...
+		return token
+	}
+
+	var args []string
+	if account == "" {
+		// If no account has been given then look for any session tokens you can find in the environment
+		token = onepasswordInferSessionToken()
+		args = []string{"signin", "--raw"}
+	} else {
+		token = os.Getenv(fmt.Sprintf(" OP_SESSION_%s", account))
```
twpayne, line 148: nit: It looks like there's an unwanted extra space in the first argument to `fmt.Sprintf` here.

nit: As this is just string concatenation, we can use `"OP_SESSION_" + account` instead of `fmt.Sprintf`.

internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -110,5 +121,79 @@ func getOnepasswordArgs(baseArgs, args []string) []string {
...
+		args = []string{"signin", account, "--raw"}
+	}
+
+	// Let's not even specify an empty session string if we din't find anything
+	var secretArgs []string
+	if token != "" {
+		secretArgs = []string{"--session", token}
```
twpayne, line 155: As above, is it possible to use an environment variable instead of a command line argument here?

internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -110,5 +121,79 @@ func getOnepasswordArgs(baseArgs, args []string) []string {
...
+
+	return token
+}
+
+// onepasswordInferSessionToken will look for any session tokens in the environment
+// and if it finds exactly one then it will return it.
+func onepasswordInferSessionToken() string {
+	var token string
+	for _, env := range os.Environ() {
+		slice := strings.Split(env, "=")
```
twpayne, line 186: nit: Use `chezmoi.CutString` instead of `strings.Split` here. `chezmoi.CutString` is a backport of the `strings.Cut` function coming in Go 1.18.

internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -110,5 +121,79 @@ func getOnepasswordArgs(baseArgs, args []string) []string {
...
+	// maybe there's already a valid token cached in this run for this account
+	token, ok := c.Onepassword.sessionTokens[account]
+	if ok {
+		return token
+	}
+
+	var args []string
+	if account == "" {
+		// If no account has been given then look for any session tokens you can find in the environment
+		token = onepasswordInferSessionToken()
```
twpayne, line 145: Instead of extracting a single session token, we iterate over all environment variables and populate `c.onepasswordConfig.sessionTokens` whenever we find an environment variable beginning with `OP_SESSION_`. This should simplify some of the overall logic.

twpayne (commented):
Thank you - this is fantastic!

I've left a few comments that I think will help improve security and simplify the code.

kdomanski:
Hey, you're very welcome. Thanks for creating this great software.

I created a feature request on the 1password forums that should address some of the shortcomings of this PR: https://1password.community/discussion/126053/reuse-a-session-token-from-environment-not-from-command-line/

internal/cmd/onepasswordtemplatefuncs.go
kdomanski: Not that I know of, unfortunately. I hope [1Password implements that feature](https://1password.community/discussion/126053/reuse-a-session-token-from-environment-not-from-command-line/).

What we could do in the meantime, is add a config option to chezmoi that will toggle this whole thing.
Two open questions I have about that are:
- Where and how would such a setting be configured?
- Should it default to enabled or disabled?

internal/cmd/onepasswordtemplatefuncs.go
kdomanski: Fixed.

internal/cmd/onepasswordtemplatefuncs.go
kdomanski: See answer above.

internal/cmd/onepasswordtemplatefuncs.go
kdomanski: That's neat, I didn't know that was coming in 1.18, thanks. Fixed.

internal/cmd/onepasswordtemplatefuncs.go
kdomanski: If you look closely, `c.onepasswordConfig.sessionTokens` only contains tokens that have been either verified to be valid, or freshly acquired, so we cannot pre-populate it without forcing refresh on every token in the environment, regardless of whether it used or not.

internal/cmd/onepasswordtemplatefuncs.go
twpayne: Good point. I agree that putting this behavior behind a switch makes sense: the key breaking change it introduces is that it changes chezmoi from exiting with an error when no 1Password token is available to prompting the user for interactive input, which will break scripts written with no interaction in mind. However, for almost all users, I think it is a clear improvement in behavior so it should be enabled by default.

I would add a `onepassword.prompt` boolean [configuration variable](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md#variables) defaulting to `true`.

Please could you also update the [documentation for the `onepassword` template function](https://github.com/twpayne/chezmoi/blob/master/docs/REFERENCE.md#onepassword-uuid-vault-uuid-account-name) to describe the behavior.

Thank you!

internal/cmd/onepasswordtemplatefuncs.go
twpayne: Thanks for the explanation.

internal/cmd/onepasswordtemplatefuncs.go
twpayne: By the way, I'm also happy to accept this PR as-is and implement the above myself if you prefer.

twpayne:
Thank you very much :)

PR #1787, opened 2021-12-26
yadm comparison

TheLocehiliosan:
I recently saw this comparison page referenced, and noticed some inaccuracies (at least regarding yadm).

Cheers


twpayne:
Thanks for the corrections :)

Issue #894, opened 2020-09-28
Symmetric encryption doesn't seem to work


flowchartsman:
## To reproduce

My chezmoi.toml is only the following:
```
[gpg]
    symmetric = true
```

```
$ chezmoi add  --encrypt ~/.local/share/system_bootstrap/Brewfile
You did not specify a user ID. (you may use "-r")

Current recipients:

Enter the user ID.  End with an empty line:
```

## Expected behavior

Prompted for password, file or directory is encrypted and moved into `~/.local/share/chezmoi`

## Output of command with `--verbose`

```
$ chezmoi add --verbose --encrypt ~/.local/share/system_bootstrap/Brewfile
You did not specify a user ID. (you may use "-r")

Current recipients:

Enter the user ID.  End with an empty line
```

## Output of `chezmoi doctor`

```
$ chezmoi doctor
warning: version dev
     ok: runtime.GOOS darwin, runtime.GOARCH amd64
     ok: /Users/<me>/.local/share/chezmoi (source directory, perm 700)
     ok: /Users/<me> (destination directory, perm 755)
warning: /Users/<me>/.config/chezmoi/chezmoi.toml (configuration file)
     ok: /bin/bash (shell)
     ok: /usr/local/bin/nvim (editor)
     ok: /usr/local/bin/vimdiff (merge command)
     ok: /usr/local/bin/git (source VCS command, version 2.28.0)
     ok: /usr/local/bin/gpg (GnuPG, version 2.2.23)
warning: op (1Password CLI, not found)
warning: bw (Bitwarden CLI, not found)
warning: gopass (gopass CLI, not found)
warning: keepassxc-cli (KeePassXC CLI, not found)
warning: lpass (LastPass CLI, not found)
warning: pass (pass CLI, not found)
warning: vault (Vault CLI, not found)
```

twpayne:
Ah, chezmoi's documentation is severely lacking here. It's been a while since I wrote the code, but chezmoi assumes that you already have gpg configured and doesn't help you do this. In short, you need to configure gpg so that `gpg --symmetric` works on its own, outside chezmoi, then you should be able to use the config that you posted.

Apologies for this, I'll keep this issue open until the documentation is improved.

flowchartsman:
[...]

PR #1826, opened 2022-01-15
feat: Switch to Material for MkDocs for website

### LINKED ISSUE #696: Add a man page

I guess usability. I love to have access to an exhaustive documentation from the terminal, and don't rely on any window manager. Also, this is extremely valuated in the *nix world.

I personally suggest mdoc (see `groff_mdoc(7)`) macros, because they offer a full DSL. They are also very widespread and portable (ex: `sudo(8)`).

## Additional context

The online documentation of `chezmoi` is of very good didactic quality, and I understand the burden of maintaining two "sources of truth". I would be happy to help.

### LINKED ISSUE #1577: [Website] Variables table on website triggers horizontal scroll
## Describe the bug

When you visit https://www.chezmoi.io/docs/reference/#variables, the table is wider than the content area width so it scrolls horizontally. This experience could be improved, particularly since the table is tall and therefore hides the scrollbar at the bottom. Ideally, the description column on the right would also be wider.

image

## To reproduce

Visit https://www.chezmoi.io/docs/reference/#variables. I used Firefox with a fullscreen window on a 1080p monitor. The problem can also be reproduced in Chrome.

## Expected behavior

I think there are a variety of ways to improve the experience. Here are some possibilities:
* Make the content area max width of the site wider. (It doesn't make sense to have whitespace on both sides with a horizontal scroll in between.)
* Make the font size in the table smaller.
* Reduce the padding in the table.
* Remove the first column, `Section`, from the table, and perhaps change that to a heading or combine with the second column, `Variable`.
* Eliminate the table completely. For example, format similarly to [commands](https://www.chezmoi.io/docs/reference/#commands) with an h3 heading per variable.

If any of these ideas sound appealing, I could take a stab at one or more of the above ideas in a PR.

## Additional Context

It's worth noting that there's a similar issue in the table on the [Comparison](https://www.chezmoi.io/docs/comparison/) page, and some solutions above might solve the issue in both places while others wouldn't. In any case, I don't think the issue on the comparison page is quite as problematic since the table's shorter there and it's also less important to see all columns at once, but it might still be worth thinking about in a broader context or coming up with a separate solution for that page.

twpayne:
This PR switches from Hugo to [Material for MkDocs](https://squidfunk.github.io/mkdocs-material) for the documentation.

The main loss is that documentation links within https://github.com/twpayne/chezmoi will stop working, but the main gain is much better and more beautiful structured documentation.

Fixes #1577.
Closes #696 and #894 (sorry!).


daurnimator:
I guess this means there is no longer local docs worth shipping along with the chezmoi package?

PR #1828, opened 2022-01-17
WIP support 1Password CLI 2.0

kdomanski:
With the changed API structure IMO I'd reconsider whether having both `onepasswordDetailsFields` and `onepasswordItemFields` with different behavior makes sense. As far as I can tell, all fields are now returned in one array, but some have a Section field and some don't.

### DIFF
--- a/internal/cmd/onepasswordtemplatefuncs.go
+++ b/internal/cmd/onepasswordtemplatefuncs.go
@@ -6,19 +6,25 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
+
+	"github.com/coreos/go-semver/semver"
 
 	"github.com/twpayne/chezmoi/v2/internal/chezmoi"
 )
 
+const onePasswordVersionRegexp string = `^(\d+\.\d+\.\d+(?:\-beta\.\d+)?)` //nolint:gosec
+
 type onepasswordConfig struct {
 	Command       string
 	Prompt        bool
+	majorVersion  int64
 	outputCache   map[string][]byte
 	sessionTokens map[string]string
 }
 
-type onePasswordItem struct {
+type onePassword1Item struct {
 	Details struct {
 		Fields   []map[string]interface{} `json:"fields"`
 		Sections []struct {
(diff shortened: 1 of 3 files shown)


internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -101,8 +210,20 @@ func (c *Config) onepasswordOutput(args []string, sessionToken string) []byte {
 }
 
 func (c *Config) onepasswordTemplateFunc(args ...string) map[string]interface{} {
+	var cmdArgs []string
+	switch c.onepasswordMajorVersion() {
+	case 1:
+		cmdArgs = []string{"get", "item"}
+		break //nolint:gosimple
```
twpayne, line 217: nit: `break` is not needed here. Go, unlike C, breaks by default.

internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -6,19 +6,25 @@ import (
...
 	"os/exec"
+	"regexp"
 	"strings"
 
+	"github.com/coreos/go-semver/semver"
+
 	"github.com/twpayne/chezmoi/v2/internal/chezmoi"
 )
 
+const onePasswordVersionRegexp string = `^(\d+\.\d+\.\d+(?:\-beta\.\d+)?)` //nolint:gosec
```
twpayne, line 17: nit: please use

```go
var onePasswordVersionRx = regexp.MustCompile(`...`)
```

This will catch invalid regexps at startup.

twpayne (commented):
Thanks, this looks very promising.

twpayne:
Presumably this a draft PR as 1Password haven't released 1Password CLI 2.0 yet, is that correct?

kdomanski:
Yeah, it's in closed beta and they're still changing some behaviors. [The changelog is public though.](https://app-updates.agilebits.com/product_history/CLI2)

twpayne:
Thanks! This is a great contribution and thank you for being ahead of the curve here :)

twpayne:
This is also an opportunity for chezmoi to provide a cleaner set of template functions for 1Password users, so as well as adding compatibility with CLI 2.0 to existing functions, we can also add new template functions when it makes sense.

kdomanski:
Huh, I just realized that this public page also contains the actual download links. ;-)

kdomanski:
The new [op://](https://developer.1password.com/docs/cli/reference/commands/read) syntax looks super promising as well. The downside right now is that getting multiple fields from one item would be slower, because of the fields being cached separately.

Do you have any plans for the near future to refactor state computation and make it concurrent? I played around with it the other day and a naive POC worked well (and fast!) but it probably wasn't concurrency-safe.

twpayne:
> Do you have any plans for the near future to refactor state computation and make it concurrent? I played around with it the other day and a naive POC worked well (and fast!) but it probably wasn't concurrency-safe.

I've been thinking about this for some time, and have done some initial experiments. There are several independent parts of chezmoi that could be made concurrent. The main complexity, as you identify, is doing this in a safe way (there is a lot of shared mutable state). The main reason I haven't done it yet is that with concurrency you lose determinism. Currently chezmoi performs all operations in a deterministic order, which makes things very repeatable.

PR #1829, opened 2022-01-18
docs: Fix small typo in User guide

rocode:
Fix small typo in the docs example under User Guide -> Manage machine-to-machine differences where the code snippet did not match the filename in the example.


twpayne:
Thank you :)

PR #1841, opened 2022-01-21
Update gh-pages: fix broken link to user manual

Qeole:
The link to the user manual at the end of the “Quick start” page is broken. This is a fix, making it point to the user guide.


twpayne:
Thanks! Note that the website is generated as [described in the contributor's guide](https://www.chezmoi.io/developer/website/) so the fix has to be applied to the source in the `master` branch, not the `gh-pages` branch. I've done this in fd5bb6a26e95d53efe9ef4006f49c970159db210, keeping you as the author.

PR #1851, opened 2022-01-25
feat: Improve performance by using concurrency

twpayne:


twpayne:
cc @kdomanski this PR makes chezmoi read the source state concurrently. In my simple benchmarks this does not improve performance, but in other tests I've done I do think there's a significant performance improvement to come from handling externals, computing the target state, and updating the destination state concurrently.

PR #1861, opened 2022-01-30
feat: Support 1Password CLI 2.0

twpayne:
Fixes #1860.
Closes #1828.
cc @cadenkriese

This builds on @kdomanski's work in #1828 (commit is co-authored by @kdomanski) to take into account the refactoring of template functions in #1859 and further refactoring.

At this instant, I believe the code is correct, but there's a bug in the tests that get confused by existing `OP_SESSION_` environment variables, and I'm most of the way through fixing it.

This, [raised by @kdomanski](https://github.com/twpayne/chezmoi/pull/1828#issue-1106185987), still needs to be addressed:

> With the changed API structure IMO I'd reconsider whether having both `onepasswordDetailsFields` and `onepasswordItemFields` with different behavior makes sense. As far as I can tell, all fields are now returned in one array, but some have a Section field and some don't.

Note that as the structure of the data returned by 1Password CLI 2.0 is different to 1Password CLI 1.x, users will have to update their templates when they switch to version 2 anyhow.

PR #1862, opened 2022-01-31
chore: Factor out common exec errors

twpayne:
This needs to be applied to 1Password too, but #1861 should probably land first.


twpayne:
#1861 has been rebased to include these changes.

PR #1885, opened 2022-02-13
Add support for xz archives

### LINKED ISSUE #1883: Support .xz for external archives

I was planing to add some external package (from GitHub), which was archived using XZ compression. At that point I realized there is no support for it.

It'd be nice to have support for `.xz` archives as well.

Another possibility for this to be solved (on Debian at least) is to install `xz-utils` package (if one is not already present).
While this does alternative removes problem, my plan was to have "slim" setup for my `dotfiles`, which does not update or adds additional packages to system (i.e. only directly executable binaries).

## Additional context

Also, I have already implemented this change [here](https://github.com/twpayne/chezmoi/compare/master...zbiljic:feature/archive-format-xz?expand=1), however I figured it would be prudent to ask before creating PR for this.

Note that there is separate commit where `go mod tidy` is performed, because external dependency was added in order to support this change. If that was done on the `master` branch, it can be removed from changeset.

zbiljic:
This change adds support for `XZ` archives.

Notes:
* adds new dependency `github.com/ulikunitz/xz` (Pure golang package for reading and writing xz-compressed files)
* new archive format comes before `zip` because I sorted list after adding this new one
* only two small documentation changes (could not find any other notable place)
* one new test was added; uses `tar` and requires `xz-utils` (or comparable package); run with `go test -v -run TestScript/importxz ./...`
* the `make test-release` run was done in `golang:1.17.7-bullseye` container

closes #1883

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/import.md
+++ b/assets/chezmoi.io/docs/reference/commands/import.md
@@ -6,7 +6,7 @@ exactly match the contents of a downloaded archive. You will generally always
 want to set the `--destination`, `--exact`, and `--remove-destination` flags.
 
 The supported archive formats are `tar`, `tar.gz`, `tgz`, `tar.bz2`, `tbz2`,
-and `zip`.
+`xz`, and `zip`.
 
 ## `--destination` *directory*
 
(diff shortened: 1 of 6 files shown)


twpayne:
Thanks very much for this and for being chezmoi's 100th contributor!

There are a couple of minor fixes that I'll open a PR for.

zbiljic:
That is great.

I saw your additional changes. Sorry I missed some things.

PR #1914, opened 2022-02-22
Allow overriding GitHub user in shell installer

kyounger:
Not set in stone, but was hoping to start the conversation with a little code. I would prefer to maintain my own fork of chezmoi and this would still allow me to use a single-line installer (obviously with my own url).

A couple things I noticed:
1. I was not sure how to accomplish this same edit in the PowerShell variant.
2. Similar to the BINDIR, maybe you want the GITHUB_USER to additionally be set via switch?

Not a life-or-death change, but does abstract the code a bit from your namespace.

### DIFF
--- a/assets/scripts/install.sh
+++ b/assets/scripts/install.sh
@@ -8,11 +8,12 @@
 set -e
 
 BINDIR=${BINDIR:-./bin}
+GITHUB_USER=${GITHUB_USER:-twpayne}
 TAGARG=latest
 LOG_LEVEL=2
 EXECARGS=
 
-GITHUB_DOWNLOAD=https://github.com/twpayne/chezmoi/releases/download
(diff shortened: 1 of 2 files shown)


twpayne:
Thanks for this, but you need to modify the install script template (`internal/cmds/generate-install.sh/install.sh.tmpl`), as [described in the contributing guide](https://www.chezmoi.io/developer/install-script/).

kyounger:
Apologies for missing that. I reviewed that full contributor guide. A few tests failed for me locally, but I think that was an environment setup situation.

twpayne:
Thanks! The test failures are because one of the tests uses the GitHub API and this easily hits rate limits if you don't have a GitHub personal access token set in an appropriate environment variable. I need to make the test less fragile.

twpayne:
Thanks for the contribution. I tided up the commits and edited the commit message to match chezmoi's conventions after merging. As chezmoi's release notes are generated from the commit messages, their format is important.

PR #1919, opened 2022-02-23
feat: Add support for GNOME Keyring on FreeBSD

### LINKED ISSUE #1918: Update zalando/go-keyring when FreeBSD support gets merged

I'm using `chezmoi` on FreeBSD, but I couldn't use the integration with `zalando/go-keyring` since FreeBSD isn't supported. I opened an issue [there](https://github.com/zalando/go-keyring/issues/67), and *BSD support should be added soon

When upstream `zalando-keyring` gets released, please test and consider requiring the new version here (so *BSD users can enjoy keyring support with chezmoi)

Crying over my lack of proficiency to implement this myself

## Additional context

NA

twpayne:
Fixes #1918.

@torculus if you [build chezmoi from source](https://www.chezmoi.io/install/#install-from-source) from this PR you should get GNOME Keyring support on FreeBSD. I'll wait for the official tag of zalando/go-keyring before merging this.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/secret.md
+++ b/assets/chezmoi.io/docs/reference/commands/secret.md
@@ -24,3 +24,10 @@ manager. Normally you would use chezmoi's existing template functions to retriev
     $ chezmoi secret keyring get --service=service --user=user
     $ chezmoi secret keyring delete --service=service --user=user
     ```
+
+!!! warning
+
+    On FreeBSD, the `secret keyring` command is only available if chezmoi was
+    compiled with cgo enabled. The official release binaries of chezmoi are
+    **not** compiled with cgo enabled, and `secret keyring` command is not
+    available.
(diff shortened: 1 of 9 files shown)


szuecs:
@twpayne https://github.com/zalando/go-keyring/releases/tag/v0.2.0

twpayne:
Thanks @szuecs - I think this PR is good to go now.

Important note: using GNOME Keyring brings in a dependency on github.com/godbus/dbus, which requires cgo. cgo is not easily available when cross-compiling, so the "official" builds of chezmoi for FreeBSD (which are built on Ubuntu Linux) will not have cgo enabled when building the FreeBSD binaries. However, if you either build chezmoi from source on FreeBSD or install it from ports then you should be OK.

cc @yzgyyang who I think has been updating chezmoi in FreeBSD's ports.

PR #1921, opened 2022-02-24
Correctly cut hostname off at . character

### LINKED ISSUE #1920: .chezmoi.hostname no longer strips text after the first .
I'm using chezmoi built by the nix package manager on OSX. I updated chezmoi a few days ago to 2.12.0. 

The `.chezmoi.hostname` variable is documented as:
```
The hostname of the machine chezmoi is running on, up to the first .
```
It used to return `swt30-mbp` for me. 

After the update this variable now returns `swt30-mbp.local`, so my templates using this variable have broken. 

```shell
$ chezmoi execute-template "{{ .chezmoi.hostname }}"
swt30-mbp.local
```

This seems like it's probably a bug? Though as a workaround, I can always include the suffix or switch to another test to distinguish between machines.

Also, I see there's a `fqdnHostname`, but this is blank:
```shell
$ chezmoi execute-template "{{ .chezmoi.fqdnHostname }}"
  
```

I hadn't updated chezmoi for a while so was probably a few versions behind - so I'm sorry I can't be more specific about the last working version.

swt30:
The `hostname` variable was not being cut off at the first `.` character,
as intended. This is because a previous revision of the code accidentally
changed the separator character to `=` instead of `.`

This change returns the behaviour of the `hostname` variable to match
its documentation.

Opening this as a draft while I work out how to add a test for this behaviour
(not a Go developer sorry, so need to find an appropriate existing test
to use as a template). I've checked that the existing tests pass. 

Fixes #1920

### DIFF
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -920,7 +920,7 @@ func (c *Config) defaultTemplateData() map[string]interface{} {
 
 	var hostname string
 	if rawHostname, err := os.Hostname(); err == nil {
-		hostname, _, _ = chezmoi.CutString(rawHostname, "=")
+		hostname, _, _ = chezmoi.CutString(rawHostname, ".")
 	} else {
 		c.logger.Info().
 			Err(err).


swt30:
Just updated the commit message hence the force-push there.

OK, in that case I'll not worry about a test!

PR #1925, opened 2022-02-26
fix: Allow user and repo to set in install script

### LINKED ISSUE #1924: Installer script / codespaces env variable conflict
## Describe the bug

The script returned from `https://chezmoi.io/get` uses the `GITHUB_USER` environment variable to determine which GitHub repo to request a chezmoi release from. But `GITHUB_USER` is also in [the set of standard environment variables in a codespace](https://docs.github.com/en/codespaces/developing-in-codespaces/default-environment-variables-for-your-codespace), so the command silently fails.

## To reproduce

1. Follow the instructions described [here](https://www.chezmoi.io/user-guide/machines/containers-and-vms/) to create an install script
2. Create a codespace
3. The dotfiles setup fails to install chezmoi
4. There is no chezmoi to initialize/apply

## Expected behavior

1. The dotfiles setup should successfully install chezmoi
2. Chezmoi should successfully initialize/apply

n/a — chezmoi was never installed

## Output of `chezmoi doctor`

n/a — chezmoi was never installed

## Additional context

My workaround was to add `GITHUB_USER=""` near the top of my install.sh. If that workaround is the best path, maybe this issue can be closed with a docs update?

twpayne:
Fixes #1924 by no longer using the $GITHUB_USER environment variable.

cc @kyounger as the person likely to be affected by this change.

### DIFF
--- a/assets/scripts/install.sh
+++ b/assets/scripts/install.sh
@@ -8,12 +8,12 @@
 set -e
 
 BINDIR=${BINDIR:-./bin}
-GITHUB_USER=${GITHUB_USER:-twpayne}
+CHEZMOI_USER_REPO=${CHEZMOI_USER_REPO:-twpayne/chezmoi}
 TAGARG=latest
 LOG_LEVEL=2
 EXECARGS=
 
-GITHUB_DOWNLOAD=https://github.com/${GITHUB_USER}/chezmoi/releases/download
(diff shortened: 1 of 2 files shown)


kyounger:
🤦‍♂️

(thanks for the CC 😄)

PR #1950, opened 2022-03-13
[RFC] feat: Consider file order in chezmoi config arguments

Mephistophiles:
This commit overrides `reverse` (false -> true) if `{{ .Target }}` stands first in `diff.args`.

### DIFF
--- a/pkg/cmd/diffcmd.go
+++ b/pkg/cmd/diffcmd.go
@@ -47,6 +47,36 @@ func (c *Config) newDiffCmd() *cobra.Command {
 	return diffCmd
 }
 
+// overrideReverseFromArgs overrides reverse arg in config.
+func (c *Config) overrideReverseFromArgs() bool {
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks for this. I think I understand what it does, but could you explain why it's needed?

Mephistophiles:
this allow to enable pernanent "reverse" mode for the diff command.  although I can just put the reverse flag in the config ...

twpayne:
> although I can just put the reverse flag in the config ...

Indeed. You can also reverse the args directly in `diff.args`. So, I think the functionality in this PR is already available in chezmoi, so I do not think this PR is needed. What do you think?

twpayne:
I've created #1951 to add the `diff.reverse` configuration variable.

Mephistophiles:
It is good enough for me! Thanks!
This PR is not longer need.

PR #1961, opened 2022-03-21
Fix `op signin` for v2

halostatue:
The release of op 2.0.0 introduced a breaking change to `op signin` so that the previous version invocation `op signin account` no longer works. Now, it is required to use the option `--account`, as in `op signin --account account`.

Note that there is likely another issue hidden by this when not using biometric authentication for `op`, because the session environment variable is no longer `OP_SESSION_<account>` but `OP_SESSION_<accountUUID>`. If the account is specified as the account UUID, everything will probably work. I’m using biometric authentication, so I cannot easily verify this.

Also note that in `op` v1, you could specify `--vault ''`; this is no longer legal.

https://1password.community/discussion/128083/problems-with-op-v2-when-integrating-with-chezmoi#latest


twpayne:
Thank you for such a high quality pull request - including tests and documentation, thank you!

> I think that the better option would be to modify the logic in [`newOnepasswordArgs`](https://github.com/twpayne/chezmoi/blob/master/pkg/cmd/onepasswordtemplatefuncs.go#L381-L400) so that it’s something more like
> ...
> This would prevent adding `--vault ''` or `--account ''` if the values are empty. I can make that change and revert some of the tests if you’d like.

This indeed does sound like the better option. Please do implement this.

twpayne:
For Windows, I'll happily still accept this PR even if you don't get the tests passing on Windows. Windows is quite a peculiar beast, and it can be extremely tiresome debugging through GitHub Action failure round-trips. I've got a local Windows machine for chezmoi development and will happily implement the Windows fixes if you don't have access to a Windows box.

halostatue:
@twpayne You’re going to have more documentation to review, because I decided to really _finalize_ the the documentation for 1Password CLI v2 as well as implementing the changes discussed above.

fix: Finalize 1Password CLI 2.x support

- Ensure that `--vault ''` and `--account ''` will not be passed when given empty strings. This did not cause problems for 1Password 1.x, but errors out on 1Password 2.x.
- Finalize documentation, emphasizing 1Password 2.x formatting first.
- Changed some announcement box types to reflect the changes that are caused by using 1Password 2.x.

Hopefully, the tests will entirely pass now.

twpayne:
Thanks very much for this, I made a couple of minor tweaks in #1966.

PR #1966, opened 2022-03-24
chore: 1Password fixes

twpayne:
Refs #1961.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/password-managers/1password.md
+++ b/assets/chezmoi.io/docs/user-guide/password-managers/1password.md
@@ -6,22 +6,24 @@ expose data as a template function.
 
 !!! note
 
-    [1Password CLI 2.0](https://developer.1password.com/) has been released.
+    The[1Password CLI 2.0](https://developer.1password.com/) has been released.
     Examples will be shown using the changed details for this version and
     examples for 1Password CLI 1.x will follow.
 
 Log in and get a session using:
 
 ```console
-# For 1Password 2.x. Neither step is necessary with biometric authentication.
 $ op account add --address $SUBDOMAIN.1password.com --email $EMAIL
 $ eval $(op signin --account $SUBDOMAIN)
 ```
 
-??? info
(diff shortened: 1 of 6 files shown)


assets/chezmoi.io/docs/user-guide/password-managers/1password.md
```diff
@@ -6,22 +6,24 @@ expose data as a template function.
 
 !!! note
 
-    [1Password CLI 2.0](https://developer.1password.com/) has been released.
+    The[1Password CLI 2.0](https://developer.1password.com/) has been released.
```
halostatue, line 9: There’s a space missing after The, here.

pkg/cmd/onepasswordtemplatefuncs.go
```diff
@@ -248,12 +248,14 @@ func (c *Config) onepasswordGetOrRefreshSessionToken(args *onepasswordArgs) (str
 		sessionToken = os.Getenv("OP_SESSION_" + args.account)
 
 		switch {
-		case c.Onepassword.version.Major < 2:
+		case c.Onepassword.version.Major == 1:
 			commandArgs = []string{"signin", args.account, "--raw"}
-		case c.Onepassword.version.Major < 3:
+		case c.Onepassword.version.Major >= 2:
```
halostatue, line 253: I had intentionally chosen _not_ to do `>= 2` because it is possible that Agilebits will change the protocol again for v3. We’re several years _away_ from it since v2 was just released, but it felt safer to put upper boundaries on compatibility.

In fact, it _might_ be worth doing the `panic(&unsupportedVersionError{…})` during version detection so that it errors out earlier.

assets/chezmoi.io/docs/user-guide/password-managers/1password.md
twpayne: Thanks! Fixed with f86b5b8551c7688a089f1b801dee27ac30685c35.

PR #1990, opened 2022-04-03
fix: Fix use of drives as home directories on Windows

### LINKED ISSUE #1964: Network drive as destDir not working
My company uses a network drive as a users home folder for Windows 10.  I added this to my `chezmoi.toml` file.
`destDir = "H:"`
I can't add files to chezmoi, this is what it looks like

```
PS H:\> chezmoi add .\.gitconfig
chezmoi: H:/.gitconfig: not in H:/
```
I have super double checked and .gitconfig is in my H drive folder.

```
PS H:\> ls .\.gitconfig
    Directory: H:\
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----          3/9/2022   9:21 AM            856 .gitconfig
```

I have tried dos and powershell.  It does work if I use my user home folder on my C drive without `desDir` set in my config file.
I have also tried different ways of listing my H drive
```
destDir = "H:/"
destDir = "H:\"
destDir = "H:\\"
destDir = "H"
```

I've googled the issue and I can't find any solutions.  I have also read through the user guide and reference guide.

* [x] I have read [chezmoi's user guide](https://chezmoi.io/user-guide/), and not found the answer.
* [x] I have searched [chezmoi's reference guide](https://chezmoi.io/reference/), and not found the answer.
* [x] Other, please give details.

## Output of any commands you've tried with `--verbose` flag

```console
PS H:\> chezmoi --verbose add .\.gitconfig
chezmoi: H:/.gitconfig: not in H:/
```

## Output of `chezmoi doctor`

<details>

```console
PS H:\> chezmoi doctor
ok        version              v2.14.0, commit 7a2edbef28cc946972799ec287f78cb3b867ead4, built at 2022-03-09T21:44:20Z, built by goreleaser
ok        latest-version       v2.14.0
ok        systeminfo           Microsoft Windows 10 Enterprise (10.0.19042 N/A Build 19042)
ok        go-version           go1.17.8 (gc)
ok        executable           C:/users/m315724/.local/bin/chezmoi.exe
ok        upgrade-method       replace-executable
ok        config-file          ~/.config/chezmoi/chezmoi.toml
ok        source-dir           ~/.local/share/chezmoi is a directory
ok        suspicious-entries   no suspicious entries
ok        working-tree         ~/.local/share/chezmoi is a directory
ok        dest-dir             H:/ is a directory
ok        shell-command        found C:/WINDOWS/system32/cmd.exe
ok        shell-args           'C:\\WINDOWS\\system32\\cmd.exe'
ok        cd-command           found C:/WINDOWS/system32/cmd.exe
ok        cd-args              'C:\\WINDOWS\\system32\\cmd.exe'
ok        edit-command         found ~/AppData/Local/Programs/Microsoft VS Code/bin/code.cmd
ok        edit-args            code --wait
ok        git-command          found ~/AppData/Local/Programs/Git/cmd/git.exe, version 2.35.1
warning   merge-command        vimdiff not found in $PATH
info      gpg-command          gpg not found in $PATH
info      keepassxc-command    keepassxc-cli not found in $PATH
```

</details>

twpayne:
Fixes #1964.

@zb140 would you mind reviewing this?

From [your comment](https://github.com/twpayne/chezmoi/issues/1964#issuecomment-1086414230):

> It may also be possible to fix it by also applying the "sometimes add a slash" logic in `TrimDirPrefix` but a) I haven't tried it, and b) I'm not as big a fan of that solution on a philosophical level because I think it's buggy behavior on Go's part to not include the slash in the first place. `H:` by itself is just a volume specifier, not a directory, so I don't think it's really a proper return value for `path.Dir()`. But reasonable minds can differ, and I'm happy to be told I'm wrong about that.

I've actually taken this approach in this PR. I agree with your assessment that it's buggy behavior from Go to not include the slash in the first place, but this is sadly something that is unlikely to be ever fixed as it would likely break too much existing code.

> The immediate source of the error message is this check: https://github.com/twpayne/chezmoi/blob/master/pkg/chezmoi/abspath.go#L150, which is adding an extra `/` to the end of the path. The `HasPrefix` check on line 153 then checks if `H:/` has a prefix of `H://`, which of course it doesn't 😄 I think the goal of the check is to make sure that paths that are not the root directory always end in a `/`.

The reason for adding the `/` is so that the `strings.HasPrefix` check can be used later in the function. Without a trailing `/`, `/some/dir` would be a considered a prefix of `/some/dirfoo`, which is obviously incorrect. The previous check was just for `/`, which Go returns as the directory of `/`. In the PR, I ensure that the prefix has the `/` suffix, which also fixes the case where Go returns `H:` as the directory of `H:/foo`.

Please take a look.

### DIFF
--- a/pkg/chezmoi/abspath.go
+++ b/pkg/chezmoi/abspath.go
@@ -53,7 +53,7 @@ func (p AbsPath) Bytes() []byte {
 
 // Dir returns p's directory.
 func (p AbsPath) Dir() AbsPath {
-	return NewAbsPath(path.Dir(p.absPath))
+	return NewAbsPath(filepath.Dir(p.absPath)).ToSlash()
 }
 
 // Empty returns if p is empty.
@@ -147,7 +147,7 @@ func (p AbsPath) TrimDirPrefix(dirPrefixAbsPath AbsPath) (RelPath, error) {
 		return EmptyRelPath, nil
 	}
 	dirAbsPath := dirPrefixAbsPath
-	if dirAbsPath.absPath != "/" {
+	if !strings.HasSuffix(dirAbsPath.absPath, "/") {
 		dirAbsPath.absPath += "/"
 	}
 	if !strings.HasPrefix(p.absPath, dirAbsPath.absPath) {
--- a/pkg/chezmoi/path_windows_test.go
+++ b/pkg/chezmoi/path_windows_test.go
@@ -1,10 +1,52 @@
 package chezmoi
 
 import (
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
+
+func TestAbsPathTrimDirPrefix(t *testing.T) {
+	for i, tc := range []struct {
+		absPath          AbsPath
+		dirPrefixAbsPath AbsPath
+		expected         RelPath
+	}{
+		{
+			absPath:          NewAbsPath("/home/user/.config"),
+			dirPrefixAbsPath: NewAbsPath("/home/user"),
+			expected:         NewRelPath(".config"),
+		},
+		{
+			absPath:          NewAbsPath("H:/.config"),
+			dirPrefixAbsPath: NewAbsPath("H:"),
+			expected:         NewRelPath(".config"),
+		},
+		{
+			absPath:          NewAbsPath("H:/.config"),
+			dirPrefixAbsPath: NewAbsPath("H:/"),
+			expected:         NewRelPath(".config"),
+		},
+		{
+			absPath:          NewAbsPath("H:/home/user/.config"),
+			dirPrefixAbsPath: NewAbsPath("H:/home/user"),
+			expected:         NewRelPath(".config"),
+		},
+		{
+			absPath:          NewAbsPath(`//server/user/.config`),
+			dirPrefixAbsPath: NewAbsPath(`//server/user`),
+			expected:         NewRelPath(".config"),
+		},
+	} {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			actual, err := tc.absPath.TrimDirPrefix(tc.dirPrefixAbsPath)
+			require.NoError(t, err)
+			assert.Equal(t, tc.expected, actual)
+		})
+	}
+}
 
 func TestNormalizeLinkname(t *testing.T) {
 	for _, tc := range []struct {


zb140 (changes requested):
Unfortunately this isn't quite enough to fix the problem. I initially wrote a whole long, rambling reply, adding more and more detail as I investigated. But by the time I got to the end, I discovered that there was a much simpler solution.  I've included the original in the collapsed section in case it's useful.

Adding a slash to `dirPrefixAbsPath` if there isn't one already does fix the initial error that was reported, but it will still panic (a little bit later) when you try to `chezmoi add` a file in the root of a drive, because `SourceState.AddDestAbsPathInfos` calls `MustTrimDirPrefix` with `p` set to `H:` and `dirPrefixAbsPath` set to `H:/`.  Those should actually be considered the same path, so you can patch `TrimDirPrefix` even further to handle that case and return `EmptyRelPath`, but even that isn't good enough -- the loop in `AddDestAbsPathInfos` will then call `path.Dir("H:")`, which (inexplicably) returns `.`.

However, it turns out that `filepath.Dir()` behaves sanely on Windows: `filepath.Dir("H:/foo")` returns `H:\`, as does `filepath.Dir("H:/")`.  So, the fix in this PR, plus changing `AbsPath.Dir` to do this:
```golang
// Dir returns p's directory.
func (p AbsPath) Dir() AbsPath {
	return NewAbsPath(filepath.Dir(p.absPath)).ToSlash()
}
```
is a complete solution to the bug. After making those two changes locally, I am able to use the root of a drive as my home directory and `chezmoi add` files without issue.  I'm not sure if there's a reason why `AbsPath.Dir` is using `path` instead of `filepath` here, though, so there may be more to it than this simple change.

<details>
<summary>A whole bunch of rambling</summary>
Apologies in advance for the "stream of consciousness" style of this comment; I kept making new discoveries while writing it up.  I tried to edit it to make it flow as well as I could, but it's still not great. Let me know if you want me to try to clarify anything :smile:

Unfortunately this isn't quite enough to fix the problem. It does fix the initial error, but if you try to `chezmoi add` a file in the root of a drive, it will panic:
```
goroutine 1 [running]:
github.com/twpayne/chezmoi/v2/pkg/chezmoi.AbsPath.MustTrimDirPrefix(...)
        C:/Users/zb/Projects/chezmoi/pkg/chezmoi/abspath.go:106
github.com/twpayne/chezmoi/v2/pkg/chezmoi.(*SourceState).AddDestAbsPathInfos(0xc00014f4a0, 0xc00079f438, {0x1a0f520, 0xc000589f60}, {{0xc00036b1f8, 0xc00079f848}}, {0x19fc918, 0xc0002eb650})
        C:/Users/zb/Projects/chezmoi/pkg/chezmoi/sourcestate.go:524 +0x226
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).destAbsPathInfos.func1({{0xc00036b1f8, 0xc0000974a0}}, {0x19fc918, 0xc0002eb650}, {0x0, 0x0})
        C:/Users/zb/Projects/chezmoi/pkg/cmd/config.go:1022 +0x1af
github.com/twpayne/chezmoi/v2/pkg/chezmoi.Walk.func1({0xc00036b1f8, 0x2}, {0x19fc918, 0xc0002eb650}, {0x0, 0x0})
        C:/Users/zb/Projects/chezmoi/pkg/chezmoi/system.go:151 +0x5d
github.com/twpayne/go-vfs/v4.walk({0x1b4bf847858, 0x2115620}, {0xc00036b1f8, 0x8}, 0xc00079f9f0, {0x19fc918, 0xc0002eb650}, {0x0, 0x0})
        C:/Users/zb/go/pkg/mod/github.com/twpayne/go-vfs/v4@v4.1.0/walk.go:34 +0xbb
github.com/twpayne/go-vfs/v4.Walk({0x1b4bf847858, 0x2115620}, {0xc00036b1f8, 0x8}, 0xc00036b1f8)
        C:/Users/zb/go/pkg/mod/github.com/twpayne/go-vfs/v4@v4.1.0/walk.go:66 +0x66
github.com/twpayne/chezmoi/v2/pkg/chezmoi.Walk({0x1a0f520, 0xc000589f60}, {{0xc00036b1f8, 0x1}}, 0x0)
        C:/Users/zb/Projects/chezmoi/pkg/chezmoi/system.go:153 +0x79
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).destAbsPathInfos(0xc0002f4a80, 0xc00014f4a0, {0xc00056d5c0, 0x1, 0x19eeb28}, {0x40, 0x9f, 0x58})
        C:/Users/zb/Projects/chezmoi/pkg/cmd/config.go:1024 +0x2f0
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).runAddCmd(0xc0002f4a80, 0xc000594be8, {0xc00056d5c0, 0xc000589fd0, 0xc000592cc0}, 0xc000573fc0)
        C:/Users/zb/Projects/chezmoi/pkg/cmd/addcmd.go:109 +0x85
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).makeRunEWithSourceState.func1(0xc0002ee000, {0xc00056d5c0, 0x1, 0x3})
        C:/Users/zb/Projects/chezmoi/pkg/cmd/config.go:1231 +0x97
github.com/spf13/cobra.(*Command).execute(0xc0002ee000, {0xc000573e40, 0x3, 0x4})
        C:/Users/zb/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x60e
github.com/spf13/cobra.(*Command).ExecuteC(0xc00048d900)
        C:/Users/zb/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3bc
github.com/spf13/cobra.(*Command).Execute(...)
        C:/Users/zb/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).execute(0xc00079fe00, {0xc00011e010, 0x4, 0x7})
        C:/Users/zb/Projects/chezmoi/pkg/cmd/config.go:1103 +0x78
github.com/twpayne/chezmoi/v2/pkg/cmd.runMain({{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}}, {0xc00011e010, 0x4, ...})
        C:/Users/zb/Projects/chezmoi/pkg/cmd/cmd.go:266 +0x151
github.com/twpayne/chezmoi/v2/pkg/cmd.Main({{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}}, {0xc00011e010, 0x4, ...})
        C:/Users/zb/Projects/chezmoi/pkg/cmd/cmd.go:128 +0xc9
main.main()
        C:/Users/zb/Projects/chezmoi/main.go:26 +0x125
```

The issue is that when AbsPath.MustTrimDirPrefix is called here (https://github.com/twpayne/chezmoi/blob/master/pkg/chezmoi/sourcestate.go#L524) it's actually `p` that doesn't end with a slash, not `dirPrefixAbsPath`. So it's trying to remove the prefix `H:/` from `H:`, which of course doesn't work.

Come to think of it, this should actually probably be caught by the early exit check on line 146 -- `H:` and `H:/` should be treated as the same path.  Unfortunately, even if you fix that, we _still_ have a problem.  `SourceState.AddDestAbsPathInfos` calls `destAbsPath.Dir()` in a loop, going up the tree till it gets to the home directory.  On Windows, when the home directory we're looking for is the root of a drive, this loop will eventually call `path.Dir("H:")`, which (and I still can't quite believe this) actually returns just `.`.  Then the SourceState will try to trim the `H:/` prefix off of `.`, which is clearly nonsense :smile:
</details>

zb140 (approved):
Awesome, looks great!

Issue #463, opened 2019-10-20
Custom data fields appear as all lowercase strings


spatenotte:
#### Troubleshooting
- [x] I ran `chezmoi doctor` and fixed as many issues as possible and retried my command.
- [x] I have double checked the instructions for [Installation](https://github.com/twpayne/chezmoi/blob/master/docs/INSTALL.md) and the [Quickstart Guide](https://github.com/twpayne/chezmoi/blob/master/docs/QUICKSTART.md).
- [x] I have search for previous issues that have been reported/opened, and none of them relate to my problem.

Failure to follow the troubleshooting steps (and to report information requested below), may make it difficult to provide the proper support required. Furthermore, your issue may be closed without reply, if the instructions were not properly followed.

---

#### Description of issue

I added this block in my `.chezmoi.toml.tmpl`:
```
[data]
   isLinux = {{if eq .chezmoi.os “linux”}}true{{else}}false{{end}}
   isMac = {{if eq .chezmoi.os “darwin”}}true{{else}}false{{end}}
   isUnix = {{if eq .chezmoi.os “linux” }}true{{else if eq .chezmoi.os “darwin”}}true{{else}}false{{end}}
   isWindows = {{if eq .chezmoi.os “windows”}}true{{else}}false{{end}}
```

The file that got generated is this one:
```
[data]
    isLinux = false
    isMac = true
    isUnix = true
    isWindows = false
```

But trying to use those variables doesn't work, unless I write them down all in lowercase, i.e `islinux`.

A quick note is that it looks like my system (macOS Mojave, 10.14.6) has a non case-sensitive file system, which might or might not be the cause of this. But I would like to not have to change that setting because this machine is my main programming device, and I am afraid changing that setting might cause other issues. Also, all other `chezMoi` data strings are cased correctly (like `chezmoi.fullHostname`) 

**EDIT**: Looks like the same thing happens in WSL (Linux subsystem on Windows), so might not be related to the filesystem being case-sensitive.

I could also just use the `.chezmoi.os` data directly, but those data fields help me keep things more readable and avoid typos better.

Also, just wanted to say I really appreciate this software, it really helps me manage my dotfiles much better!

#### Command that failed

```
✔ sammy in ~/.local/share/chezmoi [master ↑·5| ✚ 2 … 11] 
17:28:00.3N $ cat .chezmoi.toml.tmpl 
[data]
    isLinux = {{if eq .chezmoi.os "linux"}}true{{else}}false{{end}}
[...]

PR #2011, opened 2022-04-13
docs: Add clarification about all lowercase custom data fields

skaldarnar:
Hi, 

I'm new to chezmoi and started to look into templates. I've spent quite some time to figure out why I could not get custom variables in the `data` section of the config to work until I realized that I've stumbled upon #463.

I propose to add a hint or disclaimer to the documentation here to make it easier for newcomers to spot this pitfall (and avoid it). The  FAQ added in  #671 contains a reference already, but I only found it by backtracking from the issue.

Maybe adding the hint (also) to [User Guide > Templating](https://www.chezmoi.io/user-guide/templating/) would also help?

Please let me know if I need to change or update anything.


twpayne:
This is great, thank you!

> Maybe adding the hint (also) to [User Guide > Templating](https://www.chezmoi.io/user-guide/templating/) would also help?

Such a change would be welcome too :)

PR #2022, opened 2022-04-23
Use index function to access an item in the example

antonalekseev:
This PR fixes an example in the documentation.

```
chezmoi execute-template "{{ (onepassword \"$UUID\").fields[1].value }}"
```

yields the following error

```
chezmoi: template: stdin:1: bad character U+005B '['
```


twpayne:
Thank you very much!

PR #2062, opened 2022-05-10
feat: Add support for Keeper password manager

### LINKED ISSUE #2061: Support Keeper password manager

Keeper password manager is not currently natively supported by chezmoi, requiring setup of secret.command

Keeper has a semi-official CLI called keeper commander that is cross platform and has a JSON-providing [get option](https://docs.keeper.io/secrets-manager/commander-cli/using-commander/command-reference/record-commands#get-command) or a less technical (you don't need a UID) [find password](https://docs.keeper.io/secrets-manager/commander-cli/using-commander/command-reference/record-commands#get-command) option that should be easy to integrate.

I would propose either implementing only the 'get' option (allowing the use of custom types and fields) or both the get option and the password option via a parameter (i.e. "password" vs "item") similar to the syntax of the existing BitWarden integration.

Similar implementations would be the existing [password managers](https://www.chezmoi.io/user-guide/password-managers/), documentation/references to the Commander cli in the docs under a Keeper page, or simply doing nothing and letting the user figure it out.

## Additional context

N/A

twpayne:
Fixes #2061.

@tajetaje this is an initial implementation and probably contains a few bugs. Would you be able to test this?

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md
@@ -59,6 +59,8 @@ The following configuration variables are available:
 | `keepassxc`    | `args`                | []string | *none*                   | Extra args to KeePassXC CLI command                    |
 |                | `command`             | string   | `keepassxc-cli`          | KeePassXC CLI command                                  |
 |                | `database`            | string   | *none*                   | KeePassXC database                                     |
+| `keeper`       | `args`                | []string | *none*                   | Extra args to Keeper CLI command                       |
+|                | `command`             | string   | `keeper`                 | Keeper CLI command                                     |
 | `lastpass`     | `command`             | string   | `lpass`                  | Lastpass CLI command                                   |
 | `merge`        | `args`                | []string | *see `merge` below*      | Args to 3-way merge command                            |
 |                | `command`             | string   | `vimdiff`                | 3-way merge command                                    |
--- a/assets/chezmoi.io/docs/reference/templates/keeper-functions/index.md
+++ b/assets/chezmoi.io/docs/reference/templates/keeper-functions/index.md
@@ -1,0 +1,7 @@
+# Keeper functions
+
+The `keeper*` functions return data from [Keeper](https://www.keepersecurity.com/)
+[Commander CLI](https://docs.keeper.io/secrets-manager/commander-cli) (`keeper`).
+
+The command used can by changed by setting the `keeper.command` configuration
+variable, and extra arguments can be added by setting `keeper.args`.
--- a/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeper.md
+++ b/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeper.md
@@ -1,0 +1,6 @@
+# `keeper` *uid*
+
+`keeper` returns structured data retreived from
+[Keeper](https://www.keepersecurity.com/) using the [Commander
+CLI](https://docs.keeper.io/secrets-manager/commander-cli). *uid* is passed to
+`keeper get --format=json` and the output is parsed as JSON.
--- a/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeperDataFields.md
+++ b/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeperDataFields.md
@@ -1,0 +1,12 @@
+# `keeperDataFields` *uid*
+
+`keeperDataFields` returns the `.data.fields` elements of `keeper get
+--format=json *uid*` indexed by `type`.
+
+## Example
+
+```
+url = {{ (keeperDataFields "$UID").url }}
+login = {{ index (keeperDataFields "$UID").login 0 }}
+password = {{ index (keeperDataFields "$UID").password 0 }}
+```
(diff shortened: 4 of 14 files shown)


tajetaje:
> It seems to recognize the template fine, it parses it and throws an error when keeper isn't on PATH. However when I apply it with everything set up I get:
> 
> ```
> ❯ go run . apply
> User(Email): 
> 
> ^Csignal: interrupt
> ```
> 
> (I entered my email after the `User(email):` line)
> 
> I let it sit there for a minute but it did nothing. Honestly, I don't even know why it is asking me to log in as keeper as I have my password saved and the Keeper CLI doesn't ask for it when I run it myself. Does chezmoi not run it under the context of the current user?

UPDATE: While it still shouldn't even be asking for a password, it was actually working, it's just all the 2FA  prompt text was not showing up for some reason: 
```
❯ go run . apply
User(Email): # Entered email here, nothing showed up
1 # There should have been a prompt saying press 1 to receive an SMS
2fa_duration=forever code # Here and for the next one there should be a prompt saying I can adjust the 2fa trust duration or enter my 2FA code, but I just typed 1 and hit enter and it worked
860153 # Typed my passcode and it worked, kinda weird that it and the "1" above showed up but nothing else did
Password:  # Entered password here, nothing showed up
chezmoi: template: dot_gitconfig.tmpl:54:14: executing "dot_gitconfig.tmpl" at <keeperDataFields "BOX3BVx_3aiQyljiwm3QSA">: error calling keeperDataFields: invalid character 'T' looking for beginning of value # dunno what that's about
exit status 1
```

On a subsequent run, removing the line "{{ index (keeperDataFields "BOX3BVx_3aiQyljiwm3QSA").password 0 }}" (I assume I typed something wrong) from my test file, it only asked for a passoword (still no text beyond the prompt), and ran correctly, outputting the expected data

tajetaje:
OK, the reason it doesn't notice I am already signed in is because I installed using pip, the following is a quote from the docs:

> If you use Commander from a PIP3 or source installation, the configuration file will be created in the current folder where the Commander executable is started from.
>
> You can specify the config file to use when launching Commander, for example:
> `$ keeper shell --config /path/to/config.json`

Windows and Mac will probably use ~/.keeper/config.json as that is the default for binary installs, but since keeper can only be (officially) installed with pip3 all linux users will have that issue

tajetaje:
Also a note that keeper cannot necessarily be found at /bin/keeper as PIP might put it elsewhere, though I think simply documenting that it should be symlinked to that location should be enough for most anyone.

tajetaje:
Also I would suggest changing the docs for the password specific function as follows
````patch
diff --git a/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeperFindPassword.md b/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeperFindPassword.md
index 925a71fd..ca7b2d84 100644
--- a/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeperFindPassword.md
+++ b/assets/chezmoi.io/docs/reference/templates/keeper-functions/keeperFindPassword.md
@@ -1,3 +1,3 @@
-# `keeperFindPassword` *uid*
+# `keeperFindPassword` *query*
 
-`keeperFindPassword` returns the output of `keeper find-password *uid*`.
+`keeperFindPassword` returns the output of `keeper find-password *query*`. `query` can be a uid or a path.
diff --git a/assets/chezmoi.io/docs/user-guide/password-managers/keeper.md b/assets/chezmoi.io/docs/user-guide/password-managers/keeper.md
index 1b568c97..848cd954 100644
--- a/assets/chezmoi.io/docs/user-guide/password-managers/keeper.md
+++ b/assets/chezmoi.io/docs/user-guide/password-managers/keeper.md
@@ -7,10 +7,11 @@ data as a template function.
 Create a persistent login session as [described in the Command CLI
 documentation](https://docs.keeper.io/secrets-manager/commander-cli/using-commander/logging-in#persistent-login-sessions).
 
-Passwords can be retrieved with the `keeperFindPassword` template function, for example:
+Passwords can be retrieved with the simpler `keeperFindPassword` template function, for example:
 
 ```
-examplePassword = {{ keeperFindPassword "example.com" }}
+examplePasswordFromPath = {{ keeperFindPassword "office/Zoom" }}
+examplePasswordFromUid = {{ keeperFindPassword "rvwIBG_ban2VTH64OsnzLn" }}
 ```
 
 For retrieving more complex data, use the `keeper` template function with a UID

````

**Comment by twpayne** [comment]
Thank you for the extensive testing and suggestions :) I've updated the PR based on your feedback and added you as co-author. Would you be able to test again?

**Comment by twpayne** [comment]
> Maybe allow setting a keeper config location in the chezmoi config (since it is machine specific)?

I've updated the PR to a `keeper.args` configuration variable in chezmoi's configuration file, for example:

```toml
[keeper]
    args = ["--config", "/path/to/config.json"]
```

twpayne:
> Also a note that keeper cannot necessarily be found at /bin/keeper as PIP might put it elsewhere, though I think simply documenting that it should be symlinked to that location should be enough for most anyone.

chezmoi only requires that `keeper` be in your `$PATH`. You can also set a specific location for the `keeper` binary by setting the `keeper.command` configuration variable.

PR #2071, opened 2022-05-17
fix: Fix use of slashes in glob on Windows

### LINKED ISSUE #2063: glob template function does not work with joinPath on Windows
The newly added `glob` template function requires all input paths to use forward slashes. Meanwhile, the `joinPath` template function specifically coerces all slashes to the platform-specific separator. This means that on Windows, where backslashes are used, `joinPath` cannot be combined with `glob`.

To resolve this, `filepath.ToSlash()` could be called on all inputs before they are passed to `doublestar.Glob()`.

twpayne:
Fixes #2063.

Co-authored-by: @neersighted


neersighted:
That did indeed fix my usage!

PR #2086, opened 2022-05-24
feat: Add windowsVersion to builtin template data

### LINKED ISSUE #2085: Add OS version information for more platforms in built-in `data`

I'm in the process of writing some scripts (notably package installation scripts) for Windows which differ depending on the OS version.

Of course, it's possible to determine the version within the script itself, but it would certainly be nice to have this built into chezmoi if at all possible. I believe there is an existing precedent with `osRelease` being provided on Linux, though I do also acknowledge that the existence of `os-release` made this more trivial than on other platforms.

I'm less familiar with macOS so I will defer to those who are more knowledgeable.

https://stackoverflow.com/a/44376544/11000543 offers a solution for getting the Product Name, Major and Minor versions as well as the Build Number on Windows using [`golang.org/x/sys/windows/registry`](https://pkg.go.dev/golang.org/x/sys@v0.0.0-20220520151302-bc2c85ada10a/windows/registry).

[sindresorhus/windows-release](https://github.com/sindresorhus/windows-release/blob/edb4b7b5efc4c92bcd32d41701f6f727dff0c485/index.js) is a Node.js solution demonstrating a simple version mapping which even includes Windows Server.

Perhaps this could be used to populate a fake `osRelease`? Or maybe introduce `windowsRelease`/`darwinRelease` or `osInfo`etc.

Populating version information within `data` in chezmoi's config by using various `{{ output ... }}`'s etc.

This is obviously sufficient enough, but it only benefits me unless I share the template and people go out of their way to find and use it. We can prevent a lot of reinventing the wheel.

Do you think this is something which should be provided by chezmoi?

## Additional context

N/A

twpayne:
Fixes #2085.

### DIFF
--- a/pkg/cmd/config_windows.go
+++ b/pkg/cmd/config_windows.go
@@ -2,9 +2,11 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	"go.uber.org/multierr"
 	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
 	"golang.org/x/term"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/cmd/config_windows.go
```diff
@@ -42,3 +44,32 @@ func (c *Config) readPassword(prompt string) (password string, err error) {
...
+	registryKey, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, registry.QUERY_VALUE)
+	if err != nil {
+		return nil, err
+	}
+	windowsVersion := make(map[string]interface{})
+	for _, name := range []string{
+		"CurrentBuild",
+		"CurrentVersion",
+		"DisplayVersion",
+		"ProductName",
```
bradenhilton, line 58: Could we also include `"EditionID"`?

Use case(s):

- Some editions have more restrictions than others on how software can be installed.
- Enterprise and Server editions have [built-in methods to block telemetry](https://docs.microsoft.com/en-us/windows/privacy/manage-connections-from-windows-operating-system-components-to-microsoft-services) but other editions have to use third-party scripts etc.

twpayne: Of course! Thanks for the review :)

PR #2093, opened 2022-05-29
docs: chezmoi_modify_manager

VorpalBlade:
Add a hint in the section on `modify` as well as in related software.

Also in a separate commit I organized related software page into sub-sections. So far I only added a separate section for Editor integration. In the future, as there is more software it may also make sense to split "Other" into further sections.


twpayne:
This is great, thank you very much!

twpayne:
@VorpalBlade FYI https://www.lorenzobettini.it/2025/09/managing-kde-dotfiles-with-chezmoi-and-chezmoi-modify-manager/

PR #2095, opened 2022-05-29
Fixed typo in kubeconfig documentation

fire833:
When describing managing a file's permissions without managing the contents itself, and uses kubeconfig as an example, the docs note to create a file `dot_kube/private_dot_config` in the source directory, when based on the chezmoi source naming scheme, it should be `dot_kube/private_config`. This changes that disparity.

 Signed-off by: Kendall Tauser <kttpsy@gmail.com>


twpayne:
Thank you very much :)

PR #2112, opened 2022-06-04
fix: Improve quality of POSIX shell scripts

Luexa:
Hello, I've made some improvements to `install.sh` and `stow-to-chezmoi.sh`. I've made sure to update `install.sh.tmpl` and committed the output of generating that.

General improvements:
 * Use `${N}` syntax for all parameter expansions.
 * Supply `--` to commands whose positional arguments are based on environment variables, output of system utilities, or user input. This protects against an argument being interpreted as an option by a command.
 * Avoid use of `echo` which behaves inconsistently between shells and platforms.
 * Clean up temporary files when interrupted by `SIGINT` or `SIGTERM`.

`install.sh` improvements:
 * Quote all variable assignments that involve parameter expansion or command substitution to protect them from pathname expansion and quote removal.
 * Quote parameter expansions that occur within command substitutions.
 * Use `install -t <directory> <file>` instead of `install <file> <hopefully_directory_but_maybe_file>`.
 * Fixed usage of colon in unquoted word (this is unspecified behavior).
 * Rather than relying on unquoted parameter expansion of `${EXECARGS}`, run `shift "$((OPTIND - 1))"` in `main` to and use `"${@}"` to expand parameters as-is. This makes parameters safe from field splitting, quote removal, and pathname expansion, and fixes a suppressed shellcheck warning.

`stow-to-chezmoi.sh` improvements:
 * Change shebang from `/usr/bin/env bash` to `/bin/sh`. I'm fairly certain there was only one bashism in the entire script (there was also a nonstandard `echo` option, but `dash` has this too; still, `printf` is better).
 * Replace unnecessary `which` command with `command -v` (a standard shell builtin). This fixes a shellcheck warning.
 * Account for trailing whitespace in resolved symbolic links.
 * Filter out paths that contain line feeds to avoid confusing the `read` builtin.
 * Replace usage of nonstandard `read -p` by instead printing the prompt with `printf`.

### DIFF
--- a/assets/scripts/install.sh
+++ b/assets/scripts/install.sh
@@ -93,7 +93,7 @@ main() {
 		install -d "${BINDIR}"
 	fi
 	BINARY="chezmoi${BINSUFFIX}"
-	install -t "${BINDIR}/" "${tmpdir}/${BINARY}"
+	install -- "${tmpdir}/${BINARY}" "${BINDIR}/"
 	log_info "installed ${BINDIR}/${BINARY}"
 
 	if [ -n "${1+n}" ]; then
--- a/assets/scripts/install.sh
+++ b/assets/scripts/install.sh
@@ -7,19 +7,19 @@
 
 set -e
 
-BINDIR=${BINDIR:-./bin}
-CHEZMOI_USER_REPO=${CHEZMOI_USER_REPO:-twpayne/chezmoi}
+BINDIR="${BINDIR:-./bin}"
+CHEZMOI_USER_REPO="${CHEZMOI_USER_REPO:-twpayne/chezmoi}"
 TAGARG=latest
 LOG_LEVEL=2
-EXECARGS=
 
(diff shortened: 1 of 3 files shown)


Luexa:
Ah, the `-t` option is nonstandard. I'll revert that change.

Luexa:
Force pushed to make that a chore instead of a fix since it's trivial and fixes something that never made it to a release.

twpayne:
Thank you very much for this excellent pull request!

Luexa:
No problem! I have a tendency of going down the rabbit hole with shell scripts, especially ones that encourage the `curl | sh` trick.

twpayne:
chezmoi's suggested use of `curl | sh` is definitely not great from a security perspective. There are a bunch of other [installation methods](https://www.chezmoi.io/install/) available to mitigate this risk, and chezmoi is a clear high-value target for supply chain attacks. Thank you again for your attention to detail here :)

PR #2118, opened 2022-06-07
docs: Refactor Windows chassisType template

bradenhilton:
- WMI cmdlets have been [deprecated in favor of CIM](https://docs.microsoft.com/en-us/powershell/scripting/learn/ps101/07-working-with-wmi?view=powershell-7.2) for some time now:

    > The WMI cmdlets are deprecated so my recommendation is to use the CIM cmdlets instead of the older WMI ones.

- In PowerShell, `echo` is an alias for `Write-Output`, but it's possible to reassign the alias to something else, so it's better to be explicit.
- `trim` is required to remove the dangling CRLF line ending. Without this, the result is:
    ```toml
    [data]
    chassistype = "desktop\r\n" # or chassistype = "laptop\r\n"
    ```


twpayne:
Awesome, thank you!

PR #2120, opened 2022-06-07
feat: Add quoteList template function

twpayne:
Refs #2119.

cc @bradenhilton


bradenhilton:
I was actually preparing an issue intended to gather implementation details about this.

Given that we offer `quote` and `squote` via `sprig`, we should probably also offer `squoteList` at a minimum.

I was unsure if we should assume that users would use `toStrings` beforehand (which would mean 2 passes are made, one to convert items to strings, then another to quote each string), or if we should instead combine `toStrings` with `quote`/`squote` into a single pass.

Then that got me thinking, would it be better to provide a `map` higher-order function? I'm not sure how feasible this is, but perhaps we could do something like `map (list "hello" "world") {s}quote`/`map $list {upper|lower}`/`map $list (replace "foo" "bar")`/`map $list {snake|camel|kebab|swap}case`/`map $numList add1` etc.

twpayne:
As template functions are rather a mixed bag of quality (the sprig functions do not follow Go's naming convention, for example, and have not had a release in 18 months), I'd rather only add template functions that we know are needed.

For a higher level map function, I agree this would be a more general solution, but Go's text/template language does not support this.

bradenhilton:
Everything appears to be working as expected

<details>
<summary><code>quoteList_test.txt.tmpl</code></summary>

```
{{ $args := list "" -}}
{{ "{{" }} $args := list "" {{ "-}}" }}

{{ $args = list "alpha" "beta" "gamma" -}}
{{ "{{" }} $args = list "alpha" "beta" "gamma" {{ "-}}" }}
I: args = {{ "{{" }} $args | quoteList {{ "}}" }}
O: args = {{ $args | quoteList }}
I: args = [{{- "{{-" }} $args | quoteList | join ", " {{ "-}}" -}}]
O: args = [{{- $args | quoteList | join ", " -}}]
I: args = [
    {{ "{{" }} $args | quoteList | join ",\n    " {{ "}}" }}
]
O: args = [
    {{ $args | quoteList | join ",\n    " }}
]

{{ $args = list 1 2 3 -}}
{{ "{{" }} $args = list 1 2 3 {{ "-}}" }}
I: args = [{{- "{{-" }} $args | quoteList | join ", " {{ "-}}" -}}]
O: args = [{{- $args | quoteList | join ", " -}}]

{{ $args = list "" -}}
{{ $args = mustCompact $args -}}
{{ "{{" }} $args = list "" {{ "-}}" }}
{{ "{{" }} $args = mustCompact $args {{ "-}}" }}
I: args = [{{- "{{-" }} $args | quoteList | join ", " {{ "-}}" -}}]
O: args = [{{- $args | quoteList | join ", " -}}]
```

</details>

<details>
<summary>Output</summary>

```console
chezmoi-twpayne on  quote-list [?] via 🐹 v1.18.3
❯ Get-Content '.\quoteList_test.txt.tmpl' | go run . execute-template
{{ $args := list "" -}}

{{ $args = list "alpha" "beta" "gamma" -}}
I: args = {{ $args | quoteList }}
O: args = ["alpha" "beta" "gamma"]
I: args = [{{- $args | quoteList | join ", " -}}]
O: args = ["alpha", "beta", "gamma"]
I: args = [
    {{ $args | quoteList | join ",\n    " }}
]
O: args = [
    "alpha",
    "beta",
    "gamma"
]

{{ $args = list 1 2 3 -}}
I: args = [{{- $args | quoteList | join ", " -}}]
O: args = ["1", "2", "3"]

{{ $args = list "" -}}
{{ $args = mustCompact $args -}}
I: args = [{{- $args | quoteList | join ", " -}}]
O: args = []
```

</details>

<sup>(That's a lot of braces)</sup>

Do you intend to write tests?

Do you think it's worth including the `[{{- $list | quoteList | join ", " -}}]` example in the doc page?

PR #2129, opened 2022-06-12
docs: Use modeline to set filetype in VIM

CyrusYip:


twpayne:
Thank you very much :)

PR #2145, opened 2022-06-17
Support determining FQDN via /etc/myname

ryanakca:
OpenBSD hosts configure their hostname via the /etc/myname file.
This file should contain a single line that is the host's FQDN.
See myname(5) for details.

### DIFF
--- a/pkg/chezmoi/chezmoi_unix.go
+++ b/pkg/chezmoi/chezmoi_unix.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"io/fs"
 	"regexp"
+	"runtime"
 	"strings"
 
 	vfs "github.com/twpayne/go-vfs/v4"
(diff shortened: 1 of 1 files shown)


pkg/chezmoi/chezmoi_unix.go
```diff
@@ -49,6 +52,23 @@ func etcHostnameFQDNHostname(fileSystem vfs.FS) (string, error) {
...
 
+// etcMynameFQDNHostname returns the FQDN hostname from parsing /etc/myname.
+// See OpenBSD's myname(5) for details on this file.
+func etcMynameFQDNHostname(fileSystem vfs.FS) (string, error) {
+	contents, err := fileSystem.ReadFile("/etc/myname")
+	if err != nil {
+		return "", err
+	}
+	s := bufio.NewScanner(bytes.NewReader(contents))
+	for s.Scan() {
```
twpayne, line 63: This loop needs to ignore comment lines as well.

pkg/chezmoi/chezmoi_unix.go
```diff
@@ -26,6 +26,9 @@ func FQDNHostname(fileSystem vfs.FS) string {
 	if fqdnHostname, err := etcHostsFQDNHostname(fileSystem); err == nil && fqdnHostname != "" {
 		return fqdnHostname
 	}
+	if fqdnHostname, err := etcMynameFQDNHostname(fileSystem); err == nil && fqdnHostname != "" {
```
twpayne, line 29: As only OpenBSD uses this, could you please put this inside an `if runtime.GOOS == "openbsd" { ... }` conditional.

twpayne (commented):
Looks great! Just a couple of small nits.

pkg/chezmoi/chezmoi_unix.go
ryanakca: Done!

pkg/chezmoi/chezmoi_unix.go
ryanakca: Done as well!

twpayne:
Thank you very much :)

PR #2168, opened 2022-06-22
docs: Add CPU cores/threads template

bradenhilton:
Tested with:
- 4c/4t desktop (Windows + Linux)
- 2c/4t laptop (Windows + Linux)
- 4c/8t 2015 MacBook Pro (Darwin)


twpayne:
Very cool, thank you!

bradenhilton:
I've just now noticed the config template file name in the example is missing the dot prefix. Would you mind at all adding it?

PR #2226, opened 2022-07-21
feat: Add generate command to generate install script

### LINKED ISSUE #2186: Add a command to generate an appropriate install.sh for usage in codespaces or devcontainers

Usage with codespaces or devcontainers currently requires copying an appropriate `install.sh` from some other repo. 

Since IMHO this can be quite a common scenario for using chezmoi, it might be nice to include a command that generates an appropriate one. e.g. `chezmoi generate-install-sh`

Keep doing so manually... Based on docs or other repos.

twpayne:
Fixes #2186.

@segevfiner is this what you had mind? Would you be able to test this?

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/generate.md
+++ b/assets/chezmoi.io/docs/reference/commands/generate.md
@@ -1,0 +1,13 @@
+# `generate` *file*
+
+Generates *file* for use with chezmoi. The currently supported *file*s are:
+
+| File         | Description                                                |
+| ------------ | ---------------------------------------------------------- |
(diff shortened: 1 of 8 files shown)


segevfiner:
Yeah that's the idea. Seems to work. You can still consider some ergonomics here for the command, such as whether you want it to write the file by itself to the correct location, `chmod` it, add it to `.chezmoignore`, etc. But it's useful as it is for getting started with chezmoi in containers.

twpayne:
Thanks for testing. I'll merge this as-is for now, and we can always iterate further to improve it.

PR #2204, opened 2022-07-09
feat: Add prompt*Once init template functions

twpayne:
Fixes #2184.

WIP. Needs tests and documentation.

PR #2240, opened 2022-07-29
feat: Add --prompt flag to init command

twpayne:
As [suggested by](https://github.com/twpayne/chezmoi/pull/2204#issuecomment-1198823555) @felipecrs in #2204.


twpayne:
> Calling init with `--data=false` leads to **B**, so calling init with `--prompt` may results in **A**...

Good question and thank you for the review! I believe that calling init with `--prompt` always results in **B**. Note that the default value is only used when the user replies to the prompt with an empty string. However, I might have missed something...

PR #2247, opened 2022-08-06
feat: Add --interactive flag

### LINKED ISSUE #2245: Add a flag to chezmoi apply to always show a prompt when chezmoi going to make changes to destination state

When running `chezmoi apply` chezmoi shows the prompt i.e`has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]?` only if the destination state is diverted from the source state.
If the changes are in the source state. When running `chezmoi apply` chezmoi directly applies the change and shows diff if `-v` flag is used.

When managing a single file via external e.g plugins for zsh or mpv, `chezmoi apply` will apply the changes without the prompt.
Sometimes you don't want to upgrade these plugins. A prompt will be helpful to quickly diff or skip the change.

If possible a flag can be added to `chezmoi apply` which will always show a prompt when chezmoi going to make changes to destination state. It will cover the external case and will make `chezmoi apply` more controllable.

Run `chezmoi diff` before `chezmoi apply` and exclude the file if you don't want to update.

## Additional context

None

twpayne:
Fixes #2245.

With this you can do:

```console
$ chezmoi apply --interactive
```

and you'll be prompted for each changed entry.

@sm1999 is this what you had in mind? Would you be able to test this?

### DIFF
--- a/assets/chezmoi.io/docs/reference/command-line-flags/common.md
+++ b/assets/chezmoi.io/docs/reference/command-line-flags/common.md
@@ -23,6 +23,10 @@ with a `no`.
 Regenerate and reread the config file from the config file template before
 computing the target state.
 
+## `--interactive`
+
+Prompt before applying each target.
+
 ## `-r`, `--recursive`
 
 Recurse into subdirectories, `true` by default.
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -126,12 +126,12 @@ func (v VersionInfo) MarshalZerologObject(e *zerolog.Event) {
 // Main runs chezmoi and returns an exit code.
 func Main(versionInfo VersionInfo, args []string) int {
 	if err := runMain(versionInfo, args); err != nil {
-		if s := err.Error(); s != "" {
-			fmt.Fprintf(os.Stderr, "chezmoi: %s\n", s)
+		var errExitCode chezmoi.ExitCodeError
+		if errors.As(err, &errExitCode) {
+			return int(errExitCode)
 		}
-		errExitCode := chezmoi.ExitCodeError(1)
-		_ = errors.As(err, &errExitCode)
-		return int(errExitCode)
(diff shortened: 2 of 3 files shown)


sm1999:
Yes. I tested it locally and it works as expected.
Liked the new prompt. I sometimes pressed `a` accidentally on the old prompt thinking `a` for accept.
Thank you so much!

twpayne:
Great! Thank you for the testing :)

sm1999:
@twpayne in the new prompt pressing a for apply changes to all files only change the current file for which prompt is shown and fallback to the old prompt.
If I understand correctly `c.force = true` is missing.
Following diff fixes it.
```
diff --git a/pkg/cmd/config.go b/pkg/cmd/config.go
index 2b9b72e7..6dc19f2e 100644
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -881,6 +881,7 @@ func (c *Config) defaultPreApplyFunc(
 				return chezmoi.Skip
 			case choice == "all":
 				c.interactive = false
+                c.force = true
 				return nil
 			case choice == "quit":
 				return chezmoi.ExitCodeError(0)

```

sm1999:
@twpayne ping

twpayne:
@sm1999 Why are you pinging me?

sm1999:
I was uncertain that you saw my comment above.
Basically what is happening is pressing a for all fallbacks to the old prompt for the remaining changed files. 
```
❯ chezmoi apply -n
Apply .config/git/config [diff,yes,no,all,quit]? a
.config/htop/htoprc has changed since chezmoi last wrote it [diff,overwrite,all-overwrite,skip,quit]?
```
Is it intentional?

twpayne:
Yes, this is intentional. There are two conditions here:
1. Do you want to be prompted for every target that chezmoi might change? (`--interactive`)
2. Do you want to overwrite changes made by other programs? (`--force`)

It's deliberate to keep the two separate.

sm1999:
Thanks for the explanation!

PR #2272, opened 2022-08-19
feat: Add initial --watch flag to edit command

### LINKED ISSUE #752: apply file changes on write with edit --apply

twpayne:
Fixes #752.
Refs #2269.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/edit.md
+++ b/assets/chezmoi.io/docs/reference/commands/edit.md
@@ -22,6 +22,22 @@ Invoke the editor with a hard link to the source file with a name matching the
 target filename. This can help the editor determine the type of the file
 correctly. This is the default.
 
+## `--watch`
+
+Automatically apply changes when files are saved.
+
+!!! warning
+
+    `--watch` is a beta feature with limitations. Known limitations include:
+
+    * Only works when `chezmoi edit` is invoked with arguments (i.e.
+      argument-free `chezmoi edit` is not supported).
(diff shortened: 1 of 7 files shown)


eddy-geek:
[FAQ "Why doesn't chezmoi use symlinks like GNU Stow?"](https://www.chezmoi.io/user-guide/frequently-asked-questions/design/#why-doesnt-chezmoi-use-symlinks-like-gnu-stow) may need an update:

> The only advantage to using GNU Stow-style symlinks is that changes that you make to the dotfile's contents in the centralized directory are immediately visible, whereas chezmoi currently requires you to run chezmoi apply or chezmoi edit --apply. chezmoi will likely get an alternative solution to this too, see [#752](https://github.com/twpayne/chezmoi/issues/752).

twpayne:
Good catch, thanks. Done in 4d6861246e7520c30b9912aa10d63202715283f5.

Issue #2215, opened 2022-07-14
Add more config management functions


mcexit:
## Is your feature request related to a problem? Please describe.

Functions are primarily limited to the Sprig library, which does not handle a lot of common config management functions.

## Describe the solution you'd like

I'd like to see more functions available to address config management. The main functions here would to:

- `fromToml`
- `toToml`
- `fromIni`
- `toIni`
- `replaceRE`
- `partial` and `return`
- `comment`

Additionally any merge functionality that may not be covered by Sprig functions. This helps streamline a common use case of overriding default config file values.

`fromToml`, `toToml`, `fromIni`, and `toIni` are all pretty self-explanatory. Here are some examples of where those are already implemented in Go.

1. https://gohugo.io/functions/transform.unmarshal/
2. https://docs.gomplate.ca/functions/data/
3. https://github.com/go-ini/ini

`replaceRE` is the name of a function in Hugo[^1] that allows for proper pipelining[^2] and a limit parameter.

`partials` are like templates[^3], except that you can do pipelining on the results, with an optional `return` function[^4] in the partial template.

`comment` is very similar to the indent Sprig functions, and takes inspiration from an Ansible filter[^5]. This would take either a single or multiline string and adds the specified comment character(s) at the beginning of each newline. A common use case for this is a disclaimer for users that may try editing a file directly outside of the sourceDir:

---
`.local/share/chezmoi/.chezmoitemplates/disclaimer`
```
This file is managed by chezmoi. Do not edit!
To change it, modify {{ joinPath .chezmoi.sourceDir .chezmoi.sourceFile }}
```

`.local/share/chezmoi/private_dot_config/example/exampleconfig.conf.tmpl`
```
partial "examplepartial" | comment "#"
```
---

## Describe alternatives you've considered

I've already to the larger extent implemented these in each file template, but it requires a lot of duplication.

[^1]: https://gohugo.io/functions/replacere/
[^2]: https://github.com/Masterminds/sprig/issues/86
[^3]: https://gohugo.io/templates/partials/
[^4]: https://gohugo.io/templates/partials/#returning-a-value-from-a-partial
[^5]: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#adding-comments-to-files

twpayne:
Thank you for the extremely complete and well-researched feature request!

#2218 and #2237 add some of the functions.

[...]

PR #2278, opened 2022-08-22
feat: Add toIni template function

twpayne:
Refs #2215.

@mcexit would you be able to test this? The tests in this PR give an idea of what `toIni` produces for various inputs. Does this particular variant of ini look reasonable?


twpayne:
Thanks for the feedback :)

#2311 modifies `toIni` to only use quotes for strings when necessary.

Sadly the Viper issue (https://github.com/spf13/viper/issues/260) has been open for almost six years and so is unlikely to be ever fixed. chezmoi will likely have to replace Viper with something else.

Thanks for the offer of a donation to support chezmoi. Instead, I would ask you to donate to a charity of your choice.

PR #2301, opened 2022-08-27
feat: Add support for passhole password manager.

JackTheMico:
Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

Completed passhole adding.

Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

Add binaryCheck for passhole cli.

Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

Fix based on go lint

Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

Bugfix for passhole testscript

Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

Remove useless err

Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

fix(class): Using LogCmdOutput instead of IdempotentCmdOutput

fix passhole test scripts

Signed-off-by: Jack Deng <dlwxxxdlw@gmail.com>

PR #2317, opened 2022-08-31
feat: Add passhole password manager support

twpayne:
Co-authored-by: @JackTheMico

Replaces #2301.

This should not be merged until passhole supports reading a password from stdin (which [version 1.9.9 does not support](https://github.com/Evidlo/passhole/blob/v1.9.9/passhole/passhole.py#L370-L388)).


twpayne:
I opened https://github.com/Evidlo/passhole/issues/56 to track the changes needed to passhole.

PR #2358, opened 2022-09-14
docs: change remote origin - from ssh to https

muescha:
i have setup already my git to use with https - i got authentification errors when i used this settings. i have to find out why my authentification fails.

also GitHub recommends the https version

PS: maybe we show both variants here


twpayne:
Thanks :)

PR #2372, opened 2022-09-19
docs: Elaborate on once vs onchange attributes

dslatkin:
I thought it would be useful to elaborate in the documentation a bit about `once` vs. `onchange` attributes. It's a bit hard to remember and I keep going to [this thread](https://github.com/twpayne/chezmoi/discussions/2210) to remind myself when needed, and so I thought it would be nicer to just check the reference instead.


twpayne:
Thanks!

PR #2409, opened 2022-09-30
feat: Add eqFold template function

bradenhilton:
Adds a template function which utilizes [`strings.EqualFold()`](https://pkg.go.dev/strings#EqualFold) for robust comparison.

Removes the need for `{{ if eq "foo" ("FOO" | lower) }}` and similar.

The name is placeholder, as "equal fold" doesn't seem to be a particularly standard name for case-insensitive string comparison, so I'm more than happy to change it if you know a better name.

Also, do you think it makes sense to accept an arbitrary number of strings, similar to `eq`? Perhaps something other than `strings.EqualFold()` would be required for this.

### DIFF
--- a/pkg/cmd/templatefuncs.go
+++ b/pkg/cmd/templatefuncs.go
@@ -66,6 +66,18 @@ func (c *Config) commentTemplateFunc(prefix, s string) string {
 	return builder.String()
 }
 
+func (c *Config) eqFoldTemplateFunc(first, second string, more ...string) bool {
+	if strings.EqualFold(first, second) {
+		return true
+	}
+	for _, s := range more {
+		if strings.EqualFold(first, s) {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *Config) fromIniTemplateFunc(s string) map[string]any {
 	file, err := ini.Load([]byte(s))
 	if err != nil {
--- a/assets/chezmoi.io/docs/reference/templates/functions/eqFold.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/eqFold.md
@@ -1,0 +1,15 @@
+# `eqFold` *string1* *string2* [*extraString*...]
+
+`eqFold` returns the boolean truth of comparing *string1* with *string2* and
+any number of *extraString*s under Unicode case-folding.
+
+!!! example
+
+    ```
+    {{ $commandOutput := output "path/to/output-FOO.sh" }}
+    {{ if eqFold "foo" $commandOutput }}
+    # $commandOutput is "foo"/"Foo"/"FOO"...
+    {{ else if eqFold "bar" $commandOutput }}
+    # $commandOutput is "bar"/"Bar"/"BAR"...
+    {{ end }}
+    ```
--- a/assets/chezmoi.io/mkdocs.yml
+++ b/assets/chezmoi.io/mkdocs.yml
@@ -12,305 +12,306 @@ theme:
   logo: logo.svg
   language: en
   palette:
-  - media: "(prefers-color-scheme: light)"
-    scheme: default
-    primary: indigo
-    accent: indigo
-    toggle:
-      icon: material/toggle-switch
-      name: Switch to dark mode
-  - media: "(prefers-color-scheme: dark)"
-    scheme: slate
-    primary: indigo
-    accent: indigo
-    toggle:
-      icon: material/toggle-switch-off-outline
-      name: Switch to light mode
+    - media: "(prefers-color-scheme: light)"
+      scheme: default
+      primary: indigo
+      accent: indigo
+      toggle:
+        icon: material/toggle-switch
+        name: Switch to dark mode
+    - media: "(prefers-color-scheme: dark)"
+      scheme: slate
+      primary: indigo
+      accent: indigo
+      toggle:
+        icon: material/toggle-switch-off-outline
+        name: Switch to light mode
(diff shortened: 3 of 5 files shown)


twpayne:
This is a fairly specific function, and, as you say, the same functionality can be achieved by composing existing functions. How often would this function be used?

bradenhilton:
I can mainly see it being used when comparing hostnames, or comparing the output of commands/scripts/prompts etc to a known good value.

I have arbitrary extra arguments working in my local copy, I just haven't committed/pushed it yet.

It cuts down on the nested parentheses which I find difficult to keep track of on a single continuous line. Sure, you could factor out each value to a variable before checking I suppose, though I typically only really do that when I'm using it in multiple places.

I'm more than happy to close if you'd like to wait for more demand. It was still fun to work on and I will continue to do so.

twpayne:
Hmm, this still feels very specific. You can avoid parentheses by using a variable:

    {{ $hostname := lower .chezmoi.hostname }}
    {{ if eq $hostname "foo" }}
    ...
    {{ end }}

An arbitrary number of arguments (which I agree is a good enhancement) can be supported using Sprig's [`has` function](http://masterminds.github.io/sprig/lists.html):

    {{ if has (lower "BAR") (list "foo" "bar" "baz") }}
    ...
    {{ end }}

Thank you for this PR, let's wait to see if there's more demand to decide whether to merge it.

pkg/cmd/templatefuncs_test.go
```diff
@@ -277,3 +277,41 @@ func TestQuoteListTemplateFunc(t *testing.T) {
 		`"true"`,
 	}, actual)
 }
+
+func TestStrEqFoldTemplateFunc(t *testing.T) {
```
twpayne, line 281: As a high-level comment on this test, it's great that this PR includes a test. The test cases here are really testing Go's `strings.EqualFold` function. I think it's reasonable to assume that this works correctly as it's part of Go's standard library, which is already well tested, and so the extra test cases beyond the first one don't add a huge amount of value. Instead, I would suggest a high-level test like the [tests in `templatefuncs.txtar`](https://github.com/twpayne/chezmoi/blob/v2.24.0/pkg/cmd/testdata/scripts/templatefuncs.txtar) which test that the function is plumbed into chezmoi correctly.

twpayne: Correct, thanks!

pkg/cmd/templatefuncs.go
```diff
@@ -284,6 +284,21 @@ func (c *Config) statTemplateFunc(name string) any {
 	}
 }
 
+func (c *Config) strEqFoldTemplateFunc(s1 string, s2 ...string) bool {
+	if (len(s2) == 0) {
```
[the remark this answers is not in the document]
bradenhilton: Lesson learned: Don't write Go source files and PowerShell scripts at the same time.

twpayne:
Thanks for making the changes. Let's merge this when the tests pass.

bradenhilton:
Also, now that I'm looking at the function again, won't it panic by default if no arguments are passed/`s1` is missing? Passing a single argument is valid insofar as the function will run, but we need at least one extra argument. Maybe it's better to just panic with the static string `"expected 2 or more arguments, got 1"`.

pkg/cmd/templatefuncs.go
```diff
@@ -284,6 +284,21 @@ func (c *Config) statTemplateFunc(name string) any {
...
+	}
+
+	result := false
+	for _, s := range s2 {
+		result = strings.EqualFold(s1, s)
+		if !result {
+			break
+		}
+	}
+	return result
```
twpayne, line 299: We can simplify this code with an early return:

```go
for _, s := range s2 {
    if strings.EqualFold(s1, s) {
        return true
    }
}
return false
```

twpayne:
> Also, now that I'm looking at the function again, won't it panic by default if no arguments are passed/`s1` is missing? Passing a single argument is valid insofar as the function will run, but we need at least one extra argument. Maybe it's better to just panic with the static string `"expected 2 or more arguments, got 1"`.

Another option here is to write the function signature so that `text/template` catches this, i.e. make the signature

```go
func strEqFoldTemplateFunc(s, first string, more ...string) bool
```

which means that you have to pass at least two arguments and `text/template` will complain if you don't.

pkg/cmd/templatefuncs.go
bradenhilton: Isn't the intention to check if all provided strings match? My understanding is you would need to invert this as you would only return early if they don't at any point.

twpayne: What would be the use case for checking if all provided strings match the first argument? Surely you want to check if *any* provided strings match the first argument?

bradenhilton: Isn't that more of an `or` than an `eq`? We can certainly go in that direction, but the function probably needs a better name than `strEqFold`. Maybe something like `strCompareAny`?

bradenhilton:
I think everything is good to go now, thanks for all the help and patience.

twpayne:
This is great, thank you! Please do hit the "Rebase and merge" button as soon as the final test completes (probably in about 25 minutes, sorry that the `main / test-release` check is so slow).

felipecrs:
If sprig were still alive, this new function would be a great addition to it, I think.

PR #2418, opened 2022-10-04
Expose some chezmoi config in env vars

halostatue:
I've been playing with some of the runtime scripts in my configuration and realized that I could change most of them—if not all—from `run_*.sh.tmpl` to just regular shell scripts (so that I can run them without clearing state for things that occasionally get reset by the OS) if Chezmoi exported some of its configuration as environment variables.

The template values that I care about most are `.chezmoi.homeDir` and `.chezmoi.sourceDir`, but other values may be useful as well, so I have set the following values when the configuration object is constructed:

- `CHEZMOI_ARCH`
- `CHEZMOI_CACHE_DIR`
- `CHEZMOI_CONFIG_FILE`
- `CHEZMOI_HOME_DIR`
- `CHEZMOI_OS`
- `CHEZMOI_SOURCE_DIR`
- `CHEZMOI_VERSION`
- `CHEZMOI_WORKING_TREE`

In general, my preference would be to set these *immediately* before `realSystem.RunCommand` using `exec.Command.Env`, but that would require changes to a number of other APIs (the config is not passed to `RunCommand`) and be slower in any case as the variables would need to be set on each script execution.

This *should* work on Windows (it is guaranteed to work on any Unix-like OS), but I don't have a Windows system to verify. I'm uncertain what sort of tests could be run or updated to show this working, but "it works on my system".

### DIFF
--- a/pkg/chezmoi/realsystem.go
+++ b/pkg/chezmoi/realsystem.go
@@ -126,6 +126,7 @@ func (s *RealSystem) RunScript(scriptname RelPath, dir AbsPath, data []byte, int
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	cmd.Env = s.env
 
 	return s.RunCmd(cmd)
 }
--- a/pkg/chezmoi/realsystem_unix.go
+++ b/pkg/chezmoi/realsystem_unix.go
@@ -5,6 +5,7 @@ package chezmoi
 
 import (
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"sync"
@@ -23,6 +24,7 @@ type RealSystem struct {
 	scriptTempDir           AbsPath
 	devCache                map[AbsPath]uint // devCache maps directories to device numbers.
 	tempDirCache            map[uint]string  // tempDirCache maps device numbers to renameio temporary directories.
+	env                     []string
 }
 
 // RealSystemWithSafe sets the safe flag of the RealSystem.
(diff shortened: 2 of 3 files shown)


halostatue:
@twpayne Sorry for not opening a discussion on this first, but this went from idea to execution in a matter of a couple of hours (most of it reading code and documentation) and I thought that presenting a concrete example would be better. This 
and the use of `case $(uname -s) in; …; esac` would allow me to change all but one or two of my `run_once` scripts (which occasionally need to be run more than once because of changes outside chezmoi, but really should not be run regularly) from templated to non-templated.

twpayne:
Thanks for this - this is a very interesting idea with clear benefits.

> In general, my preference would be to set these _immediately_ before `realSystem.RunCommand` using `exec.Command.Env`, but that would require changes to a number of other APIs (the config is not passed to `RunCommand`) and be slower in any case as the variables would need to be set on each script execution.

I agree that these should be set in `exec.Command.Env`, however, I don't think this will provide a measurable performance hit.

What I would suggest is:
1. Add an `env []string` field to the `RealSystem` `struct`.
2. Add a `RealSystemWithEnv` `RealSystemOption` to set this field.
3. Modify `RealSystem.RunCommand` to set `exec.Command.Env` to the `RealSystem`'s `env` field.
4. Pass the `RealSystemWithEnv` option to the call to `NewRealSystem` in `pkg/cmd/config.go`.

I would also be tempted to set `CHEZMOI_` environment variables for all values in the `.chezmoi` template data. This will likely require some refactoring of `Config.defaultTemplateData`.

Of course we'll also need to document these environment variables too.

Let me know if you'd like to make these changes. I can do them if you prefer.

twpayne:
In fact, the ability to set arbitrary environment variables for scripts is likely a useful feature, see [this comment](https://github.com/twpayne/chezmoi/discussions/1426#discussioncomment-3804715).

halostatue:
I’ve put a commit in place that does this for `realsystem_unix.go`, but it causes a segfault on my M1 Mac:

```console
$ ./chezmoi apply
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x58 pc=0x102a7d850]

goroutine 1 [running]:
github.com/twpayne/chezmoi/v2/pkg/chezmoi.OSRelease({0x0, 0x0})
	/Users/austin/mess/2022/40/chezmoi/pkg/chezmoi/data.go:60 +0x80
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).defaultTemplateData(0x140001d8000)
	/Users/austin/mess/2022/40/chezmoi/pkg/cmd/config.go:978 +0x2fc
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).persistentPreRunRootE(0x140001d8000, 0x140002ed400, {0x4?, 0x1031a55a2?, 0x4?})
	/Users/austin/mess/2022/40/chezmoi/pkg/cmd/config.go:1664 +0x890
github.com/spf13/cobra.(*Command).execute(0x140002ed400, {0x103f930b8, 0x0, 0x0})
	/Users/austin/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:847 +0x418
github.com/spf13/cobra.(*Command).ExecuteC(0x140002ecc80)
	/Users/austin/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:990 +0x354
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/austin/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:918
github.com/twpayne/chezmoi/v2/pkg/cmd.(*Config).execute(0x1400091fde8?, {0x14000032050, 0x1, 0x1})
	/Users/austin/mess/2022/40/chezmoi/pkg/cmd/config.go:1161 +0x6c
github.com/twpayne/chezmoi/v2/pkg/cmd.runMain({{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}}, {0x14000032050, 0x1, ...})
	/Users/austin/mess/2022/40/chezmoi/pkg/cmd/cmd.go:264 +0x150
github.com/twpayne/chezmoi/v2/pkg/cmd.Main({{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}}, {0x14000032050?, 0x14000054000?, ...})
	/Users/austin/mess/2022/40/chezmoi/pkg/cmd/cmd.go:128 +0x54
main.main()
	/Users/austin/mess/2022/40/chezmoi/main.go:26 +0xd4
```

I’m not sure where to go with this, so you’re more than welcome to run with the concept—I’m a bit on the busy side and probably won’t do as good a job on this as you would as it covers far more of the codebase than I have dealt with so far.

twpayne:
Thank you for this excellent idea and initial implementation!

#2423 adds a `scriptEnv` configuration variable. I'll create a follow-up PR that sets `CHEZMOI_` environment variables for default template data.

PR #2425, opened 2022-10-06
feat: 1Password account lookup table

halostatue:
This is the long-promised draft of the 1Password account lookup table discussed in #1974.

Basic testing using `execute-template` with various values seems to work *very* well, but I have not yet modified the tests or test scripts to run this new functionality.


twpayne:
This is looking great, thank you.

I'm currently planning to complete #2429 before tagging a new version of chezmoi. Should I wait for this PR to be complete, too? As a rough guideline, I try to release no more than once every two weeks, so this PR can also go into a later version if that timing suits you better.

halostatue:
The tests were easier to update than I thought they would be, and the changes should just work on Windows, too.

twpayne:
Wonderful! Thank you very much for this fantastic contribution :)

PR #2423, opened 2022-10-05
feat: Add scriptEnv configuration variable

twpayne:
Refs #2418.
Refs https://github.com/halostatue/chezmoi/commit/fd75d6be45e3bb8b34d45a9ba60d4197b843b394.

cc @halostatue 

Co-authored-by: Austin Ziegler <austin@zieglers.ca>

PR #2429, opened 2022-10-07
feat: Expose template data in environment variables

twpayne:
Fixes #2418.
Refs #2423.

Not quite ready yet...

### DIFF
--- a/pkg/cmd/testdata/scripts/scriptenv.txtar
+++ b/pkg/cmd/testdata/scripts/scriptenv.txtar
@@ -1,15 +1,22 @@
-[!exec:python3] skip 'python3 not found in $PATH'
+[windows] skip 'UNIX only'
 
 # test that chezmoi sets environment variables for scripts
 exec chezmoi apply
-stdout VALUE
+stdout ^WORK=${WORK@R}$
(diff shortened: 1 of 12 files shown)


pkg/cmd/util.go
```diff
@@ -26,6 +26,35 @@ var (
 	}
 )
 
+// camelCaseToUpperSnakeCase converts a string in camelCase to UPPER_SNAKE_CASE.
+func camelCaseToUpperSnakeCase(s string) string {
```
halostatue, line 30: This and `upperSnakeCaseToCamelCase` might be good to expose as filter functions in templates.

halostatue (approved):
This looks very good.

pkg/cmd/testdata/scripts/scriptenv.txtar
```diff
@@ -1,15 +1,22 @@
-[!exec:python3] skip 'python3 not found in $PATH'
+[windows] skip 'UNIX only'
```
halostatue, line 1: This *should* work on Windows, per https://learn.microsoft.com/en-us/windows/win32/procthread/inheritance#inheriting-environment-variables.

PR #2436, opened 2022-10-09
ci: add workflow to push get.chezmoi.io files

### LINKED ISSUE #2430: Suggestion: `get.chezmoi.io` as an alternative to `chezmoi.io/get`
I'm following up #1816, because I believe I have found an alternative solution.

1. Create the `chezmoi/get.chezmoi.io` repository, configure it to deploy static HTML files as any normal GitHub page.
2. Configure the subdomain for it in the repo settings so it gets deployed to `get.chezmoi.io`.
3. Add the script to the repo and name it `index.html`, even though it's not a HTML.

Example:

https://github.com/felipecrs/get.felipecrs.com

```console
$ curl -fsSL get.felipecrs.com
#!/bin/sh

echo "Hello world!"

$ curl -fsSL get.felipecrs.com | sh
Hello world!
```

The only caveat is that GitHub will serve it using the wrong `Content-Type`:

```console
$ curl -fsSvL get.felipecrs.com 2>&1 | grep Content-Type
< Content-Type: text/html
```

This has no effect for the use case (`curl -fsSL get.chezmoi.io`), but do makes the page look weird if opened in the browser (https://get.felipecrs.com). I don't think it matters though.

For automating the update of the `index.html` once you change the script in the main repo, you can use GitHub Actions, and I can create the workflow for you if you like the idea.

What do you think? Feel totally free to say no, of course. :)

felipecrs:
Closes #2430

I've chosen to add this to the existing release job because:

1. I could not find where the trigger for publishing to chezmoi.io is
   - But I believe the correct trigger would be the same used to publish chezmoi.io docs
3. It will be tested first
4. It will only be updated when publishing newer versions of chezmoi

Please let me know what you think about the chosen place to add these steps.

## Notes

- The `.nojekyll` file [is required](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#:~:text=Otherwise%2C%20disable%20the%20Jekyll%20build%20process%20by%20creating%20an%20empty%20file%20called%20.nojekyll%20in%20the%20root%20of%20your%20publishing%20source%2C%20then%20follow%20your%20static%20site%20generator%27s%20instructions%20to%20build%20your%20site%20locally.).
- I'm testing this with https://github.com/felipecrs/source-to-get.felipecrs.com and https://github.com/felipecrs/get.felipecrs.com

## Additional actions

Also, one additional action is required on your end:

- Create the `GET_CHEZMOI_IO_SSH_DEPLOY_KEY` secret. It can be done as following: https://cpina.github.io/push-to-another-repository-docs/setup-using-ssh-deploy-keys.html#setup-ssh-deploy-keys


assets/get.chezmoi.io/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
```
felipecrs, line 21: I can also copy the LICENSE from the main repository during build time to the new repository if you want, to prevent having to maintain the two.

twpayne:
Thanks very much for this!

I've created the deploy key and added it to github.com/chezmoi/get.chezmoi.io. The serving of the pages from the `get.chezmoi.io` custom domain is not yet complete. I made a mistake in the process (I verified the domain with my personal GitHub account rather than the chezmoi organization account) and now need to wait ~3h for a DNS cache entry to expire.

Right now there is no action to deploy chezmoi.io on release. I removed it in 48973fd9da96fcc1b2d655c3d5632823c3b1e18d because it was failing and I haven't taken time to fix it. Looking at the commit now, I think it was failing because I need to specify `with: fetch-depth: 0` to the `actions/checkout` step. I'll restore this after this PR is merged.

assets/get.chezmoi.io/LICENSE
twpayne: That would be great, please do this.

PR #2440, opened 2022-10-10
chore: Refactor 1Password account map

twpayne:
Refs #2425. cc @halostatue

A couple of changes:
* Factor out `onepasswordAccountMap` function for testing.
* Add test for desired behavior, replacing comment.
* Simplify logic by detecting collisions after map has been built.
* Use [`strings.Cut`](https://pkg.go.dev/strings#Cut).

Please tell me if anything is missing.


halostatue (approved):
The changes look great.

PR #2441, opened 2022-10-10
docs: Use get.chezmoi.io for install links

twpayne:
Refs #2430. cc @felipecrs 

I think this is correct. We should probably add an equivalent to `get.chezmoi.io` for PowerShell users.

### DIFF
--- a/assets/chezmoi.io/docs/index.md
+++ b/assets/chezmoi.io/docs/index.md
@@ -7,7 +7,7 @@ dotfiles from your GitHub dotfiles repo on a new, empty machine with a single
 command:
 
 ```console
-$ sh -c "$(curl -fsLS https://chezmoi.io/get)" -- init --apply $GITHUB_USERNAME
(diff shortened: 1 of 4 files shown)


bradenhilton:
Is it possible to serve content dynamically based on the User Agent with GitHub pages? I tried searching but couldn't find a conclusive answer.

It would mean you could use the same link everywhere in the documentation, but it being dynamic would result in users with User Agent overrides being served incorrect content, so perhaps if we did that we should also offer `get.chezmoi.io/{ps1,sh}`?

halostatue:
@bradenhilton Unfortunately, apparently not.

felipecrs:
> We should probably add an equivalent to `get.chezmoi.io` for PowerShell users.

Hm... that's true. Well, maybe something like `get.chezmoi.io/ps1`?

felipecrs:
> Is it possible to serve content dynamically based on the User Agent with GitHub pages? I tried searching but couldn't find a conclusive answer.

It's not possible, you have no access to an HTTP based reverse proxy configuration like nginx. The only way would be to use something else, like Cloudflare or Netlify.

PR #2442, opened 2022-10-11
Add instructions for openSUSE Tumbleweed

FilippoBonazziSUSE:
Add instructions to install chezmoi in openSUSE Tumbleweed

PR #2445, opened 2022-10-12
chore: Add COMMIT file to source archives

twpayne:
As [discussed](https://github.com/twpayne/chezmoi/pull/2442#issuecomment-1275066645) om #2442.

@FilippoBonazziSUSE this means that chezmoi versions 2.25.0 and later will have a `COMMIT` file in the source archive containing the git commit hash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /.vagrant
+/COMMIT
 /bin/chezmoi
 /bin/gofumpt
 /bin/golangci-lint
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -2,6 +2,7 @@ project_name: chezmoi
 
 before:
   hooks:
+  - assets/scripts/generate-commit.sh -o COMMIT
   - go mod download all
 
 builds:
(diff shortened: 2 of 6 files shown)


FilippoBonazziSUSE:
Hi, sorry I'm getting back to you so late. I tried but couldn't find the COMMIT file. Did I misunderstand what you mean?

I looked in the .zip and .tar.gz listed here: https://github.com/twpayne/chezmoi/releases/tag/v2.25.0

Also I'm actually not sure that would be useful, because I think the OBS service that fetches `chezmoi` releases fetches the Git tag directly.

twpayne:
The `COMMIT` file in in the source archive:

```console
$ curl -LO https://github.com/twpayne/chezmoi/releases/download/v2.25.0/chezmoi-2.25.0.tar.gz
$ tar tzf chezmoi-2.25.0.tar.gz | grep COMMIT | head -n1
chezmoi-2.25.0/COMMIT
```

If the OBS service fetches the git commit (i.e. the SHA1 hash, like `b3a8879e30a20134c9bef48646055e57aa78d8c5` as well as the tag (`v2.25.0`)), that's fantastic.

PR #2452, opened 2022-10-14
feat: [WIP] Allow template delimiters to be configured

bradenhilton:
Refs: https://github.com/twpayne/chezmoi/discussions/2426

This is an incomplete proof of concept.

Currently the delims are not inherited when nesting, which they should be, based on my current understanding.

### DIFF
--- a/pkg/chezmoi/delims.go
+++ b/pkg/chezmoi/delims.go
@@ -1,0 +1,7 @@
+package chezmoi
+
+// Delims are template delimiters.
+type Delims struct {
+	Left  string `mapstructure:"left"`
+	Right string `mapstructure:"right"`
+}
--- a/pkg/chezmoi/sourcestate.go
+++ b/pkg/chezmoi/sourcestate.go
@@ -100,6 +100,7 @@ type SourceState struct {
 	templates               map[string]*template.Template
 	externals               map[RelPath]*External
 	ignoredRelPaths         map[RelPath]struct{}
+	delims                  map[string]*Delims
 }
 
 // A SourceStateOption sets an option on a source state.
@@ -123,6 +124,13 @@ func WithCacheDir(cacheDirAbsPath AbsPath) SourceStateOption {
 func WithDefaultTemplateDataFunc(defaultTemplateDataFunc func() map[string]any) SourceStateOption {
 	return func(s *SourceState) {
 		s.defaultTemplateDataFunc = defaultTemplateDataFunc
+	}
+}
+
+// WithDelims sets the template delimiters.
+func WithDelims(delims map[string]*Delims) SourceStateOption {
+	return func(s *SourceState) {
+		s.delims = delims
 	}
 }
 
@@ -651,8 +659,20 @@ func (s *SourceState) Encryption() Encryption {
 
 // ExecuteTemplateData returns the result of executing template data.
 func (s *SourceState) ExecuteTemplateData(name string, data []byte) ([]byte, error) {
+	leftDelim := "{{"
+	rightDelim := "}}"
+	extension := strings.TrimPrefix(path.Ext(strings.TrimSuffix(name, ".tmpl")), ".")
+	delims := s.delims[extension]
+	if len(delims.Left) > 0 {
+		leftDelim = delims.Left
+	}
+	if len(delims.Right) > 0 {
+		rightDelim = delims.Right
+	}
+
 	tmpl, err := template.New(name).
 		Option(s.templateOptions...).
+		Delims(leftDelim, rightDelim).
 		Funcs(s.templateFuncs).
 		Parse(string(data))
 	if err != nil {
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -79,6 +79,7 @@ type ConfigFile struct {
 	CacheDirAbsPath    chezmoi.AbsPath                 `mapstructure:"cacheDir"`
 	Color              autoBool                        `mapstructure:"color"`
 	Data               map[string]any                  `mapstructure:"data"`
+	Delims             map[string]*chezmoi.Delims      `mapstructure:"delims"`
 	Format             writeDataFormat                 `mapstructure:"format"`
 	DestDirAbsPath     chezmoi.AbsPath                 `mapstructure:"destDir"`
 	Interpreters       map[string]*chezmoi.Interpreter `mapstructure:"interpreters"`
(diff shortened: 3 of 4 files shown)


bradenhilton:
I've just added a testscript which demonstrates that the delims are configurable, and that they are currently not inherited when nesting templates.

twpayne:
I suspect that the reason that the nested templates do not inherit the delimiters is because they are parsed separately. Specifically, chezmoi loads and parses templates in `.chezmoitemplates` [here](https://github.com/twpayne/chezmoi/blob/v2.25.0/pkg/chezmoi/sourcestate.go#L1239). Note that inheriting delimiters here may not be desirable anyway as the interpretation of the template depends on which file it is used in.

pkg/chezmoi/sourcestate.go
```diff
@@ -651,8 +659,20 @@ func (s *SourceState) Encryption() Encryption {
...
+	if len(delims.Left) > 0 {
+		leftDelim = delims.Left
+	}
+	if len(delims.Right) > 0 {
+		rightDelim = delims.Right
+	}
+
 	tmpl, err := template.New(name).
 		Option(s.templateOptions...).
+		Delims(leftDelim, rightDelim).
```
twpayne, line 675: You should be able to pass `delims.Left` and `delims.Right` here, as, if either is empty, it will revert to the default (`{{` or `}}`). This will eliminate the `leftDelim` and `rightDelim` local variables completely.

pkg/chezmoi/delims.go
```diff
@@ -0,0 +1,7 @@
+package chezmoi
+
+// Delims are template delimiters.
+type Delims struct {
```
twpayne, line 4: As a nit, please rename this to use the full word `Delimiters`, similarly please replace other occurrences of "delims" with "delimiters".

twpayne:
This is looking promising :)

pkg/chezmoi/sourcestate.go
bradenhilton: I had it like that originally, but I recall it panicking when `delims` was `nil`. I'll double check and change appropriately.

halostatue:
There’s more places to consider this:

- `pkg/cmd/templatefuncs.go:141` (`includeTemplateTemplateFunc`)
- `pkg/chezmoi/sourcestate.go:1268` (`addTemplatesDir`)

With both of these items, it *may* be possible to do string processing (since this is *already* happening, after a fashion) as you had discussed (and for items in `.chezmoitemplate/`, it may be a *good idea* as it would allow for included templates to use different delimiters).

pkg/chezmoi/sourcestate.go
halostatue: I think that if you were to make it so that it’s `map[string]Delim`, it would work (I could be wrong; I’m not really a gopher). At a minimum, though, if there *are* no delimiters specified, we should always `make(map[string]*Delim)`.

halostatue:
There should be some parameters added to `chezmoi execute-template` (`--delimiters '#{{,}}'` or `--left-delimiter '#{{' --right-delimiter '}}'` for in-line testing.

bradenhilton:
@twpayne Assuming you are aware of the broader discussion in https://github.com/twpayne/chezmoi/discussions/2426, but here's a quick summary if not:

- Language formatters/linters do not play nicely with the template syntax, and not all of them have an equivalent to `# <linter>:ignore-line` etc.
  - At best, they will just log a complaint in your editor, but otherwise leave it as is.
  - At worst, they will attempt to map the template syntax to the syntax of the file, so you can end up with lines like

    ```
    { { template "foo" } }
    ```

    or even something like

    ```
    {
        {
            template "foo"
        }
    }
    ```

- You can prefix template syntax with a language-specific comment to force formatters/linters to ignore it.
  - At best, this leaves leftover comments at best (which is ugly)
  - At worst, the leftover comments will comment out code (which results in unexpected behavior).
- The current best approach seems to be something like `#{{ print "\n" }}{{ template "helpers/foo.sh" }}`, this doesn't comment out code but will leave leftover comments in the template result.

The only reason I can think of to not inherit delims at the moment is if you have a template (A) which includes data from templates (B..Z) which do not match A's extension.

If you're just writing, for example, a shell script template which includes helper functions from templates, they can all use the same delims, and so they should be inherited.

I would prefer to set delims for every template individually, but if they are inherited this doesn't seem possible.

halostatue:
> * The current best approach seems to be something like `#{{ print "\n" }}{{ template "helpers/foo.sh" }}`, this
> doesn't comment out code but will leave leftover comments in the template result.

That would be `#{{ print "\n" }}#{{ template "helpers/foo.sh" }}`.

> The only reason I can think of to not inherit delims at the moment is if you have a template (A) which includes
> data from templates (B..Z) which do not match A's extension.

So, that’s only true for templates that are handled with `ParseFiles` or `ParseGlob` (which is more or less the way that I *thought* that Chezmoi was handling this). What’s *actually* happening is that Chezmoi is parsing everything in `.chezmoitemplates/` as separate entities and storing that in a `map[string]*template.Template`. At `pkg/chezmoi/sourcestate.go:685-690`, all of those templates are added to the template being parsed with `template.AddParseTree`.

So…strictly speaking, the inheritance of delimiters is *not possible* with the way that Chezmoi works, since the delimiters are set with `Parse`, `ParseFiles`, or `ParseGlob`. This makes per-file configuration much better (and in-file configuration *possible*), but extension-based configuration (the current approach) is somewhat less useful.

`addTemplatesDir` at `pkg/chezmoi/sourcestate.go:1268` reads the template file, which means we could look for a template configuration line (and *remove* it from the generated template). Something like this:

```sh
# chezmoi:set-file-delimiters left="#{{" right="}}"
#{{ . }}
```

This should convert to:

```sh
#{{ . }}
```

It would then be executed exactly as expected.

> I would prefer to set delims for every template individually, but if they are inherited this doesn't seem possible.

I think that we can, and I think we can do it in the way that you had originally suggested—and it will be easier than a configuration-file-based approach, because of the way that chezmoi loads templates.

halostatue:
@bradenhilton I think you might find #2456 an interesting approach…

bradenhilton:
@halostatue Just FYI...

> > * The current best approach seems to be something like `#{{ print "\n" }}{{ template "helpers/foo.sh" }}`, this
> >   doesn't comment out code but will leave leftover comments in the template result.
> 
> That would be `#{{ print "\n" }}#{{ template "helpers/foo.sh" }}`.

This isn't correct. It's the current best approach outside of this PR, so only the first comment is required. I probably should have specified that.

It essentially becomes

```
#
{{ template "helpers/foo.sh" }}
```

± some whitespace.

---

+1 to your other thoughts.

Issue #2427, opened 2022-10-06
`--include` flag not working for some target types


sm1999:
## Describe the bug

I wanted to list files managed by chezmoi of target type encrypted. 
Running `chezmoi managed --include encrypted` does not yield any output.
After some testing, I found that files, dirs, scripts, remove, and symlinks target types are working when passed to the `--include` flag.
But encrypted and externals target types are not working.

 Running `chezmoi managed --include files,encrypted` works as expected. showing normal and encrypted  target types.
Also preceding target types with no works for all target types.

## To reproduce

Consider i have three files normal_file, encrypted_file, script_file.

1) run `chezmoi managed --include encrypted`
2) Output will be empty
3) run `chezmoi managed --include files,encrypted`
4) Output will contain normal_file, encrypted_fire. script_file is excluded.

## Expected behavior

`chezmoi managed --include encrypted` should show files with target type encrypted

## Output of command with the `--verbose` flag

```console
~ ❯ chezmoi managed --include=encrypted --verbose
~ ❯
(chezmoi exit with exit code 0)
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT    CHECK                MESSAGE
warning   version              dev
warning   latest-version       v2.24.0
ok        os-arch              linux/amd64 (Arch Linux)
ok        uname                Linux sm1999-arch 5.15.72-1-lts #1 SMP Wed, 05 Oct 2022 12:15:15 +0000 x86_64 GNU/Linux
ok        go-version           go1.19.2 (gc)
ok        executable           ~/dev/Github/chezmoi/chezmoi
ok        upgrade-method       replace-executable
ok        config-file          ~/.config/chezmoi/chezmoi.yaml, last modified 2022-09-03T22:49:28+05:30
warning   source-dir           ~/.local/share/chezmoi is a git working tree (dirty)
ok        suspicious-entries   no suspicious entries
warning   working-tree         ~/.local/share/chezmoi is a git working tree (dirty)
ok        dest-dir             ~ is a directory
ok        umask                022
ok        cd-command           found /bin/zsh
ok        cd-args              /bin/zsh
info      diff-command         not set
ok        edit-command         found /usr/local/bin/nvim
ok        edit-args            /usr/local/bin/nvim
ok        git-command          found /usr/bin/git, version 2.38.0
ok        merge-command        found /usr/local/bin/nvim
ok        shell-command        found /bin/zsh
ok        shell-args           /bin/zsh
[...]

PR #2455, opened 2022-10-14
feat: Extend --include and --exclude flags to include templates

twpayne:
Refs @sm1999 in https://github.com/twpayne/chezmoi/issues/2427#issuecomment-1278819861.

This allows you to run commands like:

```console
$ chezmoi managed --include=templates
```

to list all your templates.

chezmoi still needs some work here. The `--include` and `--exclude` flags mix two separate concepts: target state (e.g. `files`, `dirs`,`symlinks`, etc.) and source state (e.g. `encrypted`, `externals`, `templates`). This leads to conflicts when both are specified, for example should `--include=files --exclude=templates` include templated files or not?

Fixing this properly is probably a task for chezmoi v3.

### DIFF
--- a/assets/chezmoi.io/docs/reference/command-line-flags/common.md
+++ b/assets/chezmoi.io/docs/reference/command-line-flags/common.md
@@ -10,8 +10,8 @@ Set the output format.
 
 Only operate on target state entries of type *types*. *types* is a
 comma-separated list of target states (`all`, `dirs`, `files`, `remove`,
-`scripts`, `symlinks`) and/or source attributes (`encrypted`, `externals`) and
-can be excluded by preceding them with a `no`.
+`scripts`, `symlinks`) and/or source attributes (`encrypted`, `externals`,
+`templates`) and can be excluded by preceding them with a `no`.
 
 !!! example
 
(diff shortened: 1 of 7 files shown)


sm1999:
Thank you! I tested this pr locally. For `chezmoi managed --exclude templates` it is having the same issue as #2427. It is not excluding templates target types from the output.

> This leads to conflicts when both are specified, for example should --include=files --exclude=templates include templated files or not?

In general terms, I see that `files` is a domain and `encrypted`, `externals`, `templates` are its subdomains. so `--include=files --exclude=templates` should include files excluding templates target types and `--exclude=files --include=templates` should exclude all files irrespective of `--include` flag

twpayne:
This is tricky though. For example, `externals` includes some `files`, `dirs`, and `symlinks`, `templates` includes `files` and `symlinks`, i.e. it's not a strict hierarchy.

Also, sadly,  the order of command line flags is not preserved, so `--include=files --exclude=templates` must have the same behavior as `--exclude=templates --include=files`.

I'll merge this PR as-is and hope we can improve this in the future.

PR #2456, opened 2022-10-15
feat: In-template custom delimiters

halostatue:
This builds on #2452.

### DIFF
--- a/pkg/chezmoi/delims.go
+++ b/pkg/chezmoi/delims.go
@@ -1,0 +1,7 @@
+package chezmoi
+
+// Delims are template delimiters.
+type Delims struct {
+	Left  string `mapstructure:"left"`
+	Right string `mapstructure:"right"`
+}
--- a/pkg/chezmoi/sourcestate.go
+++ b/pkg/chezmoi/sourcestate.go
@@ -70,6 +70,20 @@ type External struct {
 	sourceAbsPath   AbsPath
 }
 
+type TemplateDelimiters struct {
+	Left, Right string
+}
+
+type TemplateDirectives struct {
+	Delimiters TemplateDelimiters
+}
+
+type ExecuteTemplateDataParams struct {
+	Name       string
+	Data       []byte
+	Directives *TemplateDirectives
+}
+
 // A SourceState is a source state.
 type SourceState struct {
 	sync.Mutex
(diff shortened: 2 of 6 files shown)


halostatue:
As a note, even though it builds on the previous PR, it *removes* `delims.sh` instead working toward the concept of `directives`. So that there aren’t any messy bits left behind, directives would be fully line-based *and removed* during Chezmoi processing. The rules would more or less be:

1. One directive to a line.
2. Directives begin with `chezmoi:`.
3. The line with the directive will be removed (which means that they can be in full-line comments).
4. There may be zero or more parameters to the directive.
5. The parsing rules for parameters would be directive specific, but *as much as possible* would work like `parameter=value` where `value` is a raw value and parameters are separated by spaces.
   - This could be problematic if you need to *include* spaces in a parameter value, but at that point, we need to bring in quoting, tokenization, etc.…
   - Delimiters are not allowed to have spaces as part of the specification, so this simplistic approach Just Works™️.

 Right now, the only directive that a template can specify is `chezmoi:set-file-delimiters`.

bradenhilton:
Directives are my preferred solution anyway.

We should probably bind them to an appropriate data structure instead of parsing strings.

An example would be [`algorand/msgp`](https://github.com/algorand/msgp/blob/26c81f33bddb2b3897a722290e67e9e363ca96d1/parse/directives.go), though I don't think ours need to be as involved.

halostatue:
> Directives are my preferred solution anyway.
> 
> We should probably bind them to an appropriate data structure instead of parsing strings.
> 
> An example would be [`algorand/msgp`](https://github.com/algorand/msgp/blob/26c81f33bddb2b3897a722290e67e9e363ca96d1/parse/directives.go), though I don't think ours need to be as involved.

Looking over the `msgp` directives, it’s not doing something that much different than what I’m doing—but it is more regular about it. What I’ve got is a fairly quick proof of concept for this *one* directive, and it puts that result into a struct inside of a  `TemplateDirectives` struct.

Some directives—like the delimiter one shown here—can only *meaningfully* be specified once‡ (or at least, multiple specifications *replace* earlier specifications, but affect the template exactly once). Others might stack. Still others might affect how the *rest* of the file is pre-parsed (and potentially be turned off). Others might be a for application *after* parsing and execution (which introduces a bunch of non-determinism, IMO).

The use of a `TemplateDirectives` struct means that there’s possibly room to grow template directives.

PR #2457, opened 2022-10-16
feat: Add per-template configurable delimiters

twpayne:
This builds on the work of @bradenhilton in #2452 and @halostatue in #2456. It is basically #2456 with some minor user-visible changes and some code refactoring.

User-visible changes:
* The directive syntax has changed slightly, it's now `chezmoi:template:left-delimiter=$LEFT right-delimiter=$RIGHT`.
* The delimiters `$LEFT` and `$RIGHT` can be quoted strings to include spaces.
* Rather than removing the whole line containing the directive from the template, only the delimiters themselves are removed, as this is minimum necessary to preserve the syntactic validity of the template. Note that there's a weird trick that you can use if you want to remove the whole line, this is described below.
* Added documentation.

Internal changes:
* The directive parsing and removal is done with a regular expression, which simplifies the logic.
* Several more tests have been added.
* A couple of things have been renamed for consistency with chezmoi's existing code.

In a future commit I make the syntax of the `chezmoi:template:` directive more flexible.

The one weird trick to remove whole lines is to use `text/template`'s existing support for comments. For example:

    # [[/* chezmoi:template:left-delimiter="# [[" right-delimiter=]] */ -]]

simultaneously sets the left delimiter to `# [[`, the right delimiter to `]]`, and evaluates to the empty string, including stripping the following whitespace.

This commit lists @halostatue as the author with @bradenhilton and @twpayne as co-authors.
cc @felipecrs as he's likely interested.

Closes #2452.
Closes #2456.


bradenhilton:
Looking good so far:

<details>
<summary><code>test.ps1.tmpl</code></summary>

```powershell
# ...

#{{ template "helpers/powershell/Read-Stdin.ps1" }}

# ...

#{{- /* chezmoi:template:left-delimiter="#{{" */}}

```

</details>

<details>
<summary><code>❯ .\chezmoi.exe -S .\source -D .\dest diff</code></summary>

```console
❯ .\chezmoi.exe -S .\source -D .\dest diff
diff --git a/test.ps1 b/test.ps1
new file mode 100666
index 0000000000000000000000000000000000000000..be0171c3c928a77175154f11a730c3cfa5dd8b82
--- /dev/null
+++ b/test.ps1
@@ -0,0 +1,19 @@
+# ...
+
+function Read-Stdin {
+    <#
+    .SYNOPSIS
+        Reads stdin to a string.
+
+    .OUTPUTS
+        System.String.
+
+        Read-Stdin returns a string containing the contents of stdin.
+    #>
+
+    [System.Console]::InputEncoding = [System.Text.UTF8Encoding]::new()
+    $stdin = [System.Console]::In
+    return $stdin.ReadToEnd()
+}
+
+# ...
```

</details>

I can test more thoroughly tomorrow.

Maybe we should add an FAQ entry about some of the collisions with the default delimiters that we have seen so far (the common `{{{`/`}}}` fold markers for vim for example) where we recommend setting custom delimiters. Doesn't necessarily need to be done in this PR though.

halostatue:
Overall, this looks good. From a "cleanliness of generated code" perspective, I would still suggest that Chezmoi template directives be line-oriented, because the "one neat trick" doesn’t work if one has to use a block comment, rather than a line comment, I think, and I think that doing

```
preamble
# [[ /* chezmoi:template:left-delimiter="# [[" right-delimiter="]]" */ ]]
body
```

Would result in:

```
preamble

body
```

Not 

```
preamble
body
```

By making in-template reconfiguration rules line-oriented and removing them, it makes it clearer (to me) that this is meta to the template itself, and it won’t matter if I comment the rule in any of the following ways:

```
# chezmoi:template:left-delimiter=[[[ right-delimiter=]]]
// chezmoi:template:left-delimiter=[[[ right-delimiter=]]]
/* chezmoi:template:left-delimiter=[[[ right-delimiter=]]] */
-- chezmoi:template:left-delimiter=[[[ right-delimiter=]]]
```

Other than this, I like the changes.

assets/chezmoi.io/docs/reference/templates/delimiters.md
```diff
@@ -0,0 +1,23 @@
...
+If a template contains the string
+
+    chezmoi:template:left-delimiter=$LEFT right-delimiter=$RIGHT
+
+Then the delimiters `$LEFT` and `$RIGHT` are used instead. chezmoi will remove
+the `$LEFT` and `$RIGHT` strings from the line in the template before template
+execution to avoid parse errors. Either or both of `left-delimiter=$LEFT` and
+`right-delimiter=$RIGHT` may be omitted. `$LEFT` and `$RIGHT` must be quoted if
+they contain spaces. If either `$LEFT` or `$RIGHT` is empty then the default
+delimiter (`{{` and `}}` respectively) is set instead.
```
halostatue, line 13: Two things worth noting:

1. Only the *first* instance of the delimiter directive means anything.
2. Because of the way that chezmoi loads and parses templates, each template's delimiters must be specified individually. They are *not* inherited.

halostatue (approved):

bradenhilton:
@halostatue

> I think that doing
> 
> ```
> preamble
> # [[ /* chezmoi:template:left-delimiter="# [[" right-delimiter="]]" */ ]]
> body
> ```
> 
> Would result in:
> 
> ```
> preamble
> 
> body
> ```
> 
> Not
> 
> ```
> preamble
> body
> ```

You can just use the white space controls to get the output you expect:

```
preamble
# [[- /* chezmoi:template:left-delimiter="# [[" right-delimiter="]]" */]]
body
```

Or just put it on the last line in every file, I think that's what I'm going to do.

> By making in-template reconfiguration rules line-oriented and removing them, it makes it clearer (to me) that this is meta to the template itself, and it won’t matter if I comment the rule in any of the following ways:
> 
> ```
> # chezmoi:template:left-delimiter=[[[ right-delimiter=]]]
> // chezmoi:template:left-delimiter=[[[ right-delimiter=]]]
> /* chezmoi:template:left-delimiter=[[[ right-delimiter=]]] */
> -- chezmoi:template:left-delimiter=[[[ right-delimiter=]]]
> ```
> 
> Other than this, I like the changes.

I think I'd prefer this, but I'm also quite happy with the current implementation.

assets/chezmoi.io/docs/reference/templates/directives.md
```diff
@@ -0,0 +1,46 @@
...
+
+    chezmoi:template:left-delimiter=$LEFT right-delimiter=$RIGHT
+
+Then the delimiters `$LEFT` and `$RIGHT` are used instead. Either or both of
+`left-delimiter=$LEFT` and `right-delimiter=$RIGHT` may be omitted. `$LEFT` and
+`$RIGHT` must be quoted if they contain spaces. If either `$LEFT` or `$RIGHT`
+is empty then the default delimiter (`{{` and `}}` respectively) is set
+instead.
+
+chezmoi will remove the line containing `chezmoi:template:` directives to
```
halostatue, line 16: This paragraph mostly belongs just below `# Directives`, removing the `to avoid parse errors from the delimiters`. That’s one reason, but the other would be `to avoid affecting the syntax of the resulting template`.

halostatue (approved):
Looks great and adding the `missing-key` is a good second directive.

assets/chezmoi.io/docs/reference/templates/delimiters.md
twpayne: Thanks, done.

assets/chezmoi.io/docs/reference/templates/directives.md
twpayne: Thanks, done.

assets/chezmoi.io/docs/reference/templates/directives.md
```diff
@@ -0,0 +1,46 @@
...
+that is not present in the map. This behavior can be changed globally with the
+`template.options` configuration variable or with a template directive:
+
+    chezmoi:template:missing-key=$VALUE
+
+`$VALUE` can be one of:
+
+| Value     | Effect                                                                                         |
+| --------- | ---------------------------------------------------------------------------------------------- |
+| `error`   | Return an error on any missing key.                                                            |
```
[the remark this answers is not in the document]
twpayne: Thanks, done.

halostatue (approved):

halostatue:
This is spectacular. Thank you @bradenhilton for your initial implementation and idea.

bradenhilton:
@twpayne Directives are supposed to work on templates within `.chezmoitemplates`, correct? It's the one thing I neglected to test.

twpayne:
Yes, they should work in templates within `.chezmoitemplates`. I also forgot to add a test, but the directive parsing happens at a low level so it should Just Work(tm). If you'd like to add a test for this in https://github.com/twpayne/chezmoi/blob/master/pkg/cmd/testdata/scripts/templatedirectives.txtar, that would be awesome.

bradenhilton:
Thought so!

I don't think they work at the moment. I'll open a new issue with details.

halostatue:
They don’t work at the moment. I can probably get a PR up by late this evening or tomorrow, but the use of the template data options parsing should definitely be added to:

- pkg/cmd/templatefuncs.go:161 
- pkg/chezmoi/sourcestate.go:1268


It should probably be added to:

- pkg/cmd/config.go:703
- pkg/cmd/config.go:1288

twpayne:
> I can probably get a PR up by late this evening or tomorrow

No stress on this! I try to keep the chezmoi release frequency no higher than once every two weeks, and 2.25.0 was released just a few of days ago, so there's no rush here.

Issue #2446, opened 2022-10-12
How to disable encryption on existing files?


raakasf:
## What exactly are you trying to do?

I had enabled (age) encryption on ~/.zshrc as I had some secrets in there. Now, that I have taken out the secrets, I want to disable encryption on that file.

## What have you tried so far?

Checked for an option in cli to disable encryption

## Where else have you checked for solutions?

* [x] I have read [chezmoi's user guide](https://chezmoi.io/user-guide/), and not found the answer. - THIS LINK IS INVALID (404)
* [x] I have searched [chezmoi's reference guide](https://chezmoi.io/reference/), and not found the answer.
* [x] Other, please give details. - Search

## Output of any commands you've tried with `--verbose` flag

```console
$ chezmoi encrypt -h
```

## Output of `chezmoi doctor`

```text
❯ chezmoi doctor                                                                                                                                                                      ─╯
RESULT    CHECK                MESSAGE
ok        version              v2.23.0, commit 7481b57f40c89a777e1f13be3e1d418a8b4f5a16, built at 2022-09-19T00:15:13Z, built by Homebrew
warning   latest-version       v2.24.0
ok        os-arch              linux/amd64 (CentOS Linux 7 (Core))
ok        uname                Linux 7d53da7a4e3f 5.15.0-48-generic #54-Ubuntu SMP Fri Aug 26 13:26:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
ok        go-version           go1.19.1 (gc)
ok        executable           /home/linuxbrew/.linuxbrew/Cellar/chezmoi/2.23.0/bin/chezmoi
ok        upgrade-method       brew-upgrade
ok        config-file          ~/.config/chezmoi/chezmoi.toml, last modified 2022-08-16T19:38:55Z
ok        source-dir           ~/.local/share/chezmoi is a git working tree (clean)
warning   suspicious-entries   ~/.local/share/chezmoi/dot_config/chezmoi/chezmoi.toml
ok        working-tree         ~/.local/share/chezmoi is a git working tree (clean)
ok        dest-dir             ~ is a directory
ok        umask                022
ok        cd-command           found /home/linuxbrew/.linuxbrew/bin/zsh
ok        cd-args              /home/linuxbrew/.linuxbrew/bin/zsh
info      diff-command         not set
ok        edit-command         found /usr/bin/vi
ok        edit-args            /usr/bin/vi
ok        git-command          found /usr/bin/git, version 2.37.3
[...]

PR #2464, opened 2022-10-17
chore: Fix user guide link in support issue template

bradenhilton:
I first noticed this when reading https://github.com/twpayne/chezmoi/issues/2446.

An alternative solution would be to add an index page to the user guide section.


twpayne (approved):
LGTM. We can add a contents page for the user guide later. Please merge :)

PR #2473, opened 2022-10-19
feat: Add option to exclude scripts that are always run

twpayne:
As requested by @felipecrs and @Lockszmith-GH in #1666.

With this, you can put

```toml
[diff]
    exclude = ["always"]
```

to exclude scripts that are always run from the output of `chezmoi diff`.

This PR also includes a re-work of the `--include` and `--exclude` options thanks to the discussion with @sm1999 in #2455.

### DIFF
--- a/assets/chezmoi.io/docs/reference/command-line-flags/common.md
+++ b/assets/chezmoi.io/docs/reference/command-line-flags/common.md
@@ -8,15 +8,29 @@ Set the output format.
 
 ## `-i`, `--include` *types*
 
-Only operate on target state entries of type *types*. *types* is a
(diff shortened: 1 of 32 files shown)


sm1999:
@twpayne I tried the example in the docs `--include=all,noencrypted` and it didn't work as expected.
Command output:

```
❯ ./chezmoi managed --include all,noencrypted ~/.config/rclone
.config/rclone
.config/rclone/rclone.conf
❯ ./chezmoi managed --include encrypted ~/.config/rclone
.config/rclone/rclone.conf
❯ ./chezmoi managed --exclude encrypted ~/.config/rclone
.config/rclone
```
Also running `--exclude externals` is excluding all types.

PR #2493, opened 2022-10-26
docs(bitwarden): Correct bitwardenFields example

choznerol:
The example is about `token.value`, so `token` should be part of the `fields` of `bw get` output.


twpayne:
Thank you!

Issue #2498, opened 2022-10-26
Newline is not removed with value coming from 1password


alexsanduk:
## Describe the bug

When value is rendered from 1password newline is not removed. It is important if you try to construct url based on `username`, `password` etc

## To reproduce

```console
$ chezmoi execute-template '{{- onepasswordRead "op://Personal/Postgres/username" -}}:{{- onepasswordRead "op://Personal/Postgres/password" -}}'
test_user
:test_password
```

## Expected behavior

Newline is removed since `{{-` and `-}}` are used.
```console
$ chezmoi execute-template '{{- onepasswordRead "op://Personal/Postgres/username" -}}:{{- onepasswordRead "op://Personal/Postgres/password" -}}'
test_user:test_password
```

## Output of command with the `--verbose` flag

```console
$ chezmoi execute-template --verbose '{{- onepasswordRead "op://Personal/Postgres/username" -}}:{{- onepasswordRead "op://Personal/Postgres/password" -}}'
test_user
:test_password
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT    CHECK                MESSAGE
ok        version              v2.24.0, commit 72d9846a7ae51fd3398727d48815fc2f13a681f9, built at 2022-09-25T19:19:53Z, built by goreleaser
warning   latest-version       v2.25.0
ok        os-arch              darwin/amd64
ok        uname                Darwin MyMac 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64 x86_64
ok        go-version           go1.19.1 (gc)
ok        executable           ~/.local/user/bin/chezmoi
ok        upgrade-method       replace-executable
ok        config-file          no config file found
ok        source-dir           ~/.local/share/chezmoi is a git working tree (clean)
ok        suspicious-entries   no suspicious entries
ok        working-tree         ~/.local/share/chezmoi is a git working tree (clean)
ok        dest-dir             ~ is a directory
ok        umask                022
ok        cd-command           found /bin/zsh
ok        cd-args              /bin/zsh
info      diff-command         not set
ok        edit-command         found ~/.local/user/nvim-macos/bin/nvim
ok        edit-args            ~/.local/user/nvim-macos/bin/nvim
ok        git-command          found /usr/bin/git, version 2.37.0
ok        merge-command        found /usr/bin/vimdiff
[...]

PR #2499, opened 2022-10-27
Make `onepasswordRead` respect the passed account

halostatue:
Indirectly, the lack of setting `onepasswordArgs.account` in `onepasswordReadTemplateFunc` was causing 1Password to prompt me for an account, even though it's provided.

The selection is prompted by the `withSessionToken` handling, which depends on `onepasswordAccount.account`.

Discovered while investigating #2498.


twpayne:
Awesome, thank you!

PR #2513, opened 2022-10-30
chore: Reorder eqFold template function reference page

bradenhilton:
It appears we missed this in both https://github.com/twpayne/chezmoi/commit/2c440aba38f10c04af548da3c7ae263a2c411c6f and https://github.com/twpayne/chezmoi/commit/520e4a6d0663aeb6eeafb4352906fdffecbd6702.


twpayne (approved):
Good catch, thanks!

Issue #2526, opened 2022-10-31
Allow to use templates to process modify_ files


felipecrs:
## Is your feature request related to a problem? Please describe.

Chezmoi's template functions are quite powerful, which includes the likes of `fromYaml`, `toYaml`, `fromJson`, `toJson` and so on. I believe these functions are powerful enough to replace `jq` for example in my `modify_` files, and if I could stop using interpreters as well (like Bash or Python) it would be even better for portability.

## Describe the solution you'd like

Somehow to instruct `chezmoi` that it should use go-template to process the data, by providing the original source file as a top level key, like:

```console
# Note that the file does not end with `.tmpl`
$ cat dot_docker/modify_config.json
{{- /* chezmoi:modify-with-template=true */ -}}
{{- $input := fromJson .input -}}
{{- $_ := set $input "myFlag" true -}}
{{- toJson $input -}}
```

- `{{- /* chezmoi:process-as-template=true */ -}}` is used to instruct chezmoi that it should process the file using template.
- chezmoi injects the original state of the file as `.input`

## Describe alternatives you've considered

I was thinking about using `chezmoi execute-template` as the shebang, but I then I thought: how would I handle the unpresence of `chezmoi` in `PATH`? Which may happen, like in the first installation.

I could probably use `{{ .chezmoi.executable }}`, but then I would have to mark the file as a template because of this, thus having to escape all other template instructions inside of the file as well.

## Additional context

None.

bradenhilton:
@felipecrs I don't have any thoughts on this issue at the moment. I just wanted to say I linked to this issue in https://github.com/twpayne/chezmoi/pull/2527 by mistake, though you might have assumed as much. Regardless, you can ignore it :)

bradenhilton:
Actually, +1 now that I'm reading through this.

It's especially painful with PowerShell, where it can take several hundred milliseconds just to launch, which is about 3-4 times slower than just running a simple `chezmoi execute-template ...`:

(based on https://devblogs.microsoft.com/powershell/optimizing-your-profile/#getting-a-baseline-for-pwsh)

`benchmark.ps1`:

```powershell
@(
    'powershell -noninteractive -noprofile -command "1"',
    'pwsh -noninteractive -noprofile -command "1"',
    'chezmoi execute-template ''{{1}}'''
) | ForEach-Object {
    $command = ($_ -split ' ')[0]
    $arg = ($_ -split ' ') | Select-Object -Skip 1

    $average = 0

[...]

Issue #2525, opened 2022-10-31
Several website URLs appear to be broken


bradenhilton:
## Describe the bug

The following URLs show `404 - Not found`:

1. https://www.chezmoi.io/ - The main URL
2. https://www.chezmoi.io/index.md - From `Home` in the navigation
3. https://www.chezmoi.io/reference/configuration-file/variables.md - `Reference -> Configuration File -> Variables`
4. https://www.chezmoi.io/reference/release-history.md - `Reference -> Release history`
5. https://www.chezmoi.io/links/articles-podcasts-and-videos.md - From `Links` in the navigation

## To reproduce

Visit any of the URLs above.

## Expected behavior

The URLs should not show a 404 error.

## Output of command with the `--verbose` flag

NA.

## Output of `chezmoi doctor`

NA.

## Additional context

NA.

twpayne:
Thanks for reporting this! I think the `deploy-website` step of the release process failed. I'll investigate.

In the short term I've run `cd assets/chezmoi.io && mkdocs gh-deploy` to redeploy the website, and it seems to be working again now.

bradenhilton:
Just refreshed and can confirm everything looks okay now.

bradenhilton:
https://github.com/twpayne/chezmoi/commit/d6da479e158236e14c584f5a0e915c70bf3c8370 "updated" `cpina/github-action-push-to-another-repository` to commit [`940a285`](https://github.com/cpina/github-action-push-to-another-repository/commit/940a2857e598a6392bd336330b07416c1ae8ea1f), which is from 2021-04-29T21:45:48Z.

Before this, the `build-website` job for `v2.25.0` used https://github.com/cpina/github-action-push-to-another-repository/commit/9e487f29582587eeb4837c0552c886bb0644b6b9, which is the latest commit at time of posting.

bradenhilton:
Here is the comparison: https://github.com/cpina/github-action-push-to-another-repository/compare/940a2857e598a6392bd336330b07416c1ae8ea1f...9e487f29582587eeb4837c0552c886bb0644b6b9

twpayne:
Thanks for the investigation. The failure for https://chezmoi.io occurs in the `push-chezmoi.io` step ([logs](https://github.com/twpayne/chezmoi/actions/runs/3361140865/jobs/5571946507)). A few things are going wrong here:
1. chezmoi uses Go templates to generate several Markown pages for the website, which uses MkDocs.
[...]

PR #2527, opened 2022-10-31
fix: Revert cpina/github-action-push-to-another-repository version bump

bradenhilton:
Partial fix for https://github.com/twpayne/chezmoi/issues/2526

This reverts the version "bump" done in https://github.com/twpayne/chezmoi/commit/d6da479e158236e14c584f5a0e915c70bf3c8370.

We may have to deal with the bot attempting to bump it again, should be as simple as telling it to stop in a comment if it opens another PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -456,7 +456,7 @@ jobs:
         cp assets/scripts/install.ps1 assets/get.chezmoi.io/ps1
         cp LICENSE assets/get.chezmoi.io/LICENSE
     - name: push-get.chezmoi.io
-      uses: cpina/github-action-push-to-another-repository@940a2857e598a6392bd336330b07416c1ae8ea1f
+      uses: cpina/github-action-push-to-another-repository@9e487f29582587eeb4837c0552c886bb0644b6b9
       env:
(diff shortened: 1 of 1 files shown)


twpayne:
I'd rather update the release process to work with the latest version.

bradenhilton:
It was already on the latest version. The bot bumped it to a version from over a year ago. Please refer to https://github.com/twpayne/chezmoi/issues/2525#issuecomment-1297274616 and https://github.com/twpayne/chezmoi/issues/2525#issuecomment-1297277844.

twpayne:
Ah, my bad. Sorry.

twpayne (approved):

twpayne:
We can configure Dependabot to ignore this dependency with an [`ignore` stanza](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) in `.github/dependabot.yml`.

bradenhilton:
Would it not be better to assume this is a one off (and tell the bot to ignore the commit) until we see otherwise?

twpayne:
> Would it not be better to assume this is a one off (and tell the bot to ignore the commit) until we see otherwise?

I suppose so. We need to remember to be vigilant the next time Dependabot runs.

PR #2529, opened 2022-10-31
feat: Allow modify_ scripts to be executed as templates

twpayne:
As suggested by @felipecrs in #2526.

I'll add a template function to make the modify JSON/YAML case easier to use. The `set` function in masterminds/sprig is, sadly, like most functions in masterminds/sprig, badly designed.

PR #2530, opened 2022-10-31
feat: Add dictSet template function

twpayne:
Refs #2526 and #2529.

This makes @felipecrs's example in https://github.com/twpayne/chezmoi/issues/2526#issue-1430089869 simpler:

```
{{ fromJson .chezmoi.stdin | dictSet "myFlag" true | toJson }}
```

Note that I still need to update #2529 to use `.chezmoi.stdin` instead of `.`, as suggested by @bradenhilton.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/dictSet.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/dictSet.md
@@ -1,0 +1,13 @@
+# `dictSet` *key*... *value* *dict*
+
+`dictSet` modifies *dict* to set *key* to *value* and returns *dict*. Multiple
+*key*s may be specified, in which case nested values are set. It is an
+alternative to [sprig's `set`
+function](http://masterminds.github.io/sprig/dicts.html) with a different
+argument order that supports pipelining.
+
+!!! example
+
+    ```
+    {{ dict | dictSet "key1" "value1" | dictSet "key2" "nestedKey" "value2" | toJson }}
+    ```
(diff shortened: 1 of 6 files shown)


felipecrs:
This is much better!

BTW, Helmfile has a function named `setValueAtPath` which works similarly, except that it allows to provide a "path" to keys.

- https://helmfile.readthedocs.io/en/latest/templating_funcs/#setvalueatpath
- https://helmfile.readthedocs.io/en/latest/#values-files-templates:~:text=Suppose%20values.yaml.gotmpl%20was%20something%20like%3A
- https://github.com/helmfile/helmfile/blob/3ac0cee50ad7fee8213e351df52cf429d71f1a8f/pkg/tmpl/context_funcs.go#L298

```
{{ fromJson .chezmoi.stdin | setValueAtPath "features.buildkit" true | toJson }}
```

twpayne:
> BTW, Helmfile has a function named `setValueAtPath` which works similarly, except that it allows to provide a "path" to keys.

Ooh, that's much nicer. Let's use that instead. PR coming up.

bradenhilton:
Isn't it possible to have a key which contains `.`? What would happen in that case? Would we need to use something like `key1.'key2.containingdot'.key3`?

felipecrs:
I'm not sure how they are handled in Helmfile, but we have a few options:

1. Allow to espace the dot, like: `"key\.with\.dots"`
2. Or simply tell the user that key with dots aren't supported by the function.

bradenhilton:
Generally I do prefer dot notation, but nesting with individual strings seems like it would be the most versatile, no? You only need to worry about escaping quotes with those.

twpayne:
Closing in favor of #2542.

PR #2531, opened 2022-10-31
feat: Add nativeLineEndings template function

twpayne:
Refs https://github.com/twpayne/chezmoi/pull/2529#issuecomment-1297714446.

### DIFF
--- a/pkg/cmd/templatefuncs.go
+++ b/pkg/cmd/templatefuncs.go
@@ -28,9 +28,14 @@ type ioregData struct {
 	value map[string]any
 }
 
-// needsQuoteRx matches any string that contains non-printable characters,
-// double quotes, or a backslash.
-var needsQuoteRx = regexp.MustCompile(`[^\x21\x23-\x5b\x5d-\x7e]`)
+var (
+	// lineEndingRx matches line endings.
+	lineEndingRx = regexp.MustCompile(`(?m)(?:\r\n|\r|\n)`)
+
+	// needsQuoteRx matches any string that contains non-printable characters,
+	// double quotes, or a backslash.
+	needsQuoteRx = regexp.MustCompile(`[^\x21\x23-\x5b\x5d-\x7e]`)
+)
 
 func (c *Config) commentTemplateFunc(prefix, s string) string {
 	type stateType int
@@ -221,6 +226,10 @@ func (c *Config) mozillaInstallHashTemplateFunc(path string) string {
 		panic(err)
 	}
 	return mozillaInstallHash
+}
+
+func (c *Config) nativeLineEndings(s string) string {
+	return lineEndingRx.ReplaceAllString(s, nativeLineEnding)
 }
 
 func (c *Config) outputTemplateFunc(name string, args ...string) string {
--- a/assets/chezmoi.io/docs/reference/templates/functions/nativeLineEndings.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/nativeLineEndings.md
@@ -1,0 +1,5 @@
+# `nativeLineEndings` *text*
+
+`nativeLineEndings` returns *text* with all line endings replaced with native
+line endings. On Windows systems this means that `\n` is replaced with `\r\n`
+and on non-Windows systems `\r\n` is replaced with `\n`.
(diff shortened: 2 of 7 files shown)


pkg/cmd/templatefuncs.go
```diff
@@ -28,9 +28,14 @@ type ioregData struct {
 	value map[string]any
 }
 
-// needsQuoteRx matches any string that contains non-printable characters,
-// double quotes, or a backslash.
-var needsQuoteRx = regexp.MustCompile(`[^\x21\x23-\x5b\x5d-\x7e]`)
+var (
+	// lineEndingRx matches line endings.
+	lineEndingRx = regexp.MustCompile(`(?m)(?:\r?\n)`)
```
halostatue, line 33: This should *probably* be something like `(?m)(?:\r|\r\n|\n)$` for the rare occasion that old MacOS files might be encountered (`\r` instead of either `\n` or `\r\n`).

But this will probably work regardless.

halostatue (approved):

pkg/cmd/templatefuncs.go
twpayne: Oh yeah, wow, you're right! PR updated.

pkg/cmd/templatefuncs.go
```diff
@@ -28,9 +28,14 @@ type ioregData struct {
 	value map[string]any
 }
 
-// needsQuoteRx matches any string that contains non-printable characters,
-// double quotes, or a backslash.
-var needsQuoteRx = regexp.MustCompile(`[^\x21\x23-\x5b\x5d-\x7e]`)
+var (
+	// lineEndingRx matches line endings.
+	lineEndingRx = regexp.MustCompile(`(?m)(?:\r\n|\r|\n)`)
```
halostatue, line 33: I *think* that `$` at the end of this would be advised, because `(?m)` makes it so that `$` matches end of line, not just end of text (for which one should always be using `\z`). It’s worth a local test at least.

halostatue (approved):

pkg/cmd/templatefuncs.go
twpayne: I did a quick test with this, changing this line to:

```go
	lineEndingRx = regexp.MustCompile(`(?m)$(?:\r\n|\r|\n)`)
```

(note the `$` in the middle)

and also with

```go
	lineEndingRx = regexp.MustCompile(`(?m)(?:\r\n|\r|\n)$`)
```

(note the `$` at the end)

Both approaches failed the existing tests. I think this is because the zero-width look-ahead `$` only knows about UNIX and Windows line endings, and so misbehaves when it encounters the old macOS `\r`.

Maybe we need to do something else?

bradenhilton: Could you not use `(?m)(?:\r\n$|\r$|\n$)`?

bradenhilton:
I did think about adding support for this with a directive in https://github.com/twpayne/chezmoi/discussions/2426

> ... Maybe a method for forcing a particular encoding or line ending? Not sure.

but I think a template function would be more versatile.

pkg/cmd/templatefuncs.go
halostatue: I suspect that would fail as well. It’s been a *long* time since I’ve had to do line-ending checks in RE, and the Go RE engine (RE2) is subtly and intentionally different than PCRE, Oniguruma, or the Vim RE, so my guess is that you can *either* match `\n` *or* `$`, not both.

halostatue:
> I did think about adding support for this with a directive in #2426
> 
> > ... Maybe a method for forcing a particular encoding or line ending? Not sure.
> 
> but I think a template function would be more versatile.

Why not both? I can see them having different purposes.

In the context of what this is being developed for (allowing `modify_` scripts to be templates), this would permit the modification of an existing file via a template and conversion as part of the template processing. This could also be used *inside* a file to ensure that there no mixed line endings. As a *directive*, it could ensure that a generated file is converted to native line endings *after* execution.

twpayne:
Would the template directive look something like:

    :chezmoi:template:line-endings={unchanged,unix,windows,macos}

where only one of `unchanged`, `unix`, `windows`, or `macos` is specified with the default being `unchanged`?

This, for sure, would make it much easier to make a cross-platform template than having to filter all the output though a `nativeLineEndings` function.

halostatue:
I hadn’t thought through the specifics, but yes I think that’s a good design.

bradenhilton:
@twpayne Looks reasonable, maybe also allow the raw values ("\r\n", "\r", "\n")?

twpayne:
Closing in favor of #2540.

Issue #2536, opened 2022-11-01
`chezmoi merge-all` reports issues with a file that does not exist


halostatue:
## Describe the bug

It looks like `run_once`, and `before` flags are being removed from files in `.chezmoiscripts/` during `merge-all`. 

## To reproduce

Not *entirely* clear.

```console
$ chezmoi merge-all
chezmoi: warning: config file template has changed, run chezmoi init to regenerate config file
chezmoi: .chezmoiscripts/1-prepare-system.sh: not a file
```

Steps to reproduce the behaviour.

```
.chezmoiscripts/run_fisher_update.sh
.chezmoiscripts/run_once_after_configure-apps.sh.tmpl
.chezmoiscripts/run_once_before_1-prepare-system.sh
.chezmoiscripts/run_once_before_2-1-configure-system.sh
.chezmoiscripts/run_once_before_2-2-configure-touchid-sudo.sh
.chezmoiscripts/run_once_before_3-1-install-macports-packages.sh.tmpl
.chezmoiscripts/run_once_before_3-2-install-homebrew-packages.sh.tmpl
.chezmoiscripts/run_once_before_3-3-install-rust-packages.sh
.chezmoiscripts/run_once_before_4-install-vscode-extensions.sh
.chezmoiscripts/run_once_before_5-build-apps.sh
```

## Expected behavior

Success with no errors.

## Output of command with the `--verbose` flag

```console
$ chezmoi --verbose merge-all
chezmoi: warning: config file template has changed, run chezmoi init to regenerate config file
chezmoi: .chezmoiscripts/1-prepare-system.sh: not a file
```

### Output of command with `--debug` flag

Abbreviated.

<details>

```console
$ console chezmoi merge-all --debug
…
2022-10-31T22:26:46-04:00 INF ReadFile component=system data="#!/usr/bin/env bash\n\nset -euo pipefail\n\necho run_once_before_1-p..." name=/Users/austin/.local/share/chezmoi/.chezmoiscripts/run_once_before_1-prepare-system.sh size=692
2022-10-31T22:26:46-04:00 INF Get bucket=scriptState component=persistentState key=1b4607ea1c51957fe678460614829aafce7c03fcf5ea8422b3479c37515c1a77 value=
2022-10-31T22:26:46-04:00 ERR Lstat error="lstat /Users/austin/.chezmoiscripts/1-prepare-system.sh: no such file or directory" component=system name=/Users/austin/.chezmoiscripts/1-prepare-system.sh
[...]

PR #2544, opened 2022-11-02
feat: Add --script-contents flag to diff command

twpayne:
As requested by @halostatue in https://github.com/twpayne/chezmoi/issues/2536#issuecomment-1299402155.

You can set `diff.scriptContents` to `false` in your configuration file.


halostatue (approved):

halostatue:
Not quite the output that I would have wanted, but *simpler* than what I would have wanted, so an absolute win. I think I’ll enjoy that when it’s released.

PR #2552, opened 2022-11-04
docs: Fix typos in Sourcehut repo patterns

kjerstadius:
I was a bit confused with the location of the `~` in the pattern and, sure enough, it didn't work when entered as documented. I also removed an incorrect `/`.


halostatue:
Thank you very much @kjerstadius!

PR #2566, opened 2022-11-11
chore: Make generation of versioninfo.json work without tags or git repo

twpayne:
Previously, generating `versioninfo.json` required `git describe --abbrev=0 --tags` to succeed, which is not the case when chezmoi is forked (https://github.com/deivid-rodriguez/chezmoi/pull/2) or if chezmoi is built from a source archive (where there is no git repo). With this commit, `versioninfo.json` is generated in both cases, albeit without a correct version or commit hash. This allows the `test-release` job to succeed on forks.

cc @deivid-rodriguez

### DIFF
--- a/assets/templates/versioninfo.json.tmpl
+++ b/assets/templates/versioninfo.json.tmpl
@@ -1,34 +1,35 @@
 {{- $name := "chezmoi" -}}
 {{- $filename := printf "%s.exe" $name -}}
 
-{{- /* Get the current year using the Magical Reference Date format */ -}}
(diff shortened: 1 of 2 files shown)


twpayne:
Hm, well it looks like deivid-rodriguez deleted his fork, but this is still a bug fix :)

deivid-rodriguez:
Yes, it was just for testing dependabot PR generation, but thanks for caring and fixing these things!

twpayne:
Thank you for your work on dependabot :)

Issue #2570, opened 2022-11-15
Encrypt filename.


memchr:
## Is your feature request related to a problem? Please describe.

Currently, chezmoi only encrypt content of files but not filenames, which may contain sensitive or exploitable data. It would be great if chezmoi cloud encrypt them.

### Solution # 1
Encrypt the filename of encrypted file and store it in as ASCII-only "armored" encoding.

e.g.
Instead of plain text filename:
```python
private_world_domination/encrypted_antimatter_bomb_schema.json.age
```
Use
```python
private_world_domination/encrypted_YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0IFgyNTUxOSA.age
```

name of directory can also be encrypted. Instead of:
```python
private_overthrow_galactic_empire/plan_of_republic.md
```
use
```python
encrypted_private_YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0I/encrypted_YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0IFgyNTUxOSBP.age
```
### Solution # 2
Use salted hash as name for encrypted file and folder. And store hash to filename table in a separated encrypted database.

e.g.
For file to be encrypted: `private_world_domination/antimatter_bomb_schema.json`

- `f8b1e5a20f9f74cfc234967a944e991562394e7acd7117d4b5ea93431bf0157a` is hash of 
       `world_domaination` + `Yw2dlLx+1/fe` (salt)
- `8c76e35b69ff41c8c12937a8fa13d225abff83e0e2ef0bc5a0bfd9afcee771b0` is hash of
       `antimatter_bomb_schema.json` + `V1Z0INLxmmNx` (salt)

Content of hash to filename table `filenames.json`:
```json
{
  "f8b1e5a20f9f74cfc234967a944e991562394e7acd7117d4b5ea93431bf0157a": "world_domination",
  "8c76e35b69ff41c8c12937a8fa13d225abff83e0e2ef0bc5a0bfd9afcee771b0": "antimatter_bomb_schema.json"
}
```

Encrypt `antimatter_bomb_schema.json` and store it in `encrypted_f8b1e5a20f9f74cfc234967a944e991562394e7acd7117d4b5ea93431bf0157a/encrypted_8c76e35b69ff41c8c12937a8fa13d225abff83e0e2ef0bc5a0bfd9afcee771b0.age`

Encrypt `filenames.json` and store in `filename.json.age`

twpayne:
[...]

PR #2572, opened 2022-11-15
feat: Allow encrypted filenames in source state

twpayne:
Refs #2570.

Work in progress.

cc @memchr. Please wait for follow-up comments.


twpayne:
This is the first step to implementing solution 1 as proposed by @memchr in #2570. It's a work in progress.

Right now, this PR refactors the code to make the full `Encryption` interface available to the file and directory attribute parsers. With this information available at this level, it should be possible to implement encrypted filenames.

Some currently unresolved questions and general questions:

1. This is going to make many file and directory names unreadable by humans (bad) while also failing to conceal the structure of files and directories with encrypted filenames (also bad). Is this sufficient? If you want to conceal the structure of a directory you can already use an encrypted external archive.
2. How do you determine which filenames are encrypted? By passing the encryption mechanism down to the the attribute parser, this PR should be possible to decrypt everything between the `encrypted_` prefix and the `.asc` suffix. If this decrypts without error then we can continue with the decrypted attributes, otherwise we should probably revert to parsing the filename as chezmoi currently does.
3. How should we encode the encrypted data in the filename? We need something that is at least compatible with existing filesystems (i.e. not containing any `\0`, `/`, `C:`, `/` strings, etc.) while also as short as possible. One of the existing BASE64 encodings probably already covers this case.
4. How should the user interface look? In the short term `chezmoi add --encrypt --encrypt-filename ~/.foo` is probably reasonable for advanced users, but how do we cope with reverse mappings when the user runs `chezmoi diff ~/.foo`?

memchr:
Thank you for your work!

1. There are two scenarios where encrypted filenames make sense. First is that
   some programs put information such as username, email, phone number in the
   filenames and this could be a privacy concern. Moreover, some filenames
   could possibly be linked to a software (e.g. tor) that is a subject
   of cernsorship in some places. Encrypted filenames should be sufficient for
   the first case.
2. How about to use a new prefix, like `encryptedfilename_` to determine
   which filenames are encrypted?
3. [base64url](https://datatracker.ietf.org/doc/html/rfc4648#section-5) should
   cover this.
4. I have no ideas yet.

twpayne:
Ah, there might be a problem.

I've updated this PR to include support for an `encryptname_` suffix that stores the rest of the filename as a base64 URL-encoding string, but tests fail with the error:

```console
$ go test ./pkg/cmd -run=TestScript/encryptedname
{"level":"info","path":"/Users/twp/bin/age-keygen","args":["age-keygen","--output","/private/var/folders/q8/1c3_5w_54sl_q8bcfxx__3lw0000gn/T/go-test-script1860109437/script-encryptedname/home/user/key.txt"],"combinedOutput":"Public key: age1lr7lacrj2kxk6xf8epf8u3mlcf4s5k6lcdwdk42uqylanr7d...","duration":"9.520209ms","size":75,"time":"2022-11-18T20:03:08+01:00","message":"CombinedOutput"}
--- FAIL: TestScript (0.00s)
    --- FAIL: TestScript/encryptedname (0.05s)
        testscript.go:429: # test that chezmoi add --encrypt-name adds a file and encrypts its name in the source state (0.037s)
            > cp golden/.file $HOME/.file
            > exec chezmoi add --encrypt-name $HOME${/}.file
            [stderr]
            chezmoi: lstat $WORK/home/user/.local/share/chezmoi/encryptedname_LS0tLS1CRUdJTiBBR0UgRU5DUllQVEVEIEZJTEUtLS0tLQpZV2RsTFdWdVkzSjVjSFJwYjI0dWIzSm5MM1l4Q2kwK0lGZ3lOVFV4T1NCb2IyOVNSMFp4YTBwM1VrNW5UMlJ4ClpYZE5hM05qUlZsT09VTk5VVEZVUmlzNFNHbHVkakJVYlc1bkNucHFWVU55VEhCQ2RFRnFSVFZUY21aUU1sSnMKU1dWUlIxUlNSamh4VEdKaU1ETXZheXRET1ZsMVNXOEtMUzB0SUVWcWJqWkNaV1JtWTAxTU1ESmhVbWR5WVRSVApWbE41TWs4NE4wZEJNVmRDUW5WaE1GUlRlbUpoVjBFS3ptaXZCRDl0WHVsVmc4MFR2bFZTNEZWRi9YMkIzVDJNCm9Fb1BpRGlHa3U4bm5BbjBTTXEvalE9PQotLS0tLUVORCBBR0UgRU5DUllQVEVEIEZJTEUtLS0tLQo: file name too long
            [exit status 1]
            FAIL: testdata/scripts/encryptedname.txtar:7: unexpected command failure
            
FAIL
FAIL    github.com/twpayne/chezmoi/v2/pkg/cmd   0.509s
FAIL
```

The "file name too long" error is due to the limits of the operating system (I've tried on macOS and Linux) and is something that chezmoi cannot fix. When asymmetric encryption is used, the ciphertext is larger than the plaintext, and the use of base64 encoding further doubles the size of the encrypted part of the filename compared to the unencrypted filename.

I'm not sure if there's a solution to this. Do you have any ideas?

twpayne:
As written in https://github.com/twpayne/chezmoi/issues/2570#issuecomment-1326864321, sadly this approach looks like a non-starter.

PR #2591, opened 2022-11-21
Add YAML example for interpreters

alexvy86:
Adds an example of how interpreters are configured if using a YAML config file

### DIFF
--- a/assets/chezmoi.io/docs/reference/target-types.md
+++ b/assets/chezmoi.io/docs/reference/target-types.md
@@ -133,6 +133,17 @@ Script interpreters can be added or overridden with the
         command = "tclsh"
     ```
 
+    If using YAML for the config file, each extension should be a child node of
+    `interpreters`, like this:
+
+    ```yaml title="~/.config/chezmoi/chezmoi.yaml"
+    interpreters:
+      py:
+        command: 'C:\Python39\python3.exe'
+      tcl:
+        command: "tclsh"
+    ```
+
 !!! note
 
     If you intend to use PowerShell Core (`pwsh.exe`) as the `.ps1`
--- a/assets/chezmoi.io/docs/reference/target-types.md
+++ b/assets/chezmoi.io/docs/reference/target-types.md
@@ -113,13 +113,9 @@ The default script interpreters are:
 | `.ps1`    | `powershell` | `-NoLogo` |
 | `.rb`     | `ruby`       | *none*    |
 
-Script interpreters can be added or overridden with the
-`interpreters.`*extension* section in the configuration file.
-
-!!! note
-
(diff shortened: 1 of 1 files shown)


alexvy86 (commented):
Adds an example of how interpreters are configured if using a YAML config file

halostatue (approved):
It feels like we need to be a little clearer around line 116. Something like:

```markdown
Script interpreters can be added or overridden in the `interpreters`
section of the configuration file with *extension* keys.
```

After all, the TOML version *could* be done as:

```toml
[interpreters]
py = { command = 'C:\Python39\python3.exe' }
ps1 = { command = 'push', arguments = '-NoLogo' }
tcl = { command = "tclsh" }
```

alexvy86:
@halostatue what do you think of that rewrite?

assets/chezmoi.io/docs/reference/target-types.md
```diff
@@ -113,26 +113,40 @@ The default script interpreters are:
...
-Script interpreters can be added or overridden with the
-`interpreters.`*extension* section in the configuration file.
-
-!!! note
-
-    The leading `.` is dropped from *extension*, for example to specify the
-    interpreter for `.pl` files you configure `interpreters.pl`.
+Script interpreters can be added or overridden by adding the corresponding
+extension (without the leading dot) as a key under the `interpreters`
+section of the configuration file.
```
halostatue, line 118: I think that the `!!!note` is still useful, or something like it, because it makes it clear that the key for `.pl` is `pl`.

There are other changes that I would consider to show this, like the table:

```markdown
| Extension | Key   | Command      | Arguments |
| --------- | ----- | ------------ | --------- |
| `.pl`     | `pl`  | `perl`       | *none*    |
| `.py`     | `py`  | `python3`    | *none*    |
| `.ps1`    | `ps1` | `powershell` | `-NoLogo` |
| `.rb`     | `rb`  | `ruby`       | *none*    |
```

But that’s a bit more change than I would make without @twpayne agreeing. (Also, reading above, the bit about `%PATH%` should be `` in your `%PATH` or fully specified ``, but that’s *well* outside of the changes you’re contributing.

bradenhilton:
We should probably use tabs like in https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#embedded-content at some point to give examples for each config format, but we certainly don't need to do it in this PR.

assets/chezmoi.io/docs/reference/target-types.md
alexvy86: Re-added the note.

alexvy86:
Ready for a last review. Someone from the team will merge it, right? Or if I should, just let me know; if so I'll wait for a last confirmation that it looks good.

PR #2600, opened 2022-11-29
docs: Update stat.md to reference os.Stat

eugenesvk:
I assumed that `stat(2)` in 
>stat runs stat(2) on name

was meant as a reference to the docs explaining what `stat` is , but then the `(2)` is strange, so maybe it means something else

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/stat.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/stat.md
@@ -1,7 +1,7 @@
 # `stat` *name*
 
-`stat` runs `stat(2)` on *name*. If *name* exists it returns structured data.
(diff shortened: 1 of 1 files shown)


twpayne:
`stat(2)` refers to the `stat` system call in section 2 of the UNIX manual. It's a convention to refer to distinguish between programs (section 1), system calls (section 2), and C library functions (section 3).

Compare the output of

```console
$ man 1 stat
$ man 2 stat
```

In this case, referring to section 2 is correct.

eugenesvk:
Oh, ok
Well, `man 2 stat` is empty on my mac :) (`man 1 stat` isn't ), so there is nothing to compare it to, so I think some extra info on what stat is might be warranted

twpayne:
Fair enough, I'm showing my age here :)

PR #2615, opened 2022-12-05
docs: link to reference in script user guide

iliakur:
Thanks for this tool!

I'm converting my setup to chezmoi and I would have found it helpful to find out about `.chezmoiscripts` from the user guide.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/use-scripts-to-perform-actions.md
+++ b/assets/chezmoi.io/docs/user-guide/use-scripts-to-perform-actions.md
@@ -38,6 +38,9 @@ file in a temporary directory with the executable bit set, and then executes
 the contents with `exec(3)`. Consequently, the script's contents must either
 include a `#!` line or be an executable binary.
 
+By default chezmoi will copy your scripts to a target directory. Place them inside
+(`.chezmoiscripts`)[special-files-and-directories/chezmoiscripts] to disable this behavior.
+
 ## Set environment variables
 
 You can set extra environment variables for your scripts in the `scriptEnv`


twpayne:
Thanks for this, however the text isn't correct.

chezmoi doesn't copy scripts into the target directory. As [described in the reference guide](https://www.chezmoi.io/reference/target-types/#scripts):

> Scripts will normally run with their working directory set to their equivalent location in the destination directory. If the equivalent location in the destination directory either does not exist or is not a directory, then chezmoi will walk up the script's directory hierarchy and run the script in the first directory that exists and is a directory.

The actual script file is created in a temporary directory, and the script is run with its working directory set as above. As the `.chezmoiscripts` directory does not exist the target directory, scripts in `.chezmoiscripts` are run with their working directory equal to the target directory.

Would you be able to update the text in this PR to reflect this?

iliakur:
Thanks for the clarification @twpayne. Let me try to recap it to make sure I understood it, please correct me.

Assuming the following setup:

- target dir: `~/target`, contains subdir `foo`
- script1 location: `$CHEZMOI_ROOT/bar/run_1.sh`
- script2 location: `$CHEZMOI_ROOT/foo/run_2.sh`
- script3 location: `$CHEZMOI_ROOT/.chezmoiscripts/run_3.sh`

We expect the following to be true:

1. **Each script** is copied to a temporary folder.
2. script1 wd: `$CHEZMOI_ROOT/bar`
3. script2 wd: `~/target/foo`
4. script3 wd: `$CHEZMOI_ROOT/.chezmoiscripts`

Did I miss anything?

twpayne:
This understanding is not completely correct.

> We expect the following to be true:
> 
> 1. **Each script** is copied to a temporary folder.

No. chezmoi generates each script and writes it to a temporary folder. Scripts can be templates, so copying is not sufficient.

> 2. script1 wd: `$CHEZMOI_ROOT/bar`

No. The script1 wd is `~/target`. As `~/target/bar` does not exist, the working directory is the first parent directory of `~/target/bar` that does exist, i.e. `~/target`.

> 3. script2 wd: `~/target/foo`

Yes.

> 4. script3 wd: `$CHEZMOI_ROOT/.chezmoiscripts`

No. The script3 working directory is `~/target` for the same reasons as 2. If the directory `~/target/.chezmoiscripts` existed then `~/target/.chezmoiscripts` would be the working directory.

If you want to know what the working directory of a script is when chezmoi runs it, you can put:

```sh
pwd 1>&2
```

near the top of your script, which will print the script's working directory to stderr.

PR #2621, opened 2022-12-09
feat: Add verification of external checksums

### LINKED ISSUE #2620: Verify hash-sum of externally downloaded file

This is about [external files](https://www.chezmoi.io/user-guide/include-files-from-elsewhere/). I like pinning all my dependencies, but even if I specify a URL like

https://github.com/twpayne/chezmoi/releases/download/v2.27.3/chezmoi-2.27.3-riscv64.rpm

It’s still possible that the account get compromised and a different file is uploaded with the same name.

Add a string [field](https://www.chezmoi.io/reference/special-files-and-directories/chezmoiexternal-format/) called `hash` or `checksum` or whatever name; the value would be encoded hash-sum.

Algorithms in standard library: https://pkg.go.dev/crypto@go1.19.4#Hash

I’m thinking about Hex v.s. Base64 encoding. Hex is easier to produce, but Base64 is shorter. Base64 can be achieved [with OpenSSL CLI](https://unix.stackexchange.com/questions/3675/how-can-i-get-a-base64-encoded-shax-on-the-cli/3713#3713).

(Just not implement this feature?)

## Additional context

N/A

twpayne:
Fixes #2620.

### DIFF
--- a/assets/chezmoi.io/docs/reference/special-files-and-directories/chezmoiexternal-format.md
+++ b/assets/chezmoi.io/docs/reference/special-files-and-directories/chezmoiexternal-format.md
@@ -35,6 +35,14 @@ Entries may have the following fields:
 | `refreshPeriod`   | duration | `0`           | Refresh period                                                |
 | `stripComponents` | int      | `0`           | Number of leading directory components to strip from archives |
 | `url`             | string   | *none*        | URL                                                           |
+| `checksum.sha256` | string   | *none*        | Expected SHA256 checksum of data                              |
+| `checksum.sha384` | string   | *none*        | Expected SHA384 checksum of data                              |
+| `checksum.sha512` | string   | *none*        | Expected SHA512 checksum of data                              |
+| `checksum.size`   | int      | *none*        | Expected size of data                                         |
+
+If any of the optional `checksum.sha256`, `checksum.sha384`, or
+`checksum.sha512` fields are set, chezmoi will verify that the downloaded data
+has the given checksum.
 
 The optional boolean `encrypted` field specifies whether the file or archive is
 encrypted.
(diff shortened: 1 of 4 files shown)


twpayne:
Note: `size`, `md5`, `ripemd160`, and `sha1` are supported but undocumented as they are all considered insecure now.

pkg/chezmoi/sourcestate.go
```diff
@@ -1371,6 +1380,55 @@ func (s *SourceState) getExternalData(
 		return nil, err
 	}
 
+	if external.Checksum.Size != 0 {
```
halostatue, line 1383: Not a must-have, but it would be nice if the checksum process could run all of the present checksum values and then output all of the errors at once on output.

assets/chezmoi.io/docs/reference/special-files-and-directories/chezmoiexternal-format.md
```diff
@@ -35,6 +35,13 @@ Entries may have the following fields:
 | `refreshPeriod`   | duration | `0`           | Refresh period                                                |
 | `stripComponents` | int      | `0`           | Number of leading directory components to strip from archives |
 | `url`             | string   | *none*        | URL                                                           |
+| `checksum.sha256` | string   | *none*        | Expected SHA256 checksum of data                              |
+| `checksum.sha384` | string   | *none*        | Expected SHA384 checksum of data                              |
+| `checksum.sha512` | string   | *none*        | Expected SHA512 checksum of data                              |
```
halostatue, line 40: I would *personally* document `` `checksum.size` `` with the caveat `Expected number of bytes of data, should not be used by itself`.

The benefit of `checksum.size` is that it’s a good cross-check against size stuffing to get a collision.

halostatue (approved):
Looks great. I have a couple of suggestions, but they aren’t requirements or blockers.

pkg/chezmoi/sourcestate.go
twpayne: Thanks, done.

assets/chezmoi.io/docs/reference/special-files-and-directories/chezmoiexternal-format.md
twpayne: I've added `checksum.size` to the table.

PR #2633, opened 2022-12-21
feat: Add support for source paths in managed command

twpayne:
Fixes #2627.


Lockszmith-GH:
This is EXACTLY what I was looking for. Thanks so much!

PR #2640, opened 2022-12-27
chore: Switch to github.com/twpayne/go-diff

twpayne:
Refs https://github.com/sergi/go-diff/issues/134.


twpayne:
`github.com/sergi/go-diff` just released v1.3.1 which includes the key bug fix, so this is no longer needed.

PR #2648, opened 2022-12-28
feat: Allow cd command to take the destination directory as an argument

### LINKED ISSUE #2646: `chezmoi cd` to cd into `.chezmoiroot`'s specified subdirectory

Whenever I do `chezmoi cd` to jump to my `$HOME` directory's source state. I have to do an additional `cd HOME` to actually get there. It's a small thing, but I feel that after setting up `.chezmoiroot`, the root of the git repository is rarely ever needed. 

If `.chezmoiroot` exists in the root of the source directory then I'd like `chezmoi cd` to change to the directory specified in `.chezmoiroot` instead of simply going to the root of the git repository.

```shell
$ cat ~/.local/share/chezmoi/.chezmoiroot
HOME

$ chezmoi cd
% pwd
~/.local/share/chezmoi/HOME
% exit
```

- The alternative without this feature would be to `chezmoi cd` into the root of the source directory and then `cd` into the subdirectory chezmoi reads the source state from. 
- If changing `chezmoi cd`'s default behavior is too drastic, perhaps an option to change it could be added to chezmoi's config.

twpayne:
Fixes #2646.

### DIFF
--- a/pkg/cmd/cdcmd.go
+++ b/pkg/cmd/cdcmd.go
@@ -39,6 +39,13 @@ func (c *Config) runCDCmd(cmd *cobra.Command, args []string) error {
 	var dir chezmoi.AbsPath
 	if len(args) == 0 {
 		dir = c.WorkingTreeAbsPath
+	} else if argAbsPath, err := chezmoi.NewAbsPathFromExtPath(args[0], c.homeDirAbsPath); err != nil { //nolint:gocritic
+		return err
+	} else if argAbsPath == c.DestDirAbsPath {
+		dir, err = c.getSourceDirAbsPath(nil)
+		if err != nil {
+			return err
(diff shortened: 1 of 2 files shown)


halostatue:
We may want a special case handling on Windows, since Windows doesn’t autoexpand `~` and although it generally means `$HOME`, in the event that it doesn’t get auto-expanded, we might want to do that.

twpayne:
Good point! I hadn't thought of that :)

But, I think it's OK as chezmoi already does `~` expansion for paths on Windows itself: https://github.com/twpayne/chezmoi/blob/4d5d68cbcceedafabf0cb2ad9d1f4400e02306e3/pkg/chezmoi/path_windows.go#L15-L16

Issue #2676, opened 2023-01-05
chezmoi asks password even though --no-password is set as arg in keepassxc config section


jcapiitao:
## Is your feature request related to a problem? Please describe.

`chezmoi` asks to insert password to unlock KeePassXC database even though we set `--no-password` in the config file.
```
[keepassxc]
    database = "passwords.kdbx"
    args = ["-k", "/run/secrets/keepassxc-key", "--no-password"]
```
```
$ chezmoi diff
Insert password to unlock /home/chezmoi/passwords.kdbx:
```

In this case, we want to use the key file to decrypt the KeePassXC database, instead of password.

## Describe the solution you'd like

One solution would be to turn off the password prompt when `--no-password` is set in `args`

## Describe alternatives you've considered

Make the KeepassXC CLI  support persistent login.

## Additional context

As mentioned in the documentation `The KeepassXC CLI does not currently support any persistent login`, so it's a bit
annoying having to type the password each time `chezmoi` needs to render template.
That's why I'm considering using the key file as authentication method.
I'm running `chezmoi` inside a container with podman, wit my home directory mounted as a volume.
The file `/run/secrets/keepassxc-key` is a mounted file handled by [`podman secrets`](https://www.redhat.com/sysadmin/new-podman-secrets-command)

PR #2677, opened 2023-01-05
feat: Add keepassxc.prompt config variable

jcapiitao:
Fixes https://github.com/twpayne/chezmoi/issues/2676.

### DIFF
--- a/pkg/cmd/keepassxctemplatefuncs.go
+++ b/pkg/cmd/keepassxctemplatefuncs.go
@@ -24,6 +24,7 @@ type keepassxcConfig struct {
 	Command         string          `json:"command" mapstructure:"command" yaml:"command"`
 	Database        chezmoi.AbsPath `json:"database" mapstructure:"database" yaml:"database"`
 	Args            []string        `json:"args" mapstructure:"args" yaml:"args"`
+	Prompt          bool            `json:"prompt" mapstructure:"prompt" yaml:"prompt"`
 	version         *semver.Version
 	cache           map[string]map[string]string
 	attachmentCache map[string]map[string]string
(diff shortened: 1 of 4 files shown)


halostatue:
LGTM, but I don’t use KeePassXC to be able to test it.

pkg/cmd/keepassxctemplatefuncs.go
```diff
@@ -133,7 +133,14 @@ func (c *Config) keepassxcOutput(name string, args []string) ([]byte, error) {
 		panic(errors.New("keepassxc.database not set"))
 	}
 
-	if c.Keepassxc.password == "" {
+	var noPassword bool
+	for _, v := range args {
+		if v == "--no-password" {
+			noPassword = true
+		}
+	}
```
twpayne, line 141: Thanks for this PR. Having chezmoi parse the arguments to keepassxc is fragile as it depends on chezmoi parsing arguments exactly the same way as keepassxc does.

Instead, I think it's better to have a `keepassxc.promptPassword` [configuration variable](https://www.chezmoi.io/reference/configuration-file/variables/) which defaults to `true` and can be set to `false` to disable the password prompt.

Also, below, `cmd.Stdin` should be set to `os.Stdin` if the user disables the password prompt.

Let me know if you want to update this PR, otherwise I can create one. For adding a configuration variable, commit 8cd18326 gives an idea of what files need to be updated.

jcapiitao: I think you solution is better as well :+1: 
I'm gonna update the PR soon

twpayne:
Perfect, thank you!

Issue #2609, opened 2022-12-04
Umask from config file is not used


zroug:
## Describe the bug

Chezmoi ignores a umask that is set in the configuration file.

## To reproduce

1. Check that the default umask is less restrictive than desired.
```console
$ umask
0022
```

2. Set a more restrictive umask in the configuration file.
```console
$ cat .config/chezmoi/chezmoi.yaml 
umask: 0o077
```

3.  Observe that Chezmoi does not honor the new umask.
```console
$ chezmoi diff
diff --git a/file b/file
old mode 40700
new mode 40755
```

## Expected behavior

Chezmoi should use the umask from the configuration file. So in the example above, the diff would be empty.

## Output of command with the `--verbose` flag

Output is the same.

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT    CHECK                MESSAGE
ok        version              v2.27.2, commit 882d0808feb1fc8112b411ed2216f31306656861, built at 2022-11-24T23:41:28Z, built by goreleaser
ok        latest-version       v2.27.2
ok        os-arch              linux/amd64 (Fedora Linux 37.20221130.0 (Silverblue))
ok        uname                Linux l6zdz0 6.0.10-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Nov 26 16:55:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
ok        go-version           go1.19.3 (gc)
ok        executable           ~/bin/chezmoi
ok        upgrade-method       replace-executable
ok        config-file          ~/.config/chezmoi/chezmoi.yaml, last modified 2022-11-29T19:42:35+01:00
warning   source-dir           ~/.local/share/chezmoi is a git working tree (dirty)
ok        suspicious-entries   no suspicious entries
warning   working-tree         ~/.local/share/chezmoi is a git working tree (dirty)
ok        dest-dir             ~ is a directory
ok        umask                022
ok        cd-command           found /bin/bash
ok        cd-args              /bin/bash
info      diff-command         not set
ok        edit-command         found /usr/bin/nano
ok        edit-args            /usr/bin/nano
ok        git-command          found /usr/bin/git, version 2.38.1
warning   merge-command        vimdiff not found in $PATH
ok        shell-command        found /bin/bash
ok        shell-args           /bin/bash
info      age-command          age not found in $PATH
ok        gpg-command          found /usr/bin/gpg, version 2.3.8
[...]

PR #2688, opened 2023-01-12
chore: Fix umask tests

twpayne:


twpayne:
So what happened here was:
* Umask (file permission masks) vary across operating systems. For example macOS uses `022`, Ubuntu uses `002`, and Windows has a completely different way of implementing file permissions.
* Different file permissions mean that chezmoi's behavior when file permissions differ (e.g. the output of `chezmoi diff` or `chezmoi status`) depend on the setting of umask.
* chezmoi v1 introduced some tests to make sure that it behaved correctly for common umask values in #1149.
* The tests used [a package global variable](https://github.com/twpayne/chezmoi/blob/b35dd65e9a006b5383129cc473d313266fc4f1d3/pkg/chezmoitest/chezmoitest_unix.go#LL15C7-L15C7) to set the umask for the tests.
* This variable is set via a [Go linker flag](https://pkg.go.dev/cmd/link), which includes the package's full name.
* [This variable](https://github.com/twpayne/chezmoi/blob/b35dd65e9a006b5383129cc473d313266fc4f1d3/pkg/chezmoitest/chezmoitest_unix.go#L15) used the [original package's import path](https://github.com/twpayne/chezmoi/blob/b35dd65e9a006b5383129cc473d313266fc4f1d3/Makefile#L86) (note the lack of `v2` when `UmaskStr` is set).
* When chezmoi's import path changed with the release of v2, this reference to the `UmaskStr` variable was no longer valid and was silently ignored by the Go toolchain.
* Consequently, the setting of umasks has been broken since the merge of v2.
* #2609 was the first user report of this.

This PR updates chezmoi's tests to work with different umasks. I think chezmoi is behaving correctly since #2609, so this PR only changes the tests, not the code.

halostatue (approved):
LGTM. I don’t use this functionality, but the tests do look better.

PR #2699, opened 2023-01-17
fix: walk flattened external archives

lukekarrys:
Tarballs from `registry.npmjs.org` only include a flattened list of files, and thus don't contain any tar headers where `header.Typeflag == tar.TypeDir`.

This causes `chezmoi apply` to fail the first time it tries to create a file whose parent directory does not exist.

Here's an example of an `npm` tarball vs one from `GitHub`:

```sh
> curl -sL https://github.com/npm/node-semver/tarball/v7.3.8 --output gh.tgz

> curl -sL https://registry.npmjs.org/semver/-/semver-7.3.8.tgz --output npm.tgz

> tar ztvf npm.tgz | grep "^d" # no output

> tar ztvf gh.tgz | grep "^d"
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/.github/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/.github/ISSUE_TEMPLATE/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/.github/matchers/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/.github/workflows/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/bin/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/classes/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/functions/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/internal/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/ranges/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/scripts/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/tap-snapshots/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/tap-snapshots/test/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/tap-snapshots/test/bin/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/bin/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/classes/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/fixtures/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/functions/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/internal/
drwxrwxr-x  0 root   root        0 Oct  4 12:37 npm-node-semver-dc0fe20/test/ranges/
```

And here's a `.chezmoiexternal` example that fails with the following error:

```sh
> cat .chezmoiexternal.toml
["npm"]
    type = "archive"
    url = "https://registry.npmjs.org/semver/-/semver-7.3.8.tgz"
    exact = true
    stripComponents = 1

["github"]
    type = "archive"
    url = "https://github.com/npm/node-semver/tarball/v7.3.8"
    exact = true
    stripComponents = 1

> chezmoi apply --init
chezmoi: stat /Users/lukekarrys/Desktop/chezmoi-test/dest/npm/bin: no such file or directory
```

## Solution

This PR keeps track of all seen directories inside `walkArchiveTar` and if a file is seen before its directory, the walk function will first process a directory.

## Caveats

I was not sure what to do about the `Mode` of the implicit directories, so they are currently always set to `0o755`.

Also this is my first time writing `Go` so I have no idea how idiomatic any of the code is, except all the `make` commands pass.

### DIFF
--- a/pkg/chezmoi/archive.go
+++ b/pkg/chezmoi/archive.go
@@ -132,10 +132,44 @@ func isTarArchive(r io.Reader) bool {
 	return err == nil
 }
 
+func implicitDirHeader(dir string, fileHeader *tar.Header) *tar.Header {
+	return &tar.Header{
+		Typeflag: tar.TypeDir,
+		Name:     dir,
+		Mode:     0o777,
+		Size:     0,
+		Uid:      fileHeader.Uid,
+		Gid:      fileHeader.Gid,
+		Uname:    fileHeader.Uname,
+		Gname:    fileHeader.Gname,
+		ModTime:  fileHeader.ModTime,
+	}
+}
+
 // walkArchiveTar walks over all the entries in a tar archive.
 func walkArchiveTar(r io.Reader, f WalkArchiveFunc) error {
 	tarReader := tar.NewReader(r)
 	var skippedDirPrefixes []string
+	seenDirs := newSet[string]()
+	processHeader := func(header *tar.Header, dir string) error {
+		for _, skippedDirPrefix := range skippedDirPrefixes {
+			if strings.HasPrefix(header.Name, skippedDirPrefix) {
+				return fs.SkipDir
+			}
+		}
+		if seenDirs.contains(dir) {
+			return nil
+		}
+		seenDirs.add(dir)
+		name := strings.TrimSuffix(header.Name, "/")
+		switch err := f(name, header.FileInfo(), tarReader, header.Linkname); {
+		case errors.Is(err, fs.SkipDir):
+			skippedDirPrefixes = append(skippedDirPrefixes, header.Name)
+		case err != nil:
+			return err
+		}
+		return nil
+	}
 HEADER:
 	for {
 		header, err := tarReader.Next()
@@ -145,24 +179,28 @@ HEADER:
 		case err != nil:
 			return err
 		}
-		for _, skippedDirPrefix := range skippedDirPrefixes {
-			if strings.HasPrefix(header.Name, skippedDirPrefix) {
-				continue HEADER
-			}
-		}
-		name := strings.TrimSuffix(header.Name, "/")
 		switch header.Typeflag {
-		case tar.TypeDir, tar.TypeReg:
-			switch err := f(name, header.FileInfo(), tarReader, ""); {
-			case errors.Is(err, fs.SkipDir):
-				skippedDirPrefixes = append(skippedDirPrefixes, header.Name)
-			case errors.Is(err, Break):
-				return nil
-			case err != nil:
-				return err
+		case tar.TypeReg, tar.TypeDir, tar.TypeSymlink:
+			if header.Typeflag == tar.TypeReg {
+				dirs, _ := path.Split(header.Name)
+				dirComponents := strings.Split(strings.TrimSuffix(dirs, "/"), "/")
+				for i := range dirComponents {
+					dir := strings.Join(dirComponents[0:i+1], "/")
+					if len(dir) > 0 {
+						switch err := processHeader(implicitDirHeader(dir+"/", header), dir+"/"); {
+						case errors.Is(err, fs.SkipDir):
+							continue HEADER
+						case errors.Is(err, Break):
+							return nil
+						case err != nil:
+							return err
+						}
+					}
+				}
 			}
-		case tar.TypeSymlink:
-			switch err := f(name, header.FileInfo(), nil, header.Linkname); {
+			switch err := processHeader(header, header.Name); {
+			case errors.Is(err, fs.SkipDir):
+				continue HEADER
 			case errors.Is(err, Break):
 				return nil
 			case err != nil:
--- a/pkg/chezmoi/archive_test.go
+++ b/pkg/chezmoi/archive_test.go
@@ -12,50 +12,73 @@ import (
 )
 
 func TestWalkArchive(t *testing.T) {
+	nestedRoot := map[string]any{
+		"dir1": map[string]any{
+			"subdir1": map[string]any{
+				"file1": "",
+				"file2": "",
+			},
+			"subdir2": map[string]any{
+				"file1": "",
+				"file2": "",
+			},
+		},
+		"dir2": map[string]any{
+			"subdir1": map[string]any{
+				"file1": "",
+				"file2": "",
+			},
+			"subdir2": map[string]any{
+				"file1": "",
+				"file2": "",
+			},
+		},
+		"file1":    "",
+		"file2":    "",
+		"symlink1": &archivetest.Symlink{Target: "file1"},
+		"symlink2": &archivetest.Symlink{Target: "file2"},
+	}
+	flatRoot := map[string]any{
+		"dir1/subdir1/file1": "",
+		"dir1/subdir1/file2": "",
+		"dir1/subdir2/file1": "",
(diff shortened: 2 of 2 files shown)


pkg/chezmoi/archive.go
```diff
@@ -132,10 +134,44 @@ func isTarArchive(r io.Reader) bool {
 	return err == nil
 }
 
+func ImplicitDirHeader(dir string, fileHeader *tar.Header) *tar.Header {
```
twpayne, line 137: This function should be called `implicitDirHeader` (lowercase `i`) as it's for package internal use only (the uppercase `I` makes it exported).

pkg/chezmoi/archive.go
```diff
@@ -132,10 +134,44 @@ func isTarArchive(r io.Reader) bool {
...
+		Gname:    fileHeader.Gname,
+		ModTime:  fileHeader.ModTime,
+	}
+}
+
 // walkArchiveTar walks over all the entries in a tar archive.
 func walkArchiveTar(r io.Reader, f WalkArchiveFunc) error {
 	tarReader := tar.NewReader(r)
 	var skippedDirPrefixes []string
+	var seenDirs []string
```
twpayne, line 155: `seenDirs` should be a `map[string]struct{}` for fast lookups. Using `slice.Contains` does a linear search which makes the algorithm accidentally quadratic.

pkg/chezmoi/archive.go
```diff
@@ -145,24 +181,28 @@ HEADER:
...
-			case errors.Is(err, fs.SkipDir):
-				skippedDirPrefixes = append(skippedDirPrefixes, header.Name)
-			case errors.Is(err, Break):
-				return nil
-			case err != nil:
-				return err
+		case tar.TypeReg, tar.TypeDir, tar.TypeSymlink:
+			if header.Typeflag == tar.TypeReg {
+				dirs, _ := filepath.Split(header.Name)
+				dirComponents := strings.Split(strings.TrimSuffix(dirs, "/"), "/")
```
twpayne, line 188: The path manipulation functions here shout use the [`path` package](https://pkg.go.dev/path) instead of `filepath` and `strings`. `filepath` has OS-specific path handling code (e.g. it handles `\` as a path separator on Windows, but not on Linux). Since we're reading from a tar archive here, we can assume `/`-separated paths, i.e. use the `path` package.

pkg/chezmoi/archive_test.go
```diff
@@ -11,51 +11,76 @@ import (
...
+	"dir1/subdir2/file2": "",
+	"dir2/subdir1/file1": "",
+	"dir2/subdir1/file2": "",
+	"dir2/subdir2/file1": "",
+	"dir2/subdir2/file2": "",
+	"file1":              "",
+	"file2":              "",
+	"symlink1":           &archivetest.Symlink{Target: "file1"},
+	"symlink2":           &archivetest.Symlink{Target: "file2"},
+}
```
twpayne, line 54: These two variables can be moved inside the `TestWalkArchive` function as they're only used in that function and so don't need to be global.

twpayne (commented):
Thank you, this is an excellent PR and very impressive for a first Go contribution!

There are a few small comments. If you want to address them, please do.  If you'd rather not, then I'll happily merge this PR and fix them in a follow-up PR.

pkg/chezmoi/archive.go
```diff
@@ -132,10 +134,44 @@ func isTarArchive(r io.Reader) bool {
 	return err == nil
 }
 
+func ImplicitDirHeader(dir string, fileHeader *tar.Header) *tar.Header {
+	return &tar.Header{
+		Typeflag: tar.TypeDir,
+		Name:     dir,
+		Mode:     0o755,
```
twpayne, line 141: This mode can be `0o777`. The user's umask will clear the appropriate group and world bits.

pkg/chezmoi/archive.go
```diff
@@ -11,9 +11,11 @@ import (
 	"io"
 	"io/fs"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/ulikunitz/xz"
+	"golang.org/x/exp/slices"
```
halostatue, line 18: This can be removed by using `map[string]struct{}` [as suggested](https://github.com/twpayne/chezmoi/pull/2699/files#r1072962956).

pkg/chezmoi/archive.go
lukekarrys: I wasn't sure what the `struct{}` would be in this case? I opted for `map[string]bool` to keep it simple since I'm only using it as a flag.

lukekarrys:
I pushed a `fixup!` commit implementing the feedback (I can squash too, but wanted to keep it easier to review).

I resolved the conversations that I think I handled and kept one open that I had a question about.

pkg/chezmoi/archive.go
halostatue: That would work, and I’m not sure whether `map[string]struct{}` would be better or exactly what it would be myself. At a guess, something other than `bool` would allow you to potentially put a `tar.Header` (the `implicitDirHeader` output?) or something similar in place.

halostatue: I think that @twpayne had a clearer idea on this when he wrote his original comment.

halostatue (approved):

pkg/chezmoi/archive.go
twpayne: `map[string]struct{}` is a weird Go idiomatic way of representing a set of things, in this case `string`s. The magic is that the map key type is the empty struct, which has zero size. Therefore the map contains the keys but no values.

chezmoi actually already has a generic implementation of sets that you can use in [`pkg/chezmoi/set.go`](https://github.com/twpayne/chezmoi/blob/36eec23ecd852a34d035a3f1ac5fde19c7550daf/pkg/chezmoi/set.go). With this you can write:

```go
    seenDirs := newSet[string]()
```

and then use the `seenDirs.add` and `seenDirs.contains` methods.

twpayne:
This is great, thank you for your work and patience! Please squash the commits and I'll merge.

lukekarrys:
Commits have been squashed 🎉 

@twpayne Thank you for the patient feedback and teaching me new things about go. This project is very beneficial to me so I'm happy to do a small part in helping. It was also very intuitive to go from zero->PR in a new language, so great work on that as well.

Issue #2700, opened 2023-01-18
Can't extract archive in .chezmoiexternal.yaml


szinn:
## Describe the bug

Adding in the JetBrainsMono font to be extracted via .chezmoiexternal support fails with a 'no such file or directory'

## To reproduce

Create .chezmoiexternal.yaml with below and 'chezmoi -R apply'

.chezmoiexternal.yaml:
```
"Library/Fonts/JetBrainsMono":
  url: "https://download.jetbrains.com/fonts/JetBrainsMono-2.304.zip"
  type: archive
  exact: true
  refreshPeriod: 168h
```

## Expected behavior

I would expect the archive to be extracted.

## Output of command with the `--verbose` flag

```console
$ chezmoi -R apply

chezmoi: mkdir /Users/scotte/Library/Fonts/JetBrainsMono/fonts/ttf: no such file or directory
```

## Output of `chezmoi doctor`

```console
$ chezmoi doctor
RESULT    CHECK                MESSAGE
ok        version              v2.29.2, commit b9536aae9100a2367e7b5baad8a7e251542c45bd, built at 2023-01-15T14:35:48Z, built by Homebrew
ok        latest-version       v2.29.2
ok        os-arch              darwin/arm64
ok        uname                Darwin Odin.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:03:51 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T6000 arm64 arm Darwin
ok        go-version           go1.19.5 (gc)
ok        executable           /opt/homebrew/bin/chezmoi
ok        upgrade-method       brew-upgrade
ok        config-file          ~/.config/chezmoi/chezmoi.yaml, last modified 2023-01-16T21:06:27-05:00
warning   source-dir           ~/.local/share/chezmoi is a git working tree (dirty)
warning   suspicious-entries   ~/.local/share/chezmoi/home/.chezmoi.yaml and ~/.local/share/chezmoi/home/dot_config/chezmoi/chezmoi.yaml.tmpl
warning   working-tree         ~/.local/share/chezmoi is a git working tree (dirty)
ok        dest-dir             ~ is a directory
ok        umask                022
ok        cd-command           found /opt/homebrew/bin/fish
ok        cd-args              /opt/homebrew/bin/fish
info      diff-command         not set
ok        edit-command         found /opt/homebrew/bin/nvim
ok        edit-args            /opt/homebrew/bin/nvim
[...]

PR #2701, opened 2023-01-19
fix: Support .zip files with empty internal directories

szinn:
Fix for https://github.com/twpayne/chezmoi/issues/2700

Zip files can have multiple levels of directories that don't have files in them. This fix is similar to https://github.com/twpayne/chezmoi/commit/e74a926a464961f66cdb3476eb1f2ef79f56f5e9 but applies to .zip files.

Essentially a mkdir -p on the path to the file prior to writing the file.

### DIFF
--- a/pkg/chezmoi/archive.go
+++ b/pkg/chezmoi/archive.go
@@ -12,6 +12,7 @@ import (
 	"io/fs"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/ulikunitz/xz"
 )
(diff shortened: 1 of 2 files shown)


pkg/chezmoi/archive.go
```diff
@@ -221,6 +218,29 @@ func walkArchiveZip(r io.ReaderAt, size int64, f WalkArchiveFunc) error {
...
+			}
+		}
+		if seenDirs.contains(dir) {
+			return nil
+		}
+		seenDirs.add(dir)
+		name := strings.TrimSuffix(dir, "/")
+		dirFileInfo := implicitDirHeader(dir, fileInfo.ModTime()).FileInfo()
+		switch err := f(name, dirFileInfo, nil, ""); {
+		case errors.Is(err, fs.SkipDir): {
```
twpayne, line 235: nit: the `{` isn't needed here, this is what's causing lint to complain.

twpayne (commented):
Looks good, thank you. There's a minor lint error. Otherwise, please could you squash the commits into one and we can merge this.

PR #2715, opened 2023-01-27
feat: Add zstd compression support

### LINKED ISSUE #2713: Support zstd compressed tarballs in .chezmoiexternal.toml

https://github.com/ankitects/anki/releases releases tarballs zstd compressed (.zst), I would like to install their software with an entry in `.chezmoiexternal.toml` but it seems there is no support for that compression format.

The following error is given:
```
chezmoi: .Software/anki/2.1.54-qt6: https://github.com/ankitects/anki/releases/download/2.1.54/anki-2.1.54-linux-qt6.tar.zst: unknown archive format
```
with this configuration:
```
[".Software/anki/2.1.54-qt6"]
type="archive"
url="https://github.com/ankitects/anki/releases/download/2.1.54/anki-2.1.54-linux-qt6.tar.zst"
```

Best solution would be that one can add just:
```
[".Software/anki/2.1.54-qt6"]
type="archive"
url="https://github.com/ankitects/anki/releases/download/2.1.54/anki-2.1.54-linux-qt6.tar.zst"
```
to `.chezmoiexternal.toml` and based on the extention `.zst` it does the right thing. Adding also the `format=zstd` hint would be acceptable, if auto detection is not possible.

Since it is an rather new format (not sure since when it is also supported in `tar` )would it be possible maybe to add an option `archive.extract_option` (or something like that) so that one could configure:

```
[".Software/anki/2.1.54-qt6"]
type="archive"
url="https://github.com/ankitects/anki/releases/download/2.1.54/anki-2.1.54-linux-qt6.tar.zst"
archive.extract_option=--zstd
```
So the tar command would use that extract option instead of the usual suspects like `-z` or `-J`. This would make it quite flexible and future proof, in case `tar` would add support also for other algorithms in future. And the feature would depend on when ever the local `tar` version supports it, no need to test from chezmoi.

Of course I can download it by hand and extract it by hand, but chezmoi is supposed to support me with that step.

twpayne:
Fixes #2713.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -96,6 +96,8 @@ linters:
 linters-settings:
   forbidigo:
     forbid:
+    - ^archive/zip\.
+    - ^compress/gzip\.
     - ^fmt\.Print.*$
     - ^ioutil\..*$
     - ^os\.(DirEntry|FileInfo|FileMode|Is.*|Mode.*)$
--- a/assets/chezmoi.io/docs/developer/developing-locally.md
+++ b/assets/chezmoi.io/docs/developer/developing-locally.md
@@ -36,7 +36,7 @@ $ go test ./...
 chezmoi's tests include integration tests with other software. If the other
 software is not found in `$PATH` the tests will be skipped. Running the full
 set of tests requires `age`, `base64`, `bash`, `gpg`, `perl`, `python3`,
-`ruby`, `sed`, `sha256sum`, `unzip`, `xz`, and `zip`.
+`ruby`, `sed`, `sha256sum`, `unzip`, `xz`, `zip`, and `zstd`.
 
 Run chezmoi:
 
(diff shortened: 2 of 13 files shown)


bradenhilton:
[File extension is `.zst`](https://datatracker.ietf.org/doc/html/rfc8478#section-6.1) :)

bradenhilton (approved):

pkg/cmd/testdata/scripts/importtarzstd.txtar
```diff
@@ -0,0 +1,13 @@
+[(darwin||openbsd||windows)] skip 'tar does not support zstd compression'
+[!exec:zstd] skip 'zstd not found in $PATH'
+exec tar -c -f archive.tar.zst --zstd archive
```
halostatue, line 3: This could be done with `exec tar -c -f - archive | zstd > archive.tar.zst` rather than expecting `--zstd`. macOS tar (`bsdtar`) *does* support `--use-compress-program`, so on macOS it could be `exec tar -c  --use-compress-program zstd -f archive.tar.zst archive`.

halostatue (approved):
Just a suggestion on the tests.

pkg/cmd/testdata/scripts/importtarzstd.txtar
twpayne: Thanks, done.

PR #2716, opened 2023-01-28
feat: Add external_ attribute for directories

### LINKED ISSUE #2714: Feature Request: Allow subdirectories to opt-out of script execution.

Chezmoi runs files with the `run_` prefix. However sometimes you have subdirectories where you don't want this to happen. A common example is if you use git submodules for vim plugins - any author of these could include a `run_tests.sh` or `run_evil.sh` and it would be executed next time you ran `chezmoi apply`.

Vendoring settings or configs is a pretty common pattern in dotfiles repos, and while chezmoi includes the option of using `.chezmoiexternal.toml` to include remote files, this isn't always an option.

It would be great if there was a way to indicate that chezmoi should ignore file prefixes below a certain subdirectory. For example a prefix `dot_vim/external_bundle/` or a sentinel file `dot_vim/bundle/.chezmoiexternal`.

twpayne:
Fixes #2714.

@peterbraden would you be able to test this? For each of your git submodules, you should add the `external_` attribute in their directories. In your case, the following should work:

```console
$ chezmoi chattr +external ~/.vim/bundle/*
```

You'll probably also not want chezmoi to copy `.git` directories from the git submodules. To prevent this, add

```
.vim/bundle/**/.git
```

to your `.chezmoiignore`.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/chattr.md
+++ b/assets/chezmoi.io/docs/reference/commands/chattr.md
@@ -16,6 +16,7 @@ modifiers and their abbreviations are:
 | `encrypted`        | *none*       |
 | `exact`            | *none*       |
 | `executable`       | `x`          |
+| `external`         | *none*       |
 | `once`             | `o`          |
 | `private`          | `p`          |
 | `readonly`         | `r`          |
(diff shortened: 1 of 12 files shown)


peterbraden:
It appears to work:

```sh
~/repos/chezmoi:master$ ./chezmoi apply
EVIL IS DONE!!
# Running peterbraden/dotfiles install.sh
# - Setting up a mac
~/repos/chezmoi:master$ git checkout fix-2714
branch 'fix-2714' set up to track 'origin/fix-2714'.
Switched to a new branch 'fix-2714'
~/repos/chezmoi:fix-2714$ make
go build . || ( rm -f chezmoi ; false )
~/repos/chezmoi:fix-2714$ ./chezmoi apply
# Running peterbraden/dotfiles install.sh
# - Setting up a mac
```

twpayne:
Brilliant, thank you for testing!

PR #2746, opened 2023-02-06
docs: Use country flags instead of ISO3166-1 country codes

twpayne:


halostatue:
I’m not sure that this is a good idea, as it conflates polities with languages.

halostatue:
There’s a website that talks about this, [Flags are not Languages](https://www.flagsarenotlanguages.com/blog/).

Apple [removed](https://www.flagsarenotlanguages.com/blog/apple-removes-flags-from-macos/) flag specification for languages in Monterey.

For the ones that you’ve chosen specifically:

- Chinese is spoken and written in mainland China (Cantonese and Mandarin are most common), Taiwan, Singapore, and many other languages.
- English is obviously spoken around the world, and the UK flag would be "more" appropriate, although the New Zealand, Australian, or Canadian flag (because the maple leaf is just the best flag 🇨🇦).
- French is spoken in Canada (🇨🇦 again!), Haïti, Congo, and loads of other countries in addition to France.
- Italian is primarily spoken in Italy, but of course it’s also spoken in your country of residence 😆 
- More people speak Portuguese and Spanish outside of Portugal (Brazil) and Spain (dozens of western hemisphere countries), respectively.

twpayne:
Fair point. I was hoping it would be quicker to visually scan than the country codes, but you're right.

PR #2751, opened 2023-02-07
feat: Extend toPrettyJson template function to take indent

### LINKED ISSUE #2749: Allow to use tabs for indentation in `toPrettyJson`

When you run `docker login`, unless you have a credentials helper, docker will store the credentials inside of `~/.docker/config.json`.

When docker does that, the file looks like:

```json
{
	"auths": {
		"ghcr.io": {
			"auth": "<token>"
		}
	}
}
```

However, when modifying this file using chezmoi modify templates and writing the result with `toPrettyJson`, it outputs with 2 spaces as indentation instead of tabs.

This causes chezmoi to think that there is a difference to the file, while essentially there is none.

Maybe a new function, like: `toFormattedJson "\t" $myDict`?

I'm trying to achieve something similar using `replaceAllRegex` but so far I was unable to. Any help in this regard would also be appreciated.

## Additional context

- https://github.com/twpayne/chezmoi/pull/2529#issuecomment-1298701628

twpayne:
Fixes #2749.

Another one for you, @felipecrs. I'm pretty sure this one works, but feel free to test it if you like :)

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/toPrettyJson.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/toPrettyJson.md
@@ -1,0 +1,11 @@
+# `toPrettyJson` [*indent*] *value*
+
+`toPrettyJson` returns the JSON representation of *value*. The optional *indent*
+specifies how much nested elements are indented relative to their parent.
+*indent* defaults to two spaces.
+
+!!! examples
+
+    ```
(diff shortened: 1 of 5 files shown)


felipecrs:
This is awesome! I tested it too, and it works perfectly.

twpayne:
Thank you again for the testing!

PR #2756, opened 2023-02-11
chore: Use strings.Cut{Prefix,Suffix}

twpayne:


halostatue:
It’s *always* nice to be able to remove code.

Issue #1605, opened 2021-11-09
upgrade command should work on Windows


twpayne:
## Is your feature request related to a problem? Please describe.

Currently, `chezmoi upgrade` only works on a few operating systems, and notably it is [completely disabled on Windows](https://github.com/twpayne/chezmoi/blob/83bad0ac556fae34688e9db3821f55f031745d36/internal/cmd/noupgradecmd.go#L1).

## Describe the solution you'd like

`chezmoi upgrade` should work on Windows, just as it does on UNIX-like operating systems.

## Describe alternatives you've considered

Leaving things as they are.

## Additional context

@felipecrs and I discussed this earlier today. Right now, there are a few complications for making `chezmoi upgrade` work on Windows:

1. As I currently understand it (and I might be wrong), Windows does not allow you to replace a file while it is opened, which means that a chezmoi binary cannot replace it itself while it is running without some extra external mechanism.
2. Detecting exactly how chezmoi was installed on Windows is not obvious. For example, chezmoi might be installed via @zb140's [PowerShell script](https://github.com/twpayne/chezmoi/blob/master/assets/scripts/install.ps1), via [Chocolatey](https://github.com/twpayne/chezmoi/pull/759), via [Scoop](https://github.com/twpayne/scoop-bucket/blob/master/chezmoi.json), or via some other method.

@felipecrs points out that the `deno upgrade` command works around problem 1 above using [this logic](https://github.com/denoland/deno/blob/22dd1b90c4eb5671decd458cc70935898f9ccf0d/cli/tools/upgrade.rs#L286-L298). Maybe chezmoi can use the same?

cc @zb140 as the Windows expert :)

felipecrs:
Another option would be to delegate the replacing part to a PowerShell or CMD process, so that chezmoi.exe would not be in use during the replacing. 

It's possible to do such kind of thing (replace current process with another process, thus closing the current process) in Linux in go with [`syscall.Exec`](https://stackoverflow.com/a/17356031/12156188), but I don't know how it would be done on Windows.

zb140:
Yeah, it would be great to make this work.  When I first started filling out the Windows support I intended to do it, but by the time I got most of the rest of the features working it had sort of fallen by the wayside :smile:

[...]

PR #2757, opened 2023-02-11
fix: Don't use `replace-executable` for WinGet installations

### LINKED ISSUE #2736: Add winget support to `chezmoi upgrade`

The [`chezmoi upgrade` command](https://www.chezmoi.io/reference/commands/upgrade/) attempts to detect how chezmoi was installed and then use the same mechanism to upgrade chezmoi. For example, if the user installed chezmoi with their package manager, then `chezmoi upgrade` should use the same package manager to install a new version of chezmoi.

Currently, when chezmoi is installed with winget, `chezmoi doctor` [reports](https://github.com/twpayne/chezmoi/pull/2735#issuecomment-1416390837) that `chezmoi upgrade` should use the generic `replace-executable` method to upgrade chezmoi.

##  Describe the solution you'd like

When chezmoi is installed with a winget package (#2735), then `chezmoi upgrade` should use winget to install the latest available version of chezmoi.

Keeping this as-is.

## Additional context

There are two parts:

1. Detecting if chezmoi was installed with winget. The [current heuristics](https://github.com/twpayne/chezmoi/blob/213bac53127a41c9910186b498e9ffefdcf8ed5b/pkg/cmd/upgradecmd.go#L471-L538) look at the path to the executable and file ownership. These would need to be extended to detect winget installations of chezmoi.
2. Upgrading chezmoi via winget in this case. The [current code](https://github.com/twpayne/chezmoi/blob/213bac53127a41c9910186b498e9ffefdcf8ed5b/pkg/cmd/upgradecmd.go#L166-L192) would need to be extended to call winget to upgrade chezmoi.

@bradenhilton would you like to tackle this? Please assign yourself to this issue if/when you want to take it.

bradenhilton:
Fixes #2736

Very basic, not ready for merge yet.

### DIFF
--- a/internal/cmd/upgradecmd_windows.go
+++ b/internal/cmd/upgradecmd_windows.go
@@ -1,4 +1,4 @@
-//go:build !noupgrade && !windows
+//go:build !noupgrade
 
 package cmd
 
@@ -15,6 +15,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strings"
@@ -29,59 +30,33 @@ import (
 )
 
 const (
-	upgradeMethodBrewUpgrade       = "brew-upgrade"
 	upgradeMethodReplaceExecutable = "replace-executable"
-	upgradeMethodSnapRefresh       = "snap-refresh"
-	upgradeMethodUpgradePackage    = "upgrade-package"
-	upgradeMethodSudoPrefix        = "sudo-"
-
-	libcTypeGlibc = "glibc"
-	libcTypeMusl  = "musl"
-
-	packageTypeNone = ""
-	packageTypeAPK  = "apk"
-	packageTypeAUR  = "aur"
-	packageTypeDEB  = "deb"
-	packageTypeRPM  = "rpm"
+	upgradeMethodWinGetUpgrade     = "winget-upgrade"
 )
 
-var (
-	packageTypeByID = map[string]string{
-		"alpine":   packageTypeAPK,
-		"amzn":     packageTypeRPM,
-		"arch":     packageTypeAUR,
-		"centos":   packageTypeRPM,
-		"fedora":   packageTypeRPM,
-		"opensuse": packageTypeRPM,
-		"debian":   packageTypeDEB,
-		"rhel":     packageTypeRPM,
-		"sles":     packageTypeRPM,
-		"ubuntu":   packageTypeDEB,
-	}
-
-	archReplacements = map[string]map[string]string{
-		packageTypeDEB: {
-			"386": "i386",
-			"arm": "armel",
-		},
-		packageTypeRPM: {
-			"amd64": "x86_64",
-			"386":   "i686",
-			"arm":   "armhfp",
-			"arm64": "aarch64",
-		},
-	}
-
-	checksumRx      = regexp.MustCompile(`\A([0-9a-f]{64})\s+(\S+)\z`)
-	libcTypeGlibcRx = regexp.MustCompile(`(?i)glibc|gnu libc`)
-	libcTypeMuslRx  = regexp.MustCompile(`(?i)musl`)
-)
+var checksumRx = regexp.MustCompile(`\A([0-9a-f]{64})\s+(\S+)\z`)
 
 type upgradeCmdConfig struct {
 	executable string
 	method     string
 	owner      string
 	repo       string
+}
+
+type InstallBehavior struct {
+	PortablePackageUserRoot    string `json:"portablePackageUserRoot"`
+	PortablePackageMachineRoot string `json:"portablePackageMachineRoot"`
+}
+
+func (ib *InstallBehavior) Values() []string {
+	return []string{
+		ib.PortablePackageUserRoot,
+		ib.PortablePackageMachineRoot,
+	}
+}
+
+type WinGetSettings struct {
+	InstallBehavior InstallBehavior `json:"installBehavior"`
 }
 
 func (c *Config) newUpgradeCmd() *cobra.Command {
@@ -176,26 +151,12 @@ func (c *Config) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 
 	// Replace the executable with the updated version.
 	switch method {
-	case upgradeMethodBrewUpgrade:
-		if err := c.brewUpgrade(); err != nil {
-			return err
-		}
 	case upgradeMethodReplaceExecutable:
 		if err := c.replaceExecutable(ctx, executableAbsPath, version, rr); err != nil {
 			return err
 		}
-	case upgradeMethodSnapRefresh:
-		if err := c.snapRefresh(); err != nil {
-			return err
-		}
-	case upgradeMethodUpgradePackage:
-		useSudo := false
-		if err := c.upgradePackage(ctx, version, rr, useSudo); err != nil {
-			return err
-		}
-	case upgradeMethodSudoPrefix + upgradeMethodUpgradePackage:
-		useSudo := true
-		if err := c.upgradePackage(ctx, version, rr, useSudo); err != nil {
+	case upgradeMethodWinGetUpgrade:
+		if err := c.winGetUpgrade(); err != nil {
 			return err
 		}
 	default:
@@ -218,10 +179,6 @@ func (c *Config) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 	chezmoiVersionCmd.Stdout = os.Stdout
 	chezmoiVersionCmd.Stderr = os.Stderr
 	return chezmoilog.LogCmdRun(chezmoiVersionCmd)
-}
-
-func (c *Config) brewUpgrade() error {
-	return c.run(chezmoi.EmptyAbsPath, "brew", []string{"upgrade", c.upgrade.repo})
 }
 
 func (c *Config) getChecksums(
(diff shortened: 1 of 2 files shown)


pkg/cmd/upgradecmd.go
```diff
@@ -476,6 +485,8 @@ func getUpgradeMethod(fileSystem vfs.Stater, executableAbsPath chezmoi.AbsPath)
 		return upgradeMethodBrewUpgrade, nil
 	case runtime.GOOS == "linux" && strings.Contains(executableAbsPath.String(), "/.linuxbrew/"):
 		return upgradeMethodBrewUpgrade, nil
+	case runtime.GOOS == "windows" && strings.Contains(executableAbsPath.String(), "/WinGet/"):
```
twpayne, line 488: This might need a case-insensitive match.

bradenhilton: It's matching `C:\\<username>\\AppData\\Local\\Microsoft\\WinGet\\Links\\chezmoi.exe`.

I just realised it doesn't actually match with `/WinGet/`, it matches with `\\WinGet\\`.

twpayne:
This looks very reasonable. What are the remaining steps needed to make this ready to merge?

bradenhilton:
At a minimum, I think we need to wait for go1.21 (expected August) for the necessary fix to be released.

I'd also like to figure out a nicer way to test, as I was doing everything manually when writing the code.

bradenhilton:
Some general thoughts on how this can be improved:

WinGet has configurable root directories for portable apps, for both user and machine scopes:
- [Portable Package User Root](https://github.com/microsoft/winget-cli/blob/2537239ba22091803118de71ced0e5d4571be20b/doc/Settings.md#portable-package-user-root), defaults to `%LOCALAPPDATA%\Microsoft\WinGet\Packages\`[^1][^2]
- [Portable Package Machine Root](https://github.com/microsoft/winget-cli/blob/2537239ba22091803118de71ced0e5d4571be20b/doc/Settings.md#portable-package-machine-root), defaults to `%PROGRAMFILES%\WinGet\Packages\`[^1][^2]
- [Default install root](https://github.com/microsoft/winget-cli/blob/2537239ba22091803118de71ced0e5d4571be20b/doc/Settings.md#default-install-root) (unsure at the moment if this affects portable apps)

[^1]: The defaults only apply if no corresponding value is specified, they are not set by default in the `settings.json` file.

[^2]: The default values must be expanded, non-default values must be absolute paths so expanding is not necessary.

These paths can be set in WinGet's `settings.json` file, the location of which differs depending on how WinGet was installed:
- `%LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json` if packaged, most users will likely have this
- `%LOCALAPPDATA%\Microsoft\WinGet\Settings\settings.json` if non-packaged or built from source

The absolute executable path reported by chezmoi is likely to be a symlink when installed via WinGet (`~\AppData\Local\Microsoft\WinGet\Links\chezmoi.exe`[^3])

A more robust solution would employ an approach like this:
- Resolve the executable symlink to its target path (typically `~\AppData\Local\Microsoft\WinGet\Packages\twpayne.chezmoi_Microsoft.Winget.Source_8wekyb3d8bbwe\chezmoi.exe`[^3] with default settings), if necessary
- Read and unmarshal WinGet's `settings.json`, if found
- Check if the real executable path contains any of the paths present in `settings.json`, or their default values if not specified

The current solution in this PR only really handles the default path(s), which is probably acceptable for a first pass. That said, one could run `install.ps1` to install chezmoi to `~\WinGet\`[^3] (which is admittedly a little silly, but definitely possible)[^4], and this would definitely confuse chezmoi with the current approach, so it should be improved at some point to support as many users as is feasible.

[^3]: `~` used for shorthand, not to imply it is a valid absolute path.

[^4]: Granted, one could also use `install.ps1` to install chezmoi to the actual WinGet package directory, without installing chezmoi via WinGet. I don't think there exists a perfect solution to this.

I think I'm right at my limit when it comes to Go, so if it gets any more complicated than this, I might not be able to work on it without some more learning.

bradenhilton:
https://github.com/twpayne/chezmoi/pull/2757/commits/784ec9a3f26ec6e1dc62b0be46179dcb7ecdc31f is my current snapshot. It's *super* rough and definitely needs more refinement, but the inability to test properly without go1.21 makes it rather tedious at the moment.

twpayne:
Thanks very much for the updates. Happy to wait for Go 1.21 for this.

bradenhilton:
Just FYI, if/when anyone feels like it, I'd absolutely welcome a review on the file unmarshalling, logic, flow, and error handling, as I'm not particularly well versed in these at the moment.

I attempted to map it out as:

- The default path values are used unless the `settings.json` file both exists **and** contains the appropriate value(s).
- The `settings.json` file existing but being somehow unreadable at the time of unmarshal means we should probably exit early, as we can't determine if the file contains the value(s) or not.
- If the `settings.json` file and value(s) exist, but we can't find a chezmoi binary there, we should probably also exit.

I'm reasonably confident that the command invocation works, or at least it did when I built go from source to test it a few months ago.

I think treating WinGet errors as chezmoi errors should be fine for now. We can absolutely handle them in a follow up PR if needed.

internal/cmd/upgradecmd.go
```diff
@@ -634,3 +651,66 @@ func getReleaseAssetByName(rr *github.RepositoryRelease, name string) *github.Re
...
+	}
+	if fi.Mode().Type() == os.ModeSymlink {
+		realExecutableAbsPath, err = os.Readlink(executableAbsPath)
+		if err != nil {
+			return false, err
+		}
+	}
+	winGetSettings := WinGetSettings{
+		InstallBehavior: InstallBehavior{
+			PortablePackageUserRoot:    os.ExpandEnv("${LOCALAPPDATA}\\Microsoft\\WinGet\\Packages\\"),
```
twpayne, line 688: Quick tip: You can use backquoted strings to avoid having to escape all the backslashes, e.g.

```go
os.ExpandEnv(`${LOCALAPPDATA}\Microsoft\WinGet\Packages\`)
```

bradenhilton:
Preface: In my working copy I have split the Windows-specific upgrade behavior into `upgradecmd_windows.go`, and doing this may have introduced bugs I am not aware of. Regardless of what I'm about say, I will still commit this code in a snapshot.

I don't think this is possible.

Firstly, this is just awful to test. I'm open to any ideas anyone might have to make it less painful.

I created a build from my branch with version `2.37.100` and dropped it into the WinGet install directory, this causes `chezmoi upgrade` to error out via WinGet:

```console
❯ chezmoi upgrade
Found chezmoi [twpayne.chezmoi] Version 2.38.0
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Successfully verified installer hash
Extracting archive...
Successfully extracted archive
Starting package uninstall...
Unable to remove Portable package as it has been modified; to override this check use --force
chezmoi: exit status 0x8a150057
```

After learning this, I created two more chezmoi builds (as `chezmoi-local` to avoid conflicts) with versions `2.37.100` and `2.38.0`. Then I created custom local WinGet manifests for them which download the zips from my Dropbox (this gives several anti-virus false positives which I had to manually allow in order to proceed).

Here is `chezmoi-local doctor` with version `2.37.100`:

```console
❯ chezmoi-local doctor
ok        version              v2.37.100, commit 5682362fb54c388a58ee7af9384dd8f19983a355-dirty, built at 2023-08-23T12:49:15Z
warning   latest-version       v2.38.0
ok        systeminfo           Microsoft Windows 10 Pro (10.0.19045 N/A Build 19045)
ok        go-version           go1.21.0 (gc)
ok        executable           ~/AppData/Local/Microsoft/WinGet/Links/chezmoi-local.exe
ok        upgrade-method       winget-upgrade
...
```

And here is `chezmoi-local upgrade`, again, version `2.37.100`, in which I hard coded the path to my local manifest for `2.38.0`:

```console
❯ chezmoi-local upgrade
Found chezmoilocal [bradenhilton.chezmoilocal] Version 2.38.0
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://<dropbox-url>
  ██████████████████████████████  20.0 MB / 20.0 MB
Successfully verified installer hash
Extracting archive...
Successfully extracted archive
Starting package uninstall...
An unexpected error occurred while executing the command:
remove: unknown error: "~\AppData\Local\Microsoft\WinGet\Packages\bradenhilton.chezmoilocal__DefaultSource\chezmoi-local.exe"
Uninstall failed with exit code: 0x8a150003 : Executing command failed
```

I'm not certain, but I believe this error is given because chezmoi is running at the time of upgrade/uninstall.

twpayne:
Thank you very much for working on this!



> I'm not certain, but I believe this error is given because chezmoi is running at the time of upgrade/uninstall.

Previously I've worked around this by renaming (but not deleting) the current executable. See https://github.com/twpayne/chezmoi/issues/1605#issuecomment-964698709.

bradenhilton:
> Previously I've worked around this by renaming (but not deleting) the current executable. See [#1605 (comment)](https://github.com/twpayne/chezmoi/issues/1605#issuecomment-964698709).

I don't know how to incorporate this into the upgrade flow for WinGet.

I haven't checked yet, but I'm pretty sure `chezmoi.exe` needs to exist at time of removal, and it needs to be the exact same `chezmoi.exe` that WinGet installed.

twpayne:
Thanks again for all your work here. It might well be the case that this self-upgrade through winget functionality is simply not possible to achieve on Windows. Feel free to close this and #2736 if needed.

bradenhilton:
I'm inclined to agree, but I still think we need to do *something*, as chezmoi currently thinks it can use `replace-executable` with WinGet installs, which will break them.

Should I change `c.winGetUpgrade()` to print a simple message informing the user that `chezmoi upgrade` is not currently supported for WinGet, along with the command to upgrade chezmoi via WinGet itself? We could also do this for Scoop and Chocolatey in follow up PRs if needed.

I'm thinking the commit message can be changed to `fix: Don't use replace-executable for WinGet installations` or similar.

We could also consider releasing new Windows binaries specifically for use with package managers which are built with `noupgrade`, but it seems a bit excessive.

twpayne:
> Should I change `c.winGetUpgrade()` to print a simple message informing the user that `chezmoi upgrade` is not currently supported for WinGet, along with the command to upgrade chezmoi via WinGet itself? We could also do this for Scoop and Chocolatey in follow up PRs if needed.

Yes, this sounds perfect.



> We could also consider releasing new Windows binaries specifically for use with package managers which are built with `noupgrade`, but it seems a bit excessive.

Agreed that this would be excessive at the moment.

bradenhilton:
FYI, if we *aren't* invoking `winget.exe` (which is currently the case), I don't think we require the changes that were released in go1.21.

If we needed to invoke `winget.exe`, that could technically mean we need to wait for go1.21 to be our *minimum* supported version, which would mean waiting for go1.22.

twpayne:
OK. We can go with this as the "official" versions of chezmoi are built with Go 1.21. Anyone building chezmoi with Go 1.20 will not need the upgrade command.

I'll merge this now and just do a little tidy-up to reduce the code duplication between `upgradecmd.go` and `upgradecmd_windows.go`.

PR #2773, opened 2023-02-14
feat: Find config file with $CHEZMOI_CONFIG

### LINKED ISSUE #2771: Set global flags using environment variables

When working with different chezmoi source directories on the same machine it is useful to be able to run a chezmoi command that is influenced by environment variables. For instance, `direnv` can be used to set different "cache | config | source" directory environment variables that chezmoi will use when executed.

This allows one to cd to alternative source directories and being allowed to use `chezmoi diff`, `chezmoi apply`, etc. while in the alternative directory and actually use that same source directory as a source path, and other alternative directories for the cache and configuration. This will prevent the necessity to always use `--cache` | `--config` | `--source` with every invocation of chezmoi when using alternative directories.

Have chezmoi obey environment variables to influence the used cache | config | source directories. For instance `CHEZMOI_CACHE_PATH` | `CHEZMOI_CONFIG_PATH` | `CHEZMOI_SOURCE_PATH`.

Alternatives are to add a local `.bin` directory to `$PATH` with a script called `chezmoi` that executes chezmoi with the alternative directories set as arguments and then have `direnv` add that `.bin` directory automatically upon entering a chezmoi source directory. Works, but using environment variables are more suited for this scenario and easier to maintain.

halostatue:
Closes #2771

This provides an alternative to pure command-line flag configuration when working with multiple source trees.

### DIFF
--- a/assets/chezmoi.io/docs/reference/command-line-flags/global.md
+++ b/assets/chezmoi.io/docs/reference/command-line-flags/global.md
@@ -2,9 +2,13 @@
 
 ## `--cache` *directory*
 
+> Configuration: `cacheDir`
+
 Use *directory* as the cache directory.
 
 ## `--color` *value*
+
+> Configuration: `color`
 
 Colorize diffs, *value* can be `on`, `off`, `auto`, or any boolean-like value
 recognized by `promptBool`. The default is `auto` which will colorize diffs only
@@ -12,7 +16,7 @@ if the environment variable `$NO_COLOR` is not set and stdout is a terminal.
 
 ## `-c`, `--config` *filename*
 
-Read the configuration from *filename*.
+Read the [configuration](/reference/configuration-file) from *filename*.
 
 ## `--config-format` `json`|`jsonc`|`toml`|`yaml`
 
@@ -21,6 +25,8 @@ the config filename does not have an extension, for example when it is
 `/dev/stdin`.
 
 ## `-D`, `--destination` *directory*
+
+> Configuration: `destDir`
 
 Use *directory* as the destination directory.
 
@@ -56,6 +62,8 @@ Write the output to *filename* instead of stdout.
 
 ## `--persistent-state` *filename*
 
+> Configuration: `persistentState`
+
 Read and write the persistent state from *filename*. By default, chezmoi stores
 its persistent state in `chezmoistate.boltdb` in the same directory as its
 configuration file.
@@ -82,9 +90,13 @@ if no cached external is available.
 
 ## `-S`, `--source` *directory*
 
+> Configuration: `sourceDir`
+
 Use *directory* as the source directory.
 
 ## `--use-builtin-age` *value*
+
+> Configuration: `useBuiltinAge`
 
 Use chezmoi's builtin [age encryption](https://age-encryption.org) instead of an
 external `age` command. *value* can be `on`, `off`, `auto`, or any boolean-like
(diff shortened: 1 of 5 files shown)


halostatue:
This needs further work, especially on documentation and tests. I have yet had no time to consider what testing this might look like.

twpayne:
I think that the code in this PR will cause the values read from the environment variables to be overwritten with values from the configuration file. I think the hook needs to happen at the end of `./pkg/cmd.Config.readConfig`.

halostatue:
> I think that the code in this PR will cause the values read from the environment variables to be overwritten with values from the configuration file. I think the hook needs to happen at the end of `./pkg/cmd.Config.readConfig`.

It’s possible — even probable. I think that will require some additional thinking and tracking to identify where the value came from. On the other hand, if most of these values could be found in a configuration file, then the only thing that might be necessary is `$CHEZMOI_CONFIG` — and the recommendation is that the config file has the necessary configuration to override everything else. This was mostly pushed *for the moment* to start a code discussion. I’ll look at refining it over the next day or two.

halostatue:
I’m reworking this so that some of the variable names better match the corresponding command-line flags and configuration (instead of `CHEZMOI_CACHE_PATH`, it will be `CHEZMOI_CACHE_DIR`, etc.).

Some inconsistencies as I’m looking through this:

- `--persistent-state` does not have a corresponding value in the config file.
  - Should `persistentState` be added to the config file? If not, should we remove `$CHEZMOI_PERSISTENT_STATE`?
- `scriptTempDir` does not have a corresponding command-line flag.
  - Should one be added? Should we have `$CHEZMOI_SCRIPT_TEMP_DIR`?

Since most of the flags are available in the configuration file, I think we should mark the following flags as deprecated and remove them in Chezmoi v3:

- `--cache` (use `cacheDir` instead)
- `--destination` (use `destDir` instead)
- `--persistent-state` (add `persistentState` to configuration file)
- `--source` (use `sourceDir` instead)
- `--working-tree` (use `workingTree` instead)

With the deprecation of these flags, we can change this to only support `$CHEZMOI_CONFIG`, which could be used instead of `--config`.

What do you think @twpayne?

halostatue:
## feat: Find config file with $CHEZMOI_CONFIG

Resolves #2771

When `$CHEZMOI_CONFIG` to be set, chezmoi will act as if it were called with `chezmoi --config $CHEZMOI_CONFIG`. If both `$CHEZMOI_CONFIG` and `--config PATH-TO-CONFIG` are provided, the command-line flag wins.

## doc: Show config vars for global flags

## doc: Add JSONC configuration example

## doc: Improved information for config vars

twpayne:
Sorry, I am against this PR.

Currently chezmoi **sets** a wide range of environment variables for processes that it spawns. All of these variables begin with `CHEZMOI`.

This PR adds a special environment variable which is the first and only one to be **read** by chezmoi. This is inconsistent with chezmoi's existing use of environment variables.

I don't like configuration being set via environment variables as they are generally not trivially visible to the user (unlike, say command line arguments). I think the original problem is better solved with a wrapper script or alias.

halostatue:
OK. Let me see what we can discuss on this:

1. There are some (I think good) improvements to the documentation. Should I open a PR for those (and rework the bits out which refer to `$CHEZMOI_CONFIG`)?
2. Would you accept a PR for *improved* wrapper script examples that include a `$CHEMOI_CONFIG` reader so that such a script can be responsive to `direnv` configuration?

That latter part is sort of what would be important to me—and why I thought this would be a worthwhile, as I regularly have `.envrc` files that specify `AWS_PROFILE` specific to that directory for running `terraform`, for example. That is, I don’t *want* a `czb` wrapper function / script; I just want to remember `chezmoi apply`.

twpayne:
> * There are some (I think good) improvements to the documentation. Should I open a PR for those (and rework the bits out which refer to `$CHEZMOI_CONFIG`)?
> * Would you accept a PR for _improved_ wrapper script examples that include a `$CHEMOI_CONFIG` reader so that such a script can be responsive to `direnv` configuration?

Yes, both would be fantastic, thank you!

twpayne:
Superseded by #2843.

PR #2792, opened 2023-02-24
feat: Add Dashlane password manager support

### LINKED ISSUE #2791: Add Dashlane support
[Dashlane](https://www.dashlane.com/) is a password manager that could be integrated with chezmoi.
This was initially open then closed-not-solved in #1037.

It would be great if chezmoi include a template function called `dashlane` that allowed users to retrieve secrets from Dashlane.
It could even be `dashlanePassword` for a password entry, and `dashlaneNote` for a secure note.

## Additional context
Since #1037 there is now an experimental-but-official CLI for Dashlane which is available at [Dashlane/dashlane-cli](https://github.com/Dashlane/dashlane-cli). This supports outputting passwords in the JSON format, and can be either built locally or used through binaries provided with releases. This is also seems to be actively developed at the moment.

twpayne:
Fixes #2791.

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -101,6 +101,13 @@ sections:
     custom:
       type: bool
       description: Enable custom shell completions
+  dashlane:
+    args:
+      type: '[]string'
+      description: Extra args to Dashlane CLI command
+    command:
+      default: '`dcli`'
+      description: Dashlane CLI command
   diff:
     args:
       type: '[]string'
--- a/assets/chezmoi.io/docs/reference/templates/dashlane-functions/dashlanePassword.md
+++ b/assets/chezmoi.io/docs/reference/templates/dashlane-functions/dashlanePassword.md
@@ -1,0 +1,15 @@
+# `dashlanePassword` *filter*
+
+`dashlanePassword` returns structured data from [Dashlane](https://dashlane.com)
+using the [Dashlane CLI](https://github.com/Dashlane/dashlane-cli) (`dcli`).
+*filter* is passed to `dcli password --output json`, and the output from `dcli
+password` is parsed as JSON.
+
+The output from `dcli password` cached so calling `dashlanePassword` multiple
+times with the same *filter* will only invoke `dcli password` once.
+
+!!! example
+
+    ```
+    {{ (index (dashlanePassword "filter") 0).password }}
+    ```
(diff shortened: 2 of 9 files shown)


twpayne:
@XaF this should be ready now. Would you be able to test it?

XaF:
So ran this:

```
./chezmoi execute-template '{{ (index (dashlanePassword "MyTestFilter") 0).password }}'
```

Which does work and return the password correctly! 🎉 Thank you!

However, I'd have suggestions and would love to know what you'd feel about:
- `dashlanePassword` does what you just built
- `dashlanePasswordFirst` returns the map entry for the first matching entry (can it be erroring if more than one match ?)
- `dashlanePasswordValue` returns the password for the first matching entry (can it be erroring if more than one match ?), allowing to avoid having to index + get the password field
- `dashlaneNote` returns the secure note matching the filter (this is in a separate place in dashlane)

If that helps, I'm also happy to make a PR targeting your PR for those changes! :)

twpayne:
Thanks very much for testing!

I'd love a PR that implements `dashlaneNote`.

For `dashlanePasswordFirst` and `dashlanePasswordValue`, I'm not so sure, for two reasons:

1. I prefer to provide low-level components that the user can assemble to achieve the effect they want, rather than high-level components that do multiple things. This maximizes flexibility and minimizes the number of components that need to be maintained.

2. `dashlanePasswordFirst` only makes sense if the Dashlane CLI returns entries in a deterministic order. Otherwise, whichever entry appears first will vary from time to time, making `dashlanePasswordFirst` sporadically unreliable.

Thinking about this a bit more, do you know what exactly *filter* matches on when you call `dcli password *filter*`? Does it match on title, id, both, or something else? It might make more sense to have a function like `dashlaneLookup *id*` that calls `dcli password --output json *id*` and then returns the element of the returned array with the given *id*. This should be robust. What are your thoughts?

XaF:
Yeah I guess `First` is not the right word, I meant something like "One" or "Unique" or any other word that would make more sense:
- if 0 are matching, error
- if >1 are matching, error
- if 1 is matching, return it

That shouldn't be an issue with the ordering of things, if we only match when a unique one matches.

The `filter` field matches on `url` and `title` as you can read here: https://github.com/Dashlane/dashlane-cli/blob/master/src/middleware/getPasswords.ts#L63-L67
Which means that even having the unique ID is not going to help... unless we're ready to filter things ourselves? But from your comments I (understandably) don't think you'd want some of that logic in `chezmoi` (and I do agree that logic should exist in the client).

twpayne:
I'll merge this now. Please do submit a PR implementing `dashlaneNote` against chezmoi's `master` branch.

PR #2798, opened 2023-02-27
correct configuration file syntax for edit.args

nate-duke:
Slight tweak to reference documentation to correct editor argument key.

Pretty sure this is right. It's at least what worked for me on
```bash
❯ chezmoi --version
chezmoi version v2.31.0, commit 4d2bc846212e27fae1e5bbd45d70e00908da603b, built at 2023-02-17T11:19:43Z, built by Homebrew
```


twpayne:
Good catch, thank you!

PR #2801, opened 2023-03-01
chore: Add post-apply hook test

### LINKED ISSUE #2799: Possible to run scripts on "add" or "re-add"
Hi,

Currently I have a few (not many, but a few) folders with lots of files. The actual important files are not many, but I need all the folder. Adding this folder/managing it, is somewhat slow with chezmoi.

One idea to work around this would be to have scripts manage some folders with rsync.

If it would be possible to have scripts run when "add"ing new files or "re-add"ing new files. With those scripts I could use the rsync functionality to either sync the "source" folder to "target" (which already works with apply), and sync the "target" folder to "source", by syncing with "add" or "re-add".

Example:

* Managing logseq files and config: I could manage the logseq user files and preferences, but use rsync to sync all the plugins/etc files that take a long time in chezmoi.

Currently  I am just adding everything into managed state, which is kinda slow and unnecessary, since those files aren't important to be "managed by git", although I would like to already have them there.

## Additional context

Maybe add a mechanism to know if the script would be ran as part of "apply" or "add"/"re-add".

As an ending note, in this issue I am thinking of using script to manage rsync in both ways.... but it could be probably be useful for other situations!

Thank you!

twpayne:
Fixes #2799.

### DIFF
--- a/pkg/cmd/testdata/scripts/issue2799.txtar
+++ b/pkg/cmd/testdata/scripts/issue2799.txtar
@@ -1,0 +1,14 @@
+[windows] skip 'UNIX only'
+
+# test that chezmoi can run hooks after certain commands
+exec chezmoi diff
+! stdout .
+exec chezmoi apply
+stdout 'apply hook'
+
+-- home/user/.local/share/chezmoi/run_after_hook.sh.tmpl --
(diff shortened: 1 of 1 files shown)


twpayne:
With hindsight, this is not a general solution.

Issue #2820, opened 2023-03-05
Chezmoi test case for #2818


EvanCarroll:
Follow up on #2818 

I'll just keep this technical here is a test case,

```text
cat <<EOF | podman run -i alpine:3 /bin/sh -

  apk update;
  apk add chezmoi git;

  echo "THIS IS OLD" > ~/.profile;
  cat ~/.profile | sed -e 's/^/CHEZMOI PRE_APPLY:\t/'

  echo "Running chezmoi init"
  chezmoi init "https://github.com/EvanCarroll/dotfiles_chezmoi.git";

  echo "DEBUG: Running - chezmoi apply"
  echo "DEBUG: Let's check out the profile, still old..."
  chezmoi apply; 
 
  echo "Going back to chezmoi dir"
  chezmoi cd;

  # Let's go to git
  cat ~/.profile | sed -e 's/^/[CHEZMOI_DIR POST_APPLY]:\t/'

EOF
```

In the above case, when it prints the `CHEZMOI_DIR POST_APPLY` copy, I would expect that to be the actual copy on chezmoi repo.. which is this

* https://github.com/EvanCarroll/dotfiles_chezmoi/blob/main/dot_profile

Now, I can tell what's happening from further investigation, there is a conflicting dotfile which is here

* https://github.com/EvanCarroll/dotfiles_chezmoi/blob/main/private_dot_local/private_share/chezmoi/literal_dot_profile

This is weird though because when I'm on the other machine that I used to create this repo, and I do `chezmoi cd; git status` it shows "nothing to commit, working tree is clean". So I take the other machine is in sync, and the `~/.profile` there is, as confirmed, the same as the one above at:

* https://github.com/EvanCarroll/dotfiles_chezmoi/blob/main/dot_profile

However, when I do a subsequent `chezmoi init; chezmoi apply` on the repo in question the file is not 

* https://github.com/EvanCarroll/dotfiles_chezmoi/blob/main/dot_profile

But instead,

* https://github.com/EvanCarroll/dotfiles_chezmoi/blob/main/private_dot_local/private_share/chezmoi/literal_dot_profile

I only see [`literal` documented here,](https://www.chezmoi.io/reference/source-state-attributes/)

>The literal_ prefix and .literal suffix can appear anywhere and stop attribute parsing. This permits filenames that would otherwise conflict with chezmoi's attributes to be represented.

And I have no idea how that file got created. I can create a container of the other machine that created these files too to show you what's going on there.

[...]

PR #2822, opened 2023-03-05
feat: Refuse to add chezmoi files to chezmoi

twpayne:
Refs #2820.

@bradenhilton I'd appreciate your thoughts on this. Right now, this stops `chezmoi add` from adding any path where `chezmoi` is one of the components, but it's a pretty simple check and there are probably improvements/fixes.

### DIFF
--- a/pkg/cmd/testdata/scripts/issue1832.txtar
+++ b/pkg/cmd/testdata/scripts/issue1832.txtar
@@ -3,7 +3,7 @@
 # test that chezmoi add succeeds when changing the permissions of an intermediate directory
 exec chezmoi add $HOME/.config/fish/config
 chmod 700 $HOME/.config
-exec chezmoi add --force $HOME/.config
+exec chezmoi add --force --recursive=false $HOME/.config
 
 -- home/user/.config/fish/config --
 # contents of .config/fish/config
--- a/pkg/chezmoi/sourcestate.go
+++ b/pkg/chezmoi/sourcestate.go
@@ -284,18 +284,19 @@ type ReplaceFunc func(targetRelPath RelPath, newSourceStateEntry, oldSourceState
 
 // AddOptions are options to SourceState.Add.
 type AddOptions struct {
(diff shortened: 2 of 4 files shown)


EvanCarroll:
This certainly seems like it would make chezmoi far easier to reason about, and less error prone.

pkg/cmd/testdata/scripts/issue1832.txtar
```diff
@@ -1,9 +1,9 @@
 [windows] skip 'UNIX only'
 
 # test that chezmoi add succeeds when changing the permissions of an intermediate directory
-exec chezmoi add $HOME/.config/fish/config
-chmod 700 $HOME/.config
```
twpayne, line 5: ...after `chezmoi add`, chezmoi has created its state in `.config/chezmoi/chezmoistate.boltdb`, so this line was adding a chezmoi dotfile to chezmoi. Changing the directory name in the test from `.config` to `.dir` was the easy fix.

bradenhilton:
@twpayne Just wanted to first check if you had any thoughts on https://github.com/twpayne/chezmoi/issues/2820#issuecomment-1455191338?

twpayne:
> @twpayne Just wanted to first check if you had any thoughts on [#2820 (comment)](https://github.com/twpayne/chezmoi/issues/2820#issuecomment-1455191338)?

Good point. We should also check for files in the config directory. I'll update this PR.

halostatue:
Why not check for partial matches of `cacheDir`, `persistentState`, `sourceDir` or `workingTree` instead of just looking for `chezmoi`?

If I specify `--config`, I can specify something that does not include `chezmoi`—which means that this check does not help in that case, or if we add `CHEZMOI_CONFIG` support as suggested in #2773.

felipecrs:
> Why not check for partial matches of `cacheDir`, `persistentState`, `sourceDir` or `workingTree` instead of just looking for `chezmoi`?

That's exactly what I thought when I first read this PR as well.

halostatue (approved):

PR #2827, opened 2023-03-06
feat: Add support for the JSONC format

### LINKED ISSUE #2825: Support JSON with Comments in `fromJson` function

I'm trying to use modify templates to handle both my VS Code settings.json and also my Windows Terminal settings.json. However, their formats are JSON with Comments, which differs from regular JSONs a little.

Currently, if there are comments in the JSON, the `fromJson` will silently load `no value`. Effectively, at the end of the modify template, it will basically clean everything up that was in the file previously.

```console
$ printf "%s\n" "{" "  // This is a comment" '  "a": "whatever",' "}" | 
  chezmoi execute-template --with-stdin '{{ fromJson .chezmoi.stdin }}'
<no value>
```

I believe it would be nice to make fromJson accept JSON with Comments as well (which is a superset of JSON), or, a new function like `fromJsonc`.

I have a `.chezmoiscript` that backs up my configuration before actually changing it so that if something wrong happens, I can still recover.

## Additional context

I found this library for Go which seems to be actively maintained:

- https://github.com/tailscale/hujson
- https://pkg.go.dev/github.com/tailscale/hujson#section-readme
- https://ianmuchina.com/blog/13-json-golang/

twpayne:
Fixes #2825.

@felipecrs can you test this?

### DIFF
--- a/pkg/chezmoi/format.go
+++ b/pkg/chezmoi/format.go
@@ -6,14 +6,16 @@ import (
 	"strings"
 
 	"github.com/pelletier/go-toml/v2"
+	"github.com/tailscale/hujson"
 	"gopkg.in/yaml.v3"
 )
 
 // Formats.
 var (
-	FormatJSON Format = formatJSON{}
-	FormatTOML Format = formatTOML{}
-	FormatYAML Format = formatYAML{}
+	FormatJSON  Format = formatJSON{}
+	FormatJSONC Format = formatJSONC{}
+	FormatTOML  Format = formatTOML{}
+	FormatYAML  Format = formatYAML{}
 )
 
 // A Format is a serialization format.
@@ -26,6 +28,9 @@ type Format interface {
 // A formatJSON implements the JSON serialization format.
 type formatJSON struct{}
 
+// A formatJSONC implements the JSONC serialization format.
+type formatJSONC struct{}
+
 // A formatTOML implements the TOML serialization format.
 type formatTOML struct{}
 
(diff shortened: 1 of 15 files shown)


felipecrs:
This is awesome, and I'll test it.

I'd just say that I suggested `fromJsonc` because this is the language identifier used by vscode itself as well:

image

felipecrs:
GitHub Markdown itself as well:

```md

```jsonc
{
  // comment and trailing comma
  "test": true,
}
```


```

**Comment by felipecrs** [comment]
Deno also accepts `.jsonc` as a configuration file format:

https://deno.land/manual@v1.31.1/getting_started/configuration_file

**Comment by halostatue** [comment]
I think that adding `jsonc` to the supported types/extensions (with a note that it is Visual Studio Code’s variant of JSONC) but leaving the function as `fromJWCC` and a note (to improve search-ability) that it will successfully parse VS Code’s JSONC is probably the right choice here, as JWCC is *extremely* specific about the limited extension that it does to JSON parsing.

**Comment by twpayne** [comment]
Thanks for the super-fast testing and feedback!

So we need two things:

1. The name of the format, for use in documentation, and an optional link to the specification, or at least a link to where users can find out more about the format. The template function should be named after this format, e.g. `fromJwcc` or `fromJsonc` (the ugly lowecased acronym is for consistently with minorminds/sprig's ugly lowercased acronyms (#2668)).
3. A list of file extensions that we read in this format. Zero or more of `.jsonc`, `.jwcc`, `.hujson`, and maybe others.

Obviously, as there is no official standard here, there is no right answer, only a least-bad answer. The argument "do what VSCode does" is a very strong one as this is what most developers use these days.

So, is the following reasonable?
1. We call the format JSONC.
2. We never link to a specification because there is none.
3. We use the extension `.jsonc` only.

Thoughts?

**Comment by felipecrs** [comment]
> So, is the following reasonable?
> 
> 1. We call the format JSONC.
> 2. We never link to a specification because there is none.
> 3. We use the extension `.jsonc` only.
> 
> Thoughts?

This is very reasonable to me, although, I believe it would be nice to mention somewhere which Go library is used to parse JSONC (which would be the closest thing to a reference for a specification IMO).

**Comment by felipecrs** [comment]
Actually I tested it only now, but it is working great:

```console
❯ printf "%s\n" "{" "  // This is a comment" '  "a": "whatever",' "}" | /mnt/c/Users/felip/Downloads/chezmoi-linux-amd64/chezmoi execute-template --with-stdin '{{ fromJson .chezmoi.stdin | toPrettyJson }}'
null
❯ printf "%s\n" "{" "  // This is a comment" '  "a": "whatever",' "}" | /mnt/c/Users/felip/Downloads/chezmoi-linux-amd64/chezmoi execute-template --with-stdin '{{ fromJwcc .chezmoi.stdin | toPrettyJson }}'
{
  "a": "whatever"
}
```

pkg/chezmoi/format.go
```diff
@@ -35,22 +40,48 @@ type formatYAML struct{}
...
+		"yaml":   FormatYAML,
 	}
 
 	// FormatsByExtension is a map of all Formats by extension.
 	FormatsByExtension = map[string]Format{
-		"json": FormatJSON,
-		"toml": FormatTOML,
-		"yaml": FormatYAML,
-		"yml":  FormatYAML,
+		"hujson": FormatJSONC,
```
felipecrs, line 52: Do you think it's really worth supporting this extension?

twpayne: Fair point, removed extension.

felipecrs (approved):
This is awesome, as everything about chezmoi.

Issue #2826, opened 2023-03-06
`chezmoi modify` command


felipecrs:
## Is your feature request related to a problem? Please describe.

Since I started using chezmoi modify template I never looked back. The solution is pretty clean, fast, and platform independent. But I believe it could be leverage for more than just managing some files in the sourceDir. For example, see this:

- https://github.com/felipecrs/dotfiles/blob/master/home/.chezmoiscripts/run_after_92-configure-vscode-on-windows.sh.tmpl#L22

Where I'm using a `modify_settings.json` file to modify the VS Code settings for Windows from within chezmoi installed in WSL.

## Describe the solution you'd like

I believe it would be awesome if we had a command like `chezmoi modify <template> [<files>]`, which would modify `<files>` with `<template>` being a modify template and would respect some other chezmoi flags like `--verbose` to output a diff or also `--force` to ask/not ask for confirmation.

## Describe alternatives you've considered

```
result=$(cat <file> | chezmoi execute-template --with-stdin <(cat modify_template_file))
tee <file> <<<"$result"
```

## Additional context

N/A

twpayne:
Thanks, this is an interesting idea, but...

* Modify templates are a clever hack for modifying arbitrary text files and have very practical value in the context of chezmoi, but they are very much a hack. If you were asked "design a system to make changes to a small part of a file while keeping the rest intact" then you almost certainly wouldn't chose `text/template` as a starting language and add a couple of `fromJson | setValueAtPath | toJson`-type functions to it. I mean, maybe you would have those functions, but you wouldn't choose `text/template` as your base language.
* The `chezmoi modify` command would turn chezmoi from being a dotfile manager into a dotfile manager and also an interpreter for a poorly-specified, ad-hoc implementation of Common Lisp. Fundamentally, there are better grammars for modifying files (e.g. see https://github.com/twpayne/chezmoi/issues/2814#issuecomment-1454080285, also `sed`, `jq`, and similar).

In summary, I don't think chezmoi should become an interpreter for a clever-hack language. Instead, I think it would be better for chezmoi to provide something like https://github.com/itchyny/gojq as a template function.

felipecrs:
I totally understand. Adding such command would expose chezmoi to being more than just a dotfiles manager, and I understand your concerns about the consequences of doing so.

For my use case, I'd still be using it in the dotfiles context though, but I'm fine with the alternative (it's only sad that I can't achieve a similar diff style output as the other files being modified by my dotfiles).

[...]

PR #2833, opened 2023-03-08
feat: Add jq template function

twpayne:
Refs #2826.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/jq.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/jq.md
@@ -1,0 +1,16 @@
+# `jq` *query* *input*
+
+`jq` runs the [jq](https://stedolan.github.io/jq/) query *query* against *input*
+and returns a list of results.
+
+!!! example
+
+    ```
+    {{ dict "key" "value" | jq ".key" | first }}
+    ```
+
+!!! warning
+
+    `jq` uses [`github.com/itchyny/gojq`](https://github.com/itchyny/gojq),
+    which behaves slightly differently to the `jq` command in some [edge
+    cases](https://github.com/itchyny/gojq#difference-to-jq).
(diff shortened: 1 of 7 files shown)


halostatue:
I’m getting an unexpected `[]` wrapping the output:

```console
$ go run . execute-template '{{ dict "hello" 1  | jq "." | toJson }}'
[{"hello":1}]
$ go run . execute-template '{{ 1  | jq "." | toJson }}'
[1]⏎
```

halostatue:
This is going to lead to very hard to read template functions:

```
op document list --format json |
  go run . execute-template --with-stdin \
    '{{ .chezmoi.stdin | fromJson | jq ". | map(select(.title | startswith(\"aws-credentials: \")) | { (.title): .id }) | add" | first | toJson }}'
```

I’m also not sure how I would embed dynamic variables given the overloading of `.` and the fact that the JQ source string isn’t a template itself.

I don’t think this is a bad addition, but I don’t know how *useful* it will be.

halostatue:
I think that this will be more useful if we approach the jq source file as `. | jq (includeTemplate "…" (data))`. There’s no need for any additional functions (I was going to suggest `jqFile`, but that probably isn’t necessary with `include` and `includeTemplate`).

halostatue (approved):
We should be on the lookout for more complex examples to demonstrate, because the value of `jq` here will be in the *flexibility* provided.

twpayne:
> I don’t think this is a bad addition, but I don’t know how _useful_ it will be.

That's a very fair point. I'll close this PR for now. We can re-open it if it turns out that there's a good use for this functionality.

twpayne:
Re-opening as @halostatue identified a possible use case in https://github.com/twpayne/chezmoi/discussions/2864#discussioncomment-5338970.

Another use case is a replacement for some of the password manager functions like [`onepasswordDetailsFields`](https://www.chezmoi.io/reference/templates/1password-functions/onepasswordDetailsFields/) and [`bitwardenFields`](https://www.chezmoi.io/reference/templates/bitwarden-functions/bitwardenFields/). In short, the `jq` template function allows extraction of specific data from complex structures in a succinct way.

halostatue:
It would be nice to have this be able to call out to `jq` instead of using `gojq` *if desired* (similar to opting into the built-in `age`, but instead opting into the use of the `jq` binary).

PR #2857, opened 2023-03-15
feat: Add support for ejson

XaF:
This adds support for [ejson](https://github.com/Shopify/ejson) which allows to store secrets in a git repository using asymmetric encryption. Since this is basically encrypted json, once decrypted it can be parsed pretty easily and made accessible to `chezmoi` as the full JSON data structure.

**Note:** I put this as part of the password managers, as it feels like it'd be used in a similar way (this does not encrypt full files like `age` or `gpg` but keeps the structured JSON view and only encrypts the values, i.e. encrypts secrets and not files).

e.g.

```bash
$ ejson keygen
Public Key:
d5f0647bc4eea75296381787c55d52996873f156874885d685ddaf26f3e61f13
Private Key:
xxxxxxxxxxxxxxx
```
```bash
$ cat /tmp/ejson-test.ejson
{
	"_public_key": "d5f0647bc4eea75296381787c55d52996873f156874885d685ddaf26f3e61f13",
	"test": "value",
	"password": "this is me"
}
```
```bash
$ ejson encrypt /tmp/ejson-test.ejson
Wrote 349 bytes to /tmp/ejson-test.ejson.
```
```bash
$ cat /tmp/ejson-test.ejson
{
	"_public_key": "d5f0647bc4eea75296381787c55d52996873f156874885d685ddaf26f3e61f13",
	"test": "EJ[1:JRfDTIvV+/MdH5mK9bjo24s8Sk3zt5TANINbvGLJllw=:I2Dz0WOXA8yWgcC1mj+KormMq+4ZHmuJ:QDMyYLhLUJiz6bNHzWI8vD42jRXR]",
	"password": "EJ[1:JRfDTIvV+/MdH5mK9bjo24s8Sk3zt5TANINbvGLJllw=:Gln2YPhBMv0L7zneR/n2kWImImbNbW3j:7PwFn84jC3og8GPlUpA0gVDzS+kdXgJwx1s=]"
}
```
```bash
$ echo "xxxxxxxxxxxxxxx" > /tmp/d5f0647bc4eea75296381787c55d52996873f156874885d685ddaf26f3e61f13
```
```bash
$ EJSON_KEYDIR=/tmp ./chezmoi execute-template '{{ ejsonDecrypt "/tmp/ejson-test.ejson" }}'
map[_public_key:d5f0647bc4eea75296381787c55d52996873f156874885d685ddaf26f3e61f13 password:this is me test:value]
```
```bash
$ EJSON_KEYDIR=/tmp ./chezmoi execute-template '{{ (ejsonDecrypt "/tmp/ejson-test.ejson").password }}'
this is me
```

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -165,6 +165,13 @@ sections:
     watch:
       type: bool
       description: Automatically apply changes when files are saved
+  ejson:
+    keyDir:
+      type: 'string'
+      description: Path to directory containing private keys. Defaults to /opt/ejson/keys. Setting the EJSON_KEYDIR environment will also set this value, with lower precedence.
+    key:
+      type: 'string'
+      description: The private key to use for decryption, will supersede using the keyDir if set.
   git:
     autoAdd:
       type: bool
--- a/assets/chezmoi.io/docs/reference/templates/ejson-functions/ejsonDecrypt.md
+++ b/assets/chezmoi.io/docs/reference/templates/ejson-functions/ejsonDecrypt.md
@@ -1,0 +1,16 @@
+# `ejsonDecrypt` *filePath*
+
+`ejsonDecrypt` returns the decrypted content of an
+[ejson](https://github.com/Shopify/ejson)-encrypted file.
+
+*filePath* indicates where the encrypted file is located.
+
+The decrypted file is cached so calling `ejsonDecrypt` multiple
+times with the same *filePath* will only run through the decryption
+process once. The cache is shared with `ejsonDecryptWithKey`.
+
+!!! example
+
+    ```
+    {{ (ejsonDecrypt "my-secrets.ejson").password }}
+    ```
--- a/assets/chezmoi.io/docs/reference/templates/ejson-functions/ejsonDecryptWithKey.md
+++ b/assets/chezmoi.io/docs/reference/templates/ejson-functions/ejsonDecryptWithKey.md
@@ -1,0 +1,17 @@
+# `ejsonDecryptWithKey` *filePath* *key*
+
+`ejsonDecryptWithKey` returns the decrypted content of an
+[ejson](https://github.com/Shopify/ejson)-encrypted file.
+
+*filePath* indicates where the encrypted file is located,
+and *key* is used to decrypt the file.
+
+The decrypted file is cached so calling `ejsonDecryptWithKey` multiple
+times with the same *filePath* will only run through the decryption
+process once. The cache is shared with `ejsonDecrypt`.
+
+!!! example
+
+    ```
+    {{ (ejsonDecryptWithKey "my-secrets.ejson" "top-secret-key").password }}
+    ```
--- a/assets/chezmoi.io/docs/reference/templates/ejson-functions/index.md
+++ b/assets/chezmoi.io/docs/reference/templates/ejson-functions/index.md
@@ -1,0 +1,4 @@
+# ejson functions
+
+The `ejson*` functions return data from
+[ejson](https://github.com/Shopify/ejson)-encrypted files.
--- a/assets/chezmoi.io/docs/user-guide/password-managers/ejson.md
+++ b/assets/chezmoi.io/docs/user-guide/password-managers/ejson.md
@@ -1,0 +1,18 @@
+# ejson
+
+chezmoi includes support for [ejson](https://github.com/Shopify/ejson).
+
+Structured data can be retrieved with the `ejsonDecrypt` template function,
+for example:
+
+```
+examplePassword = {{ (ejsonDecrypt "my-secrets.ejson").password }}
+```
+
+If you want to specify the private key to use for the decryption,
+structured data can be retrieved with the `ejsonDecryptWithKey` template
+function, for example:
+
+```
+examplePassword = {{ (ejsonDecryptWithKey "my-secrets.ejson" "top-secret-key").password }}
+```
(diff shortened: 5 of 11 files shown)


twpayne:
Thanks for this, it looks very solid.

Instead of `exec`'ing `ejson decrypt`, how about importing the [`github.com/Shopify/ejson` module](https://github.com/Shopify/ejson) and calling [`ejson.Decrypt`](https://pkg.go.dev/github.com/Shopify/ejson#Decrypt)? It's one less external binary to depend on.

XaF:
Are we ok with having to maintain the default path for keys for `ejson` ? (in case `ejson` decides to change it anytime)

```diff
@@ -23,13 +25,18 @@ func (c *Config) ejsonDecryptTemplateFunc(filePath string) any {
                c.Ejson.cache = make(map[string]any)
        }

-       output, err := c.ejsonOutput("decrypt", filePath)
+       keyDir, ok := os.LookupEnv("EJSON_KEYDIR")
+       if !ok {
+               keyDir = "/opt/ejson/keys"
+       }
+
+       decrypted, err := ejson.DecryptFile(filePath, keyDir, "")
        if err != nil {
                panic(err)
        }

        var data any
-       if err := json.Unmarshal(output, &data); err != nil {
+       if err := json.Unmarshal(decrypted, &data); err != nil {
                panic(err)
        }
```

twpayne:
> Are we ok with having to maintain the default path for keys for `ejson` ?

Yes. I would make the key directory a configuration variable `ejson.keyDir` with the default value set the same way as the `ejson` command (i.e. the default is `$EJSON_KEYDIR` if the environment variable is set, `/opt/ejson/keys` otherwise). This way, the default works the same way as `ejson`, and the user can override it if they want to.

If `ejson` changes its default value for the key directory, then we can update chezmoi to match its behavior.

Another option would be to make the key directory a second optional argument to the `ejsonDecrypt` template function, but I don't think this is needed yet.

XaF:
> Another option would be to make the key directory a second optional argument to the `ejsonDecrypt` template function, but I don't think this is needed yet.

I do prefer avoiding that too, would make for a bad UX is most cases!
If it was possible with template functions to have optional arguments that'd be different, but I think it's fine to expect that for a given dotfiles repository, all the private keys need to be in a single directory!

The one case that could be useful is to support passing the private key as argument, and this could be a separate method (`ejsonDecryptWithPrivateKey`), so it would allow people to load a private key from another password manager for instance.

I'll look at the refactor to use the module!

XaF:
With the new changes, built and tested:

```bash
$ EJSON_KEYDIR=/tmp ./chezmoi execute-template '{{ (ejsonDecrypt "/tmp/ejson-test.ejson").password }}'; echo
this is me
```

And the new template func to provide the key:
```bash
$ EJSON_KEYDIR=/tmp ./chezmoi execute-template '{{ (ejsonDecryptWithKey "/tmp/ejson-test.ejson" "'$(</tmp/d5f0647bc4eea75296381787c55d52996873f156874885d685ddaf26f3e61f13)'").password }}'; echo
this is me
```

halostatue (approved):

twpayne (commented):
Thank you, this is looking excellent.

For the tests, please could you use the existing `testscript` infrastructure (see `pkg/cmd/testdata/scripts/*.txtar`), rather than introducing mocks. This should also allow you to eliminate `decryptFileFunc` and the `newEjsonConfig` function. Making these changes will help make the code fit chezmoi's existing style.

If you don't want to make these changes for any reason, I'm happy to do them in a follow-up PR. Let me know.

XaF:
> For the tests, please could you use the existing `testscript` infrastructure (see `pkg/cmd/testdata/scripts/*.txtar`), rather than introducing mocks. This should also allow you to eliminate `decryptFileFunc` and the `newEjsonConfig` function. Making these changes will help make the code fit chezmoi's existing style.

Do you have a specific example for a test like this? I wanted to write tests making sure we're making the right call to the ejson lib, without thinking too much about files needing to exist around.
I am not too familiar with txtar and thought it was mostly for overriding binaries and side-effects of binaries.

XaF:
Yes, but this is not using a binary anymore (initially it had the `txtar` when using a binary), it's calling directly the module. I thus cannot have any stand-in binary for the module call! Hence my question :)

halostatue:
> Would that make sense to do? (I generally like unit tests to be testing only what we control, but here it'd more be end-to-end tests)

The `*.txtar` system is an end-to-end test. Certain external commands are mocked (e.g., command-lines as with the 1Password and Bitwarden tooling), but others just use internals like `pkg/cmd/testdata/scripts/templatefuncs.txtar`. Other ones that use internal functions only are:

- `pkg/cmd/testdata/scripts/githubtemplatefuncs.txtar`
- `pkg/cmd/testdata/scripts/inittemplatefuncs.txtar`
- `pkg/cmd/testdata/scripts/templatedata.txtar`

(I would argue that `pkg/cmd/testdata/scripts/encryptiontemplatefuncs.txtar` *should* have a variant which tests the *internal* use of `age`,  not just the command-line use of `age`.)

twpayne:
> Just as a clarification: I could write tests with generated ejson keys and an encrypted file and test the different things, but that also means that we wouldn't only be testing our call process, but also the whole ejson lib process. Would that make sense to do? (I generally like unit tests to be testing only what we control, but here it'd more be end-to-end tests)

Yes, exactly. Specifically:

1. Use the `ejson` CLI to generate a private key and some ejson files.
2. Write a `testcript` test that includes this private key and the ejson files.
3. Test the `ejsonDecrypt` and `ejsonDecryptWithKey` template functions in this `testscript` test.

I am also a huge fan of unit tests, especially when testing that a function works correctly for a wide range of inputs. In this particular case, the range of inputs is quite narrow: given this file and this key, do the `ejsonDecrypt` and `ejsonDecryptWithKey` template functions decrypt the file correctly? We're not testing ejson, we're testing chezmoi's integration with ejson, and so a `testscript` test is the most appropriate.

twpayne (commented):
This looks great, thanks.

As a final thing, please could you squash all the commits into one.

XaF:
Sure! Didn't know if you did a merge-squash to keep PR history :)
Will squash and force-push!

PR #2872, opened 2023-03-19
docs: Wait for elevated PS process to return

alexvy86:
I wish I had done this sooner to when I made the change locally to remember the exact details. But if I recall correctly, this was necessary to make sure that a PS script that was auto-elevating would fully complete before the next chezmoi step/script executes.  Without `-Wait`, `Start-Process` returns immediately and the elevated script could still be running while the "base" one that kicked it off exits and lets chezmoi progress; if anything after that step depends on the elevated script having finished, it might not work as expected.

Of note, in my local setup I removed `-NoExit` from the args passed to the elevated powershell, so it would exit on its own once it's done instead of leaving a open window that might have no output or indication if anything is happening, that the user needs to close for chezmoi to continue doing its thing. But then I decided to replace that with adding an output message inside the script, and a `Read-Host` at the end, so I get a chance to see what happened in the elevated script before it closes, but also so I had some indication that it had completed. I did not include those changes here because they feel a bit more to-each-their-own, but if they feel like a good alternative, I can update the PR.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/machines/windows.md
+++ b/assets/chezmoi.io/docs/user-guide/machines/windows.md
@@ -23,8 +23,24 @@ Put the following at the top of your script:
 if (-Not ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] 'Administrator')) {
   if ([int](Get-CimInstance -Class Win32_OperatingSystem | Select-Object -ExpandProperty BuildNumber) -ge 6000) {
     $CommandLine = "-NoExit -File `"" + $MyInvocation.MyCommand.Path + "`" " + $MyInvocation.UnboundArguments
-    Start-Process -FilePath PowerShell.exe -Verb Runas -ArgumentList $CommandLine
+    Start-Process -Wait -FilePath PowerShell.exe -Verb Runas -ArgumentList $CommandLine
     Exit
   }
 }
 ```
+
+If you use [gsudo](https://gerardog.github.io/gsudo/docs/intro), it has tips on writing
+[self-elevating scripts](https://gerardog.github.io/gsudo/docs/tips/script-self-elevation#self-elevate-script).
+
+## Notes on running elevated scripts
+
+However you decide to run a script in an elevated prompt, as soon as the non-elevated script returns, chezmoi will move to the next step in its
+processing (running more scripts, creating files, etc.).
+Ensure that the elevated script completes *before the non-elevated script exits*, or subsequent steps may not run as expected.
+In the example above, this is accomplished by passing `-Wait` to PowerShell's `Start-Process` cmdlet.
+
+Note that by including `-NoExit` in `$CommandLine`, the new (elevated) PowerShell process/window will not exit automatically on completion.
+This means you'll need to close the new window by hand for chezmoi to continue its steps. If this manual intervention is desired, it would
+be convenient to print a message as the script's last command to indicate completion for you to safely close the elevated window.
+If you want no manual intervention, you can remove `-NoExit` from `$CommandLine`, but then you likely won’t see the output of the elevated
+script, which will make it more difficult to determine if something went wrong during its execution.


twpayne:
Thanks for this. I don't know much about Windows PowerShell. The change makes total sense to me but I'll wait for @bradenhilton's or @zb140's input for a proper review.

bradenhilton:
We might as well remove `if ([int](Get-CimInstance -Class Win32_OperatingSystem | Select-Object -ExpandProperty BuildNumber) -ge 6000)`. All it does is verify that the build number is above 6000, which corresponds to anything after Windows Vista (Windows Vista Service Pack 1 having a build number of 6001, for example).

I think adding `-Wait` should be fine. I use it in some of my own self-elevating scripts:

```powershell
$isAdmin = ([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
if (-not $isAdmin) {
    $Arguments = @('-NoProfile', '-NoLogo', '-NonInteractive', '-File', "`"$($MyInvocation.MyCommand.Path)`"")
    Start-Process -FilePath 'powershell.exe' -ArgumentList $Arguments -Verb 'RunAs' -WindowStyle 'Hidden' -Wait
    exit
}
```

I'm trying to find a way to run the script elevated and acquire its output, but PowerShell makes this unnecessarily difficult.

I really do think we should optionally recommend `gsudo`. Specifically, providing the link to gsudo's [recommended self-elevations](https://gerardog.github.io/gsudo/docs/tips/script-self-elevation#self-elevate-script).

At the end of the day, this is technically outside the scope of chezmoi. We can only provide a best effort solution, not a one size fits all solution.

alexvy86:
I agree that this is accompanying "support" documentation. Maybe we can expand a bit on this docs page to point out the things one needs to be careful with, regardless of how they decide to get the scripts to run elevated. For example, "However you decide to get a script to run in an elevated prompt, make sure the elevated script actually completes before the non-elevated one exits (e.g. by passing `-Wait` to PowerShell's `Start-Process`), so subsequent chezmoi steps can rely on the fact that all previous scripts ran to completion."

bradenhilton:
> ... "However you decide to get a script to run in an elevated prompt, make sure the elevated script actually completes before the non-elevated one exits (e.g. by passing `-Wait` to PowerShell's `Start-Process`), so subsequent chezmoi steps can rely on the fact that all previous scripts ran to completion."

I'm happy with something like that if @twpayne is.

twpayne:
> "However you decide to get a script to run in an elevated prompt, make sure the elevated script actually completes before the non-elevated one exits (e.g. by passing `-Wait` to PowerShell's `Start-Process`), so subsequent chezmoi steps can rely on the fact that all previous scripts ran to completion."

This sounds great, please do update this PR with the extra documentation.

alexvy86:
Additional docs ready, including a note pointing to `gsudo`'s docs in case someone is using it. Let me know if you'd like any changes.

assets/chezmoi.io/docs/user-guide/machines/windows.md
```diff
@@ -28,3 +28,23 @@ if (-Not ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdent
...
 }
 ```
+
+If you use [gsudo](https://gerardog.github.io/gsudo/docs/intro), their documentation specifies how to write self-elevating scripts [here](https://gerardog.github.io/gsudo/docs/tips/script-self-elevation#self-elevate-script).
+
+## Notes on running elevated scripts
+
+However you decide to get a script to run in an elevated prompt, as soon as the non-elevated script returns chezmoi will move to the next step in
+its processing (running more scripts, creating files, etc).
+So make sure that the elevated script actually completes *before the non-elevated one exits*, or subsequent scripts/steps might not run as expected
```
halostatue, line 38: The word `actually` can be removed without changing the meaning of the sentence.

I’d probably rewrite the content here as:

```markdown
However you decide to run a script in an elevated prompt, as soon as the
non-elevated script returns, chezmoi will move to the next step in its
processing (running more scripts, creating files, etc.). Ensure that the
elevated script completes *before the non-elevated script exits*, or subsequent
steps may not run as expected. In the example above, this is accomplished by
passing `-Wait` to PowerShell's `StartProcess` cmdlet.
```

assets/chezmoi.io/docs/user-guide/machines/windows.md
```diff
@@ -28,3 +28,23 @@ if (-Not ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdent
   }
 }
 ```
+
+If you use [gsudo](https://gerardog.github.io/gsudo/docs/intro), their documentation specifies how to write self-elevating scripts [here](https://gerardog.github.io/gsudo/docs/tips/script-self-elevation#self-elevate-script).
```
halostatue, line 32: I would rewrite this to:

```markdown
If you use [gsudo](https://gerardog.github.io/gsudo/docs/intro),
it has tips on writing [self-elevating scripts](https://gerardog.github.io/gsudo/docs/tips/script-self-elevation#self-elevate-script).
```

It’s generally bad practice to put a link behind `[here]`, because the context is lost.

assets/chezmoi.io/docs/user-guide/machines/windows.md
```diff
@@ -28,3 +28,23 @@ if (-Not ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdent
...
+
+## Notes on running elevated scripts
+
+However you decide to get a script to run in an elevated prompt, as soon as the non-elevated script returns chezmoi will move to the next step in
+its processing (running more scripts, creating files, etc).
+So make sure that the elevated script actually completes *before the non-elevated one exits*, or subsequent scripts/steps might not run as expected
+if they depend on what the elevated script does.
+In the example above this is accomplished by passing the `-Wait` switch to PowerShell's `Start-Process` cmdlet.
+
+Also note that by including `-NoExit` as part of the parameters in `$CommandLine`, we're telling the new (elevated) PowerShell process/window to not exit
```
halostatue, line 42: This can be simplified, too:

```markdown
Note that by including `-NoExit` in `$CommandLine`, the new (elevated)
PowerShell process/window will not exit automatically on completion.
This means you'll need to close the new window by hand for chezmoi
to continue its steps.

If this manual intervention is desired, it would be convenient to have print
a message as its last command to indicate completion for you to safely
close the elevated window.

If you want no manual intervention, you can remove `-NoExit` from
`$CommandLine`, but then you likely won’t see the output of the elevated
script, which will make it more difficult to determine if something went wrong
during its execution.
```

Note that this latter will be helped when #2875 is merged, because then you
can test choose to include `-NoExit` based on whether `--verbose` is used
during `chezmoi apply`.

halostatue (commented):
I have some suggestions on the wording that, I think, make it read more easily. Unless either of the other maintainers say, these should be considered optional.

assets/chezmoi.io/docs/user-guide/machines/windows.md
halostatue: This change should probably not be optional, because it’s an accessibility issue. The rest are verbosity/tone suggestions.

alexvy86:
Thanks @halostatue , I do like your suggested wording better, so I applied all changes.

twpayne:
Thank you all for all your contributions here!

@alexvy86 would you mind squashing all the commits together with a commit message with your original commit message `docs: Wait for elevated PS process to return`. The commit message will directly end up in the release notes, which is why it matters.

alexvy86:
Squashed!

twpayne:
Thank you :)

PR #2875, opened 2023-03-19
feat: Set $CHEZMOI_VERBOSE based on --verbose

halostatue:
I want to make my chezmoi scripts more responsive to the --verbose flag, so that they can output their name when running under --verbose (amongst other things).

This change passes `$CHEZMOI_VERBOSE` as the value of `c.Verbose`. With this, I can write:

```bash
#! /usr/bin/env bash

set -euo pipefail

"${CHEZMOI_VERBOSE}" && echo "$(basename "$0")" || true
```


pkg/cmd/config.go
```diff
@@ -1937,6 +1937,7 @@ func (c *Config) persistentPreRunRootE(cmd *cobra.Command, args []string) error
 	for key, value := range c.ScriptEnv {
 		scriptEnv = append(scriptEnv, key+"="+value)
 	}
+	scriptEnv = append(scriptEnv, "CHEZMOI_VERBOSE="+fmt.Sprintf("%v", c.Verbose))
```
twpayne, line 1940: You can add an entry to the map at line 1921 instead.

pkg/cmd/config.go
```diff
@@ -1918,6 +1918,7 @@ func (c *Config) persistentPreRunRootE(cmd *cobra.Command, args []string) error
 		"SOURCE_DIR":    templateData.SourceDir.String(),
 		"UID":           templateData.UID,
 		"USERNAME":      templateData.Username,
+		"VERBOSE":       fmt.Sprintf("%v", c.Verbose),
```
twpayne, line 1921: nit: using `strconv.FormatBool` will avoid all the `fmt.Sprintf` machinery.

twpayne (approved):
LGTM. One very minor nit, but feel free to ignore it.

twpayne:
Actually, thinking again, how should `$CHEZMOI_VERBOSE` be set when `--verbose` is not specified?

Should it be `false`, `0`, or not set at all?

halostatue:
> Actually, thinking again, how should `$CHEZMOI_VERBOSE` be set when `--verbose` is not specified?
> 
> Should it be `false`, `0`, or not set at all?

I chose to make it `false` so that, at least on \*nix, you can *run* the variable.

As long as the value is `true` or `false`, then you can do `"${CHEZMOI_VERBOSE}" && set -x` or something like that. It is simpler than:

```bash
if [[ "${CHEZMOI_VERBOSE}" = 1 ]]; then
  set -x
fi
```

twpayne:
> As long as the value is true or false, then you can do `"${CHEZMOI_VERBOSE}" && set -x` or something like that.

Sorry, I really don't like using the contents of an environment variable as the name as a command to run. I agree that it works, but it's not common practice and I don't think it should be encouraged.

Instead, I would say:

* If `--verbose` is set then set `$CHEZMOI_VERBOSE=1`.
* Otherwise, don't set `$CHEZMOI_VERBOSE`.

The extra characters required are not many:

```bash
[ -n "${CHEZMOI_VERBOSE}" ] && set -x
```

I'll merge this PR as-is and make the change in a follow-up PR, as I'd like to release 2.33.0.

PR #2876, opened 2023-03-19
docs: Add missing `jsonc` to some formats

bradenhilton:
I _think_ these changes are correct. I wasn't sure if `jsonc` was only supported for the config file or not.


twpayne:
Good catch, thanks!

PR #2877, opened 2023-03-19
chore: Deprecate `--autotemplate`

### LINKED ISSUE #2873: autotemplate with a scope
# Problem reproduction
chezmoi data
```
chezmoi:
  ...
  uid: "1000"
  ...
color:
  red: "#FF0000"
  ...
```
original config
```
number: "1000"
color: "FF0000"
```
autotemplate generate
```
number: "{{ .chezmoi.uid }}"
color: "{{ .color.red }}"
```
expect autotemplate
```
number: "1000"
color: "{{ .color.red }}"
```
So, I think chezmoi should give `autotemplate` a scope, for example `chezmoi add --autotemplate=".color" xxx`

# Further discussion
I realize that `chezmoi re-add` will ignore the template file. It's natural that user should edit template file with `chezmoi edit`. If `autotemplate` has a scope, should `chezmoi re-add` include the autotemplate file?

halostatue:
Remove `--autotemplate` from `chezmoi --help` and print a message when it is used.

> > @twpayne Should `--autotemplate` be deprecated for removal in v3?
> Yes, I think it was too ambitious and too fragile.

Resolves #2873 and other related issues.

### DIFF
--- a/pkg/cmd/addcmd.go
+++ b/pkg/cmd/addcmd.go
@@ -53,6 +53,10 @@ func (c *Config) newAddCmd() *cobra.Command {
 	flags.BoolVarP(&c.Add.template, "template", "T", c.Add.template, "Add files as templates")
 	flags.BoolVar(&c.Add.TemplateSymlinks, "template-symlinks", c.Add.TemplateSymlinks, "Add symlinks with target in source or home dirs as templates") //nolint:lll
 
+	if err := flags.MarkDeprecated("autotemplate", "it will be removed in a future release"); err != nil {
+		panic(err)
+	}
+
 	registerExcludeIncludeFlagCompletionFuncs(addCmd)
 
 	return addCmd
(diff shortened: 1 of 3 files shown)


pkg/cmd/addcmd.go
```diff
@@ -53,6 +53,8 @@ func (c *Config) newAddCmd() *cobra.Command {
 	flags.BoolVarP(&c.Add.template, "template", "T", c.Add.template, "Add files as templates")
 	flags.BoolVar(&c.Add.TemplateSymlinks, "template-symlinks", c.Add.TemplateSymlinks, "Add symlinks with target in source or home dirs as templates") //nolint:lll
 
+	flags.MarkDeprecated("autotemplate", "it will be removed in a future release")
```
twpayne, line 56: This needs a quick

```go
    if err := flags.MarkDeprecated(...); err != nil {
        panic(err)
    }
```

so we get an error if we remove the flag and forget to remove the `flag.MarkDeprecated`.

twpayne:
Thank you :)

PR #2893, opened 2023-03-26
chore: Add rbw version check to doctor command

twpayne:


twpayne:
Refs https://github.com/doy/rbw/issues/110.

PR #2894, opened 2023-03-26
docs: Document how to escape `{{` and `}}` in templates

mgeisler:
I ran into this today and had to google a bit until I found the [answer on Stackoverflow](https://stackoverflow.com/q/17641887/110204).

Somewhat surprisingly, the [documentation](https://pkg.go.dev/text/template) doesn't seem to explain this directly: instead it just says that an "argument" can be a value like a Boolean, a string, etc, and a "pipeline" is a series of "commands" and, finally, a "command" can be a single "argument" :smile:

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/tools/diff.md
+++ b/assets/chezmoi.io/docs/user-guide/tools/diff.md
@@ -18,7 +18,7 @@ specify:
 !!! hint
 
     If you generate your config file from a config file template, then you'll
-    need to escape the `{{` and `}}` as `{{"{{"}}` and `{{"}}"}}`. That way 
+    need to escape the `{{` and `}}` as `{{"{{"}}` and `{{"}}"}}`. That way
     your generated config file contains the `{{` and `}}` you expect.
 
 ## Don't show scripts in the diff output
--- a/assets/chezmoi.io/docs/user-guide/tools/diff.md
+++ b/assets/chezmoi.io/docs/user-guide/tools/diff.md
@@ -18,8 +18,8 @@ specify:
 !!! hint
 
     If you generate your config file from a config file template, then you'll
-    need to escape the `{{` and `}}` in your config file template so that they
-    appear in your generated config file.
+    need to escape the `{{` and `}}` as `{{"{{"}}` and `{{"}}"}}`. That way 
+    your generated config file contains the `{{` and `}}` you expect.
 
 ## Don't show scripts in the diff output
 
--- a/assets/chezmoi.io/docs/user-guide/tools/merge.md
+++ b/assets/chezmoi.io/docs/user-guide/tools/merge.md
@@ -18,5 +18,5 @@ state, source state, and target state respectively. For example, to use
 !!! hint
 
     If you generate your config file from a config file template, then you'll
-    need to escape the `{{` and `}}` in your config file template so that they
-    appear in your generated config file.
+    need to escape the `{{` and `}}` as `{{"{{"}}` and `{{"}}"}}`. That way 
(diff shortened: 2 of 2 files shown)


mgeisler:
Sorry for the second commit — I tried doing this in the github.dev code space and I don't see an option to amend a commit there. Perhaps you can squash-merge this?

assets/chezmoi.io/docs/user-guide/tools/diff.md
```diff
@@ -18,8 +18,8 @@ specify:
 !!! hint
 
     If you generate your config file from a config file template, then you'll
-    need to escape the `{{` and `}}` in your config file template so that they
-    appear in your generated config file.
+    need to escape the `{{` and `}}` as `{{"{{"}}` and `{{"}}"}}`. That way
```
halostatue: Two small things:

1. There is trailing whitespace after `way`, causing the build failure.
2. Remove the `the` in this fragment (both files): `` the `{{` ``. 

If I look at this with my documentation editor hat on, I would probably rephrase the whole thing to

```markdown
If you generate your config file from a template, then you need to escape `{{` and `}}`
as `{{ "{{" }}` and `{{ "}}" }}` in your template so that your generated config file
contains `{{` and `}}` characters.
```

It *may* also be worth pointing people to the [delimiters template directive](https://www.chezmoi.io/reference/templates/directives/#delimiters) (`[delimiters template directives](reference/templates/directives/#delimiters)`)  as this can make the question of `{{` and `}}` irrelevant.

This comment applies to both files you have modified.

halostatue (commented):
Looks good, but a couple of ideas.

assets/chezmoi.io/docs/user-guide/tools/diff.md
halostatue: Reapplying this comment as the whitespace has been removed.

assets/chezmoi.io/docs/user-guide/tools/diff.md
```diff
@@ -18,8 +18,8 @@ specify:
...
+    your generated config file contains the `{{` and `}}` you expect.
```
halostatue, line 22: Can you remove the `the` in this fragment (both files): `` the `{{` ``.?

If I look at this with my documentation editor hat on, I would probably rephrase the whole thing to

```markdown
If you generate your config file from a template, then you need to escape `{{` and `}}`
as `{{ "{{" }}` and `{{ "}}" }}` in your template so that your generated config file
contains `{{` and `}}` characters.
```

It *may* also be worth pointing people to the [delimiters template directive](https://www.chezmoi.io/reference/templates/directives/#delimiters) (`[delimiters template directives](reference/templates/directives/#delimiters)`)  as this can make the question of `{{` and `}}` irrelevant.

This comment applies to both files you have modified.

twpayne:
Thanks! I've merged and squashed.

assets/chezmoi.io/docs/user-guide/tools/diff.md
mgeisler: Sorry for not coming back to this, I was traveling.

One thing I learned much more recently than I would like to admit is the GitHub suggestion feature — the left-most icon in the toolbar above. It looks like this:

```suggestion
    your generated config file contains `{{` and `}}` you expect.
```

I use this a lot in [another repository](https://github.com/google/comprehensive-rust) to apply all sorts of small and large fixups to PRs. Often, contributors don't really care about the exact wording, so as a maintainer it's much easier if I fix things and then squash-merge the branch on the spot.

PR #2912, opened 2023-04-05
fix: Change where fish completions are installed

J-M0:
According to the fish documentation[1], `/usr/share/fish/vendor_completions.d/` is the proper place for third-parties to install completion files.

[1] https://fishshell.com/docs/current/completions.html


halostatue (approved):

twpayne:
Thanks, but this is for chezmoi's `.deb` and `.rpm` packages. As far as I can tell, on both Debian and Fedora distributions, fish completions go in `/usr/share/fish/completions`. On these distributions, `/usr/share/fish/vendor_completions.d` is not used.

Issue #2942, opened 2023-04-21
Digging into .chezmoi.config does not work in templates


halostatue:
## Describe the bug

#2940 introduced `.chezmoi.config` for use in templates, but trying to request the data results in a template execution error.

## To reproduce

```console
$ chezmoi execute-template '{{ .chezmoi.config.keepassxc.command  }}' 
chezmoi: template: arg1:1:11: executing "arg1" at <.chezmoi.config.keepassxc.command>: can't evaluate field keepassxc in type interface {}
```

## Expected behavior

```console
$ chezmoi execute-template '{{ .chezmoi.config.keepassxc.command  }}'
keepassxc-cli
```

## Additional context

This can be worked around by piping through JSON:

```console
$ chezmoi execute-template '{{ (.chezmoi.config | toJson | fromJson).keepassxc.command  }}'
keepassxc-cli
```

It is probably sufficient to use a JSON-serialized copy of the read config rather than storing the read config itself.

twpayne:
Doh! Good spot. I forgot to rename the fields. Using uppercase names (as they appear in the [`ConfigFile` struct definition](https://github.com/twpayne/chezmoi/blob/ebc536f0dbf7c54ab8677b202532736be8d61c6c/pkg/cmd/config.go#L91-L151) works (e.g. `chezmoi execute-template '{{ .chezmoi.config.Keepassxc.Command }}'`) but the case is wrong. Thanks for testing and fix coming up.

twpayne:
Fixed with #2943 and tagged v2.33.3.

PR #2943, opened 2023-04-21
fix: Correct capitalization of .chezmoi.config template variables

twpayne:
`.chezmoi.config` introduces so many more template variables that it renders `--autotemplate` useless, so I've just removed `--autotemplate`.


twpayne:
Fixes #2942.

PR #2948, opened 2023-04-24
docs: Update encryption documentation to change to the root directory

danielo515:
Related to this discussion: #2946

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
+++ b/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
@@ -12,10 +12,10 @@ This can be achieved with the following process:
 4. Configure chezmoi to use the private key.
 5. Add encrypted files.
 
-First, change to chezmoi's source directory:
+First, change to chezmoi's root directory:
 
 ```console
-$ chezmoi cd
+$ chezmoi cd ~
 ```
 
 Generate an age private key encrypted with a passphrase in the file
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -237,7 +237,7 @@ type templateData struct {
 	args           []string
 	cacheDir       chezmoi.AbsPath
 	command        string
-	config         ConfigFile
+	config         map[string]any
 	configFile     chezmoi.AbsPath
 	executable     chezmoi.AbsPath
 	fqdnHostname   string
@@ -2084,7 +2084,7 @@ func (c *Config) newTemplateData(cmd *cobra.Command) *templateData {
 		args:         os.Args,
 		cacheDir:     c.CacheDirAbsPath,
 		command:      cmd.Name(),
-		config:       c.ConfigFile,
+		config:       c.ConfigFile.toMap(),
 		configFile:   c.configFileAbsPath,
 		executable:   chezmoi.NewAbsPath(executable),
 		fqdnHostname: fqdnHostname,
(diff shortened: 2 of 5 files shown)


halostatue (changes requested):

assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
```diff
@@ -18,6 +18,12 @@ First, change to chezmoi's source directory:
 $ chezmoi cd
 ```
 
+Note that, if you are using a [chezmoiroot](/user-guide/advanced/customize-your-source-directory/#use-a-subdirectory-of-your-dotfiles-repo-as-the-root-of-the-source-state) differrent than the default one, you also need to cd into that directory
```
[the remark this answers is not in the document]
danielo515: Thanks for your comments on this PR @halostatue 
Will update the PR to reflect your suggestion and I will see if I am able to find a spot to document the beyond this.

assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
```diff
@@ -15,13 +15,7 @@ This can be achieved with the following process:
 First, change to chezmoi's source directory:
```
halostatue, line 15: Can you change the word `source` to either `home` or `root` so that it’s clearer that we’re talking about `.chezmoiroot`?

halostatue (changes requested):

assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
danielo515: Sure, but first a note about `chezmoi cd ~`. I noticed that, in my case, the shell expands the tilde to my home directory. Does it need any escaping or does any other alternative name with less collision chance exist?

danielo515:
```suggestion
First, change to chezmoi's root directory:
```

halostatue: The answer to that is a little subtle. For at least 99.44% of all users, the answer is no. `chezmoi cd ~` and `chezmoi cd $HOME` and `chezmoi cd /Users/austin`would all be the same result for me, because `$HOME` is my chezmoi target directory, which is the root directory that `.chezmoiroot` maps to.

> As a side example, if I’m in `~/.config`, then doing `chezmoi cd .vim` will result in an error, `chezmoi: .vim: not in source state`. However, `chezmoi cd fish` *does* work, because `~/.config/fish` is present as `~/.local/share/chezmoi/home/private_dot_config/fish`. When additional arguments were added to `chezmoi cd`, they were added in ways that make it *easy* to get the right thing done regardless.

For remaining users, they’re explicitly doing something unusual and the place to document this particular exception is where a different target is documented.

This same principle is why I’m suggesting this file be minimally updated and that, if further work is needed, the documentation for `.chezmoiroot` be updated to indicate what portions of chezmoi’s operations would be updated.

halostatue (approved):

halostatue:
Now we just need you to rebase and squash your commits and update the commit message and PR title, as [documented](https://www.chezmoi.io/developer/contributing-changes/).

I’d suggest `doc: Update encryption documentation to change to the root directory` (I think it’s `doc`; it might be `docs`; check the commit history to be sure.)

danielo515:
> Now we just need you to rebase and squash your commits and update the commit message and PR title, as [documented](https://www.chezmoi.io/developer/contributing-changes/).
> 
> I’d suggest `doc: Update encryption documentation to change to the root directory` (I think it’s `doc`; it might be `docs`; check the commit history to be sure.)

According to the spec, it is docs 😉 
However, this feature does not work in my installation. Was it added recently maybe? I can't provide `chezmoi cd` any argument.

PR #2984, opened 2023-05-09
fix: ensure the sourceDir is updated before creating config file

### LINKED ISSUE #2865: .chezmoiroot not working in GitHub Actions
## Describe the bug

`.chezmoiroot` seems to be ignored when using one line install with GitHub Actions

## To reproduce

add `.chezmoiroot`
```
home
```

add `home/.chezmoiscripts/run_once_echo.sh.tmpl`

```
{{ if eq .chezmoi.os "darwin" -}}
#!/usr/bin/env bash

set -euo pipefail

echo {{ .chezmoi.sourceDir }}

{{ end -}}

```

add `.github/workflows/test.yaml`
```
name: Test install

on:
  workflow_dispatch:
    inputs:
      username:
        description: 'GitHub UseName'
        required: true
        type: string

jobs:
  test:
    strategy:
      fail-fast: false
      matrix:
        platform: [ubuntu-latest, macos-latest]
    runs-on: ${{ matrix.platform }}
    timeout-minutes: 30
    steps:
    - name: Run install
      run: sh -c "$(curl -fsLS get.chezmoi.io)" -- init --apply ${{ inputs.username }} 
```

when dispatch Test install workflow, echo `/Users/runner/.local/share/chezmoi`

## Expected behavior

echo `/Users/runner/.local/share/chezmoi/home`

4ooc:
Close #2865

### DIFF
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -665,6 +665,10 @@ func (c *Config) colorAutoFunc() bool {
 // createAndReloadConfigFile creates a config file if it there is a config file
 // template and reloads it.
 func (c *Config) createAndReloadConfigFile(cmd *cobra.Command) error {
+	c.templateData.sourceDir, _ = c.getSourceDirAbsPath(&getSourceDirAbsPathOptions{
+		refresh: true,
+	})
+
 	// Find config template, execute it, and create config file.
 	configTemplate, err := c.findConfigTemplate()
 	if err != nil {
--- a/pkg/cmd/initcmd.go
+++ b/pkg/cmd/initcmd.go
@@ -201,14 +201,6 @@ func (c *Config) runInitCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var err error
-	c.SourceDirAbsPath, err = c.getSourceDirAbsPath(&getSourceDirAbsPathOptions{
-		refresh: true,
-	})
-	if err != nil {
-		return err
-	}
-
 	if err := c.createAndReloadConfigFile(cmd); err != nil {
 		return err
 	}
--- a/pkg/cmd/testdata/scripts/issue2865.txtar
+++ b/pkg/cmd/testdata/scripts/issue2865.txtar
@@ -25,11 +25,11 @@ stdout ${CHEZMOISOURCEDIR@R}/home
 exec chezmoi execute-template '{{ .testDir }}'
 stdout ${CHEZMOISOURCEDIR@R}/home
 
--- golden/chezmoiroot --
-home
 -- golden/chezmoi.toml --
 [data]
   testDir="{{ .chezmoi.sourceDir }}"
+-- golden/chezmoiroot --
+home
 -- golden/run_once_echo.sh --
 #!/bin/sh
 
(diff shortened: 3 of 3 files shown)


4ooc:
`chezmoi: fork/exec $WORK\.tmp\2538963876.echo.sh: %1 is not a valid Win32 application.`

Does this mean that Windows is unable to run a ﻿`.sh` file?
I’m sorry that I don't know how to get the test to pass on Windows.

4ooc:
I’m having trouble when running ﻿`make test` on my machine—some of the test cases are failing, but they are successful when running on GitHub Actions. As a newcomer to Go, is there anything I might have missed?

For example:
```
    --- FAIL: TestScript/external (0.22s)
        testscript.go:520: # test that chezmoi diff includes external files by default (0.037s)
            # test that chezmoi diff --exclude=externals excludes diffs from external files (0.026s)
            # test that chezmoi reads external files from .chezmoiexternal.toml (0.048s)
            # test that chezmoi reads executable external files from .chezmoiexternal.toml (0.047s)
            # test that chezmoi managed --include=externals lists external targets (0.031s)
            > exec chezmoi managed --include=externals
            [stdout]
            .dir
            .dir/._dir
            .dir/dir
            .dir/dir/._file
            .dir/dir/._symlink
            .dir/dir/file
            .dir/dir/symlink
            > cmp stdout golden/managed
            diff stdout golden/managed
            --- stdout
            +++ golden/managed
            @@ -1,7 +1,4 @@
             .dir
            -.dir/._dir
             .dir/dir
            -.dir/dir/._file
            -.dir/dir/._symlink
             .dir/dir/file
             .dir/dir/symlink

            FAIL: testdata/scripts/external.txtar:30: stdout and golden/managed differ
```

halostatue:
Those look, somehow, like [AppleDouble encoding files](https://apple.stackexchange.com/questions/414947/how-to-prevent-macos-from-adding-dot-underscore-files-whatever-html-in-zip-a) from a zip. Does the txtar format do something with zip files? (It’s been a while since I’ve run the tests; I get "odd" results sometimes, too because I use `fish` as my root shell, but I don’t recall anything like this.)

twpayne:
As far as I know, txtar doesn't have anything to do with zip files. If this problem was due to the use of txtar files, I would expect to have seen it on other machines. As it is, it's only been observed on @4ooc's machine so far.

twpayne:
Thank you very much for this!

I tidied up the code and tests a little in #2989, keeping you as the author.

4ooc:
My machine information:
`Darwin Kernel Version 22.4.0: Mon Mar  6 21:01:02 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8112 arm64 arm4`

> Those look, somehow, like [AppleDouble encoding files](https://apple.stackexchange.com/questions/414947/how-to-prevent-macos-from-adding-dot-underscore-files-whatever-html-in-zip-a) from a zip. Does the txtar format do something with zip files? 

Thank you, `AppleDouble files` is the cause of the problem.`tar` always creates and archives them in compressed file on my machine running macOS 13, but other utilities won't remove them and only extract them. Initially, I erroneously thought that the ﻿`dot_file` had not been fully converted.

> As far as I know, txtar doesn't have anything to do with zip files. If this problem was due to the use of txtar files, I would expect to have seen it on other machines. As it is, it's only been observed on @4ooc's machine so far.

I utilized "gnu-tar" instead of "/usr/bin/tar", and all tests were successful.

halostatue:
That's odd, because I’m also on macOS 13 and my `tar` is `/usr/bin/tar`. I’ll have to see what happens when I run the tests in the future.

PR #3038, opened 2023-06-12
chore: Simplify code by using strings.EqualFold, time.Before

twpayne:
Re-opens #3034. The original branch was deleted so I couldn't simply re-open the original PR.

Thanks for this @alexandear!

@bradenhilton was right in https://github.com/twpayne/chezmoi/pull/3034#issuecomment-1586907929 that this subtly changes the semantics of refresh periods, but refresh periods are more guidelines than rules and as Go records timestamps in nanoseconds, the chances of hitting this are minuscule and it wouldn't matter anyway.


bradenhilton:
Thanks for the explanation @twpayne!

I hope my comment didn't discourage @alexandear, it certainly wasn't my intention.

Issue #2787, opened 2023-02-22
build for AIX


arozehnal:
I really like chezmoi, thanks for it. 
I tried to create my own build for AIX, but the compilation end with errors. Is there any way to fix it?"

## Describe the bug
build error

## To reproduce
```
make install-from-git-working-copy
go install -ldflags "-X main.version=v2.31.0 \
        -X main.commit=461b175d1bb70657c249194bc95452ca82fc16d9 \
        -X main.date=1676974502 \
        -X main.builtBy=source"
# github.com/atotto/clipboard
/go/pkg/mod/github.com/atotto/clipboard@v0.1.4/clipboard.go:10:9: undefined: readAll
/go/pkg/mod/github.com/atotto/clipboard@v0.1.4/clipboard.go:15:9: undefined: writeAll
# github.com/containerd/console
/go/pkg/mod/github.com/containerd/console@v1.0.3/console.go:83:12: undefined: checkConsole
/go/pkg/mod/github.com/containerd/console@v1.0.3/console.go:86:9: undefined: newMaster
# github.com/go-git/go-billy/v5/osfs
/go/pkg/mod/github.com/go-git/go-billy/v5@v5.4.1/osfs/os_posix.go:15:14: undefined: unix.Flock
/go/pkg/mod/github.com/go-git/go-billy/v5@v5.4.1/osfs/os_posix.go:22:14: undefined: unix.Flock

# github.com/twpayne/chezmoi/v2/pkg/shell
pkg/shell/shellcgo.go:33:94: cannot use _Ctype_size_t(buflen) (value of type _Ctype_ulong) as type _Ctype_int in variable declaration
make: *** [Makefile:40: install-from-git-working-copy] Error 2

```
## Expected behavior
binary for AIX

## Additional context
OS AIX v.7.2
go version go1.18.10 aix/ppc64

halostatue:
In addition to the build issue in `shell/shellcgo.go` (which doesn't look pretty to solve), there are dependencies that require update.

1. atotto/clipboard#65 prevents compiling for AIX.
2. containerd/console is also missing `aix` tags (see below).  It seems that simply adding those will be sufficient, but…
3. go-git/go-billy is not missing `aix` tags (the posix file uses `+build !plan9,!windows,!js`), but AIX does not have a `flock` system call (see below) and it will need an `os_aix.go` file created that uses what is provided by Go in `flock_aix.go`.

[...]

PR #3060, opened 2023-06-22
fix: Drop Solaris and Illumos support

twpayne:
Sadly, it looks like there is no appetite for maintaining support for Solaris and Illumos in chezmoi's dependencies.

Refs https://github.com/charmbracelet/bubbletea/issues/737.
Refs https://github.com/containerd/console/issues/75.
Refs https://github.com/containerd/console/pull/76.
Refs https://github.com/charmbracelet/console/pull/1.


twpayne:
Labeled `do-not-merge` pending https://github.com/charmbracelet/console/pull/2.

halostatue:
Should we also *officially* drop AIX (#2787) since it’s blocked on console and two other dependencies, as well as a change required in chezmoi itself which is hard to do without access to an AIX system for testing?

halostatue (approved):

twpayne:
I'm happy to leave #2787 open for now as a top-level tracking issues for all the dependencies that need to be patched. Not that I have any hope that this will happen any time soon.

PR #3069, opened 2023-06-29
docs: Remove reference to tar and /dev/stdin so code works on Windows

Sacro:
Unsure why `command = "tar"` doesn't work on windows, but for some reason it wants to read from a tape, not stdin.

Windows doesn't have /dev/stdin so any commands referencing it fail. It is the default anyway, so no need to specify it.


twpayne:
Thank you very much!

Issue #3072, opened 2023-06-29
PR 3069 makes things worse 


Sacro:
## Describe the bug

It doesn't seem to be passing things via tar at all now (after #3069), just dumping the compressed data into the destination.

## To reproduce

Apply anything that should use filter.cmd/filter.args

## Expected behavior

Data is un-tarred 

## Output of command with the `--verbose` flag

```console
$ chezmoi --verbose $COMMAND
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
```

</details>

## Additional context

[Edited by @halostatue to make the referenced PR clearer.]

twpayne:
OK, I've reverted #3069 with 439ccf8f4f092a3b1f909e430775db933dc7c740.

Sacro:
I've raised a feature request as 3079, to look at the idea of doing zip/tar stuff internally, rather than shelling out where possible.

Is there a Discord or something, I'm happy to take a look at this myself.

twpayne:
This is being added in #3080.

There is no Discord for chezmoi. I don't have time to do chat support.

PR #3080, opened 2023-07-04
feat: Add archive-file externals

### LINKED ISSUE #3079: Allow getting files from tar / zip without shelling out

Given that Windows seems to have issues with filter.args for `tar`, it might be better to have the option for chezmoi to pull the file itself.

When using "type: file", don't shell out, or maybe have a "type: archive-file" that doesn't shell out.

I've tried hacking around with tar on Windows, but can't seem to get it to be happy.

twpayne:
Fixes #3079.
Refs #3069.
Refs #3072.

@Sacro is this what you had in mind? Would you be able to test this?

### DIFF
--- a/assets/chezmoi.io/docs/reference/special-files-and-directories/chezmoiexternal-format.md
+++ b/assets/chezmoi.io/docs/reference/special-files-and-directories/chezmoiexternal-format.md
@@ -13,32 +13,33 @@ externals to be included on different machines.
 
 Entries are indexed by target name relative to the directory of the
 `.chezmoiexternal.$FORMAT` file, and must have a `type` and a `url` field.
-`type` can be either `file`, `archive`, or `git-repo`. If the entry's parent
-directories do not already exist in the source state then chezmoi will create
-them as regular directories.
+`type` can be either `file`, `archive`, `archive-file`, or `git-repo`. If the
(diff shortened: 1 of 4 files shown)


Sacro:
This is exactly what I had in mind. 

I've converted all my externals to use this, and it works beautifully on Windows and Linux.

https://github.com/Sacro/dotfiles/commit/71fad7d97606ea5841802b763d9b31d310cbae71

felipecrs:
Wow, this is amazing!

dhruvinsh:
does the `archive-file` only mean to support `.tar.gz` files? can it be expanded to work with `.gz` files? 

```jinja-toml
.local/bin/taplo:
  type: file
  url: >-
    https://github.com/tamasfe/taplo/releases/download/0.8.0/taplo-full-linux-x86_64.gz
  executable: true
  path: taplo
  filter:
    command: gzip
    args:
      - "-d"
```

twpayne:
`archive-file` works on archives (`.tar` and `.zip`) and supports several compression methods.

In your case you have a compressed file, not an archive, so the use of `filter.command` and `filter.args` is correct.

dhruvinsh:
Good to know difference between compressed and archive. :) Thanks.

Issue #3081, opened 2023-07-05
2.34.3 fails to detect hostname properly on Void linux; 2.34.2 works as expected


oliver-cfc:
## Describe the bug

2.34.3 incorrectly thinks my hostname is localhost
2.34.2 correctly thinks my hostname is n22void 

## To reproduce

Tested on up-to-date void linux.  
Run: chezmoi data | grep hostname

## Expected behavior

2.34.3 should show n22void as the hostname.  It is what is set in /etc/hostname and what uname -n shows.

## Output of command with the `--verbose` flag

```console
$ chezmoi --verbose data | grep -E '(version|hostname)'                                                             
    "hostname": "localhost",                                                                                                            
      "version": "#1 SMP PREEMPT_DYNAMIC Thu Jun 29 20:28:47 UTC 2023"                                                                  
    "version": {                                                                                                                        
      "version": "2.34.3"      
```

```console
$ chezmoi --verbose data | grep -E '(version|hostname)'                                                             
    "hostname": "n22void",                                                                                                              
      "version": "#1 SMP PREEMPT_DYNAMIC Thu Jun 29 20:28:47 UTC 2023"                                                                  
    "version": {                                                                                                                        
      "version": "2.34.2"         
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor                                                                                                    
RESULT   CHECK                MESSAGE                                                                                                   
ok       version              v2.34.3, commit v2.34.3, built at 2023-07-04, built by xbps                                               
ok       latest-version       v2.34.3                                                                                                   
ok       os-arch              linux/amd64 (Void)                                                                                        
ok       uname                Linux n22void 6.3.10_1 #1 SMP PREEMPT_DYNAMIC Thu Jun 29 20:28:47 UTC 2023 x86_64 GNU/Linux               
ok       go-version           go1.20.5 (gc)                                                                                             
ok       executable           /usr/bin/chezmoi                                                                                          
ok       config-file          ~/.config/chezmoi/chezmoi.toml, last modified 2023-04-15T17:08:09-04:00                                   
ok       source-dir           ~/.local/share/chezmoi is a git working tree (clean)                                                      
ok       suspicious-entries   no suspicious entries                                                                                     
ok       working-tree         ~/.local/share/chezmoi is a git working tree (clean)                                                      
ok       dest-dir             ~ is a directory                                                                                          
ok       umask                022                                                                                                       
ok       cd-command           found /bin/bash                                                                                           
[...]

PR #3082, opened 2023-07-05
fix: Never consider localhost.localdomain in /etc/hosts as the FQDN

twpayne:
Refs #3081.

### DIFF
--- a/pkg/chezmoi/chezmoi.go
+++ b/pkg/chezmoi/chezmoi.go
@@ -297,10 +297,26 @@ func etcHostsFQDNHostname(fileSystem vfs.FS) (string, error) {
 		text = strings.TrimSpace(text)
 		text, _, _ = strings.Cut(text, "#")
 		fields := whitespaceRx.Split(text, -1)
-		if len(fields) > 1 && net.ParseIP(fields[0]).IsLoopback() &&
-			strings.Contains(fields[1], ".") {
-			return fields[1], nil
+		if len(fields) < 2 {
+			continue
 		}
+		if !net.ParseIP(fields[0]).IsLoopback() {
+			continue
+		}
+		hostname, domainname, found := strings.Cut(fields[1], ".")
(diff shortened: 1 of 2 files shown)


pkg/chezmoi/chezmoi.go
```diff
@@ -298,7 +298,7 @@ func etcHostsFQDNHostname(fileSystem vfs.FS) (string, error) {
 		text, _, _ = strings.Cut(text, "#")
 		fields := whitespaceRx.Split(text, -1)
 		if len(fields) > 1 && net.ParseIP(fields[0]).IsLoopback() &&
-			strings.Contains(fields[1], ".") {
+			strings.Contains(fields[1], ".") && fields[1] != "localhost.localdomain" {
```
halostatue, line 301: I would maybe make this `!strings.StartsWith(fields[1], "localhost")`. It would break if the name is `localhostfoo`, but it would prevent the case of `localhost.localdomain localhost` or `localhost localhost.localdomain`. I also think that `localdomain` may be a Linux thing; it could be `localhost.local` for macOS.

halostatue (approved):

pkg/chezmoi/chezmoi.go
twpayne: Good point -- thanks! I've updated the PR to explicitly exclude `.localdomain` and `.local`.

roobre:
@twpayne Could we get a patch release for this? I think it might actually be an important bug. Not sure about others, but I happen to have all my dotfiles if-else-d by `.chezmoi.hostname`, so running `chezmoi apply` after a system upgrade broke a huge part of my scripts 😅 

Thankfully it was easy to find this issue and downgrade, but I think it's likely that this happens to others as well.

halostatue:
I think that we should possibly resolve the breakage introduced by #3095 before cutting a release.

PR #3111, opened 2023-07-22
Correct private files support in yadm

vaygr:
Mention of private files support in yadm is confusing. As per https://yadm.io/docs/faq#directory-creation, the only 2 directories that can optionally be made private are `~/.ssh` and `~/.gnupg`.

It feels that chezmoi is the only dotfiles manager that can attach attributes to files/directories indicating of private permissions?

### DIFF
--- a/assets/chezmoi.io/docs/comparison-table.md
+++ b/assets/chezmoi.io/docs/comparison-table.md
@@ -17,7 +17,7 @@
 | Source repos                           | Single        | Single            | Multiple          | Multiple                 | Single                       | Single     |
 | dotfiles are...                        | Files         | Symlinks          | Files             | Files                    | Files                        | Files      |
 | Config file                            | Optional      | Required          | Optional          | None                     | Optional                     | Optional   |
-| Private files                          | ✅            | ❌                | ❌                | ❌                       | ✅                           | ❌         |
+| Private files                          | ✅            | ❌                | ❌                | ❌                       | ❌                           | ❌         |
 | Show differences without applying      | ✅            | ❌                | ❌                | ✅                       | ✅                           | ✅         |
 | Whole file encryption                  | ✅            | ❌                | ❌                | ❌                       | ✅                           | ❌         |
 | Password manager integration           | ✅            | ❌                | ❌                | ❌                       | ❌                           | ❌         |


bradenhilton:
I'm not familiar with yadm.

Perhaps `Partial` would be more accurate?

cc @TheLocehiliosan if you wouldn't mind.

twpayne:
Refs #1787 and a8720add8570d2724b81628099480a2921ef4221.

vaygr:
I'm not sure. Private != Encrypted that a8720ad mentions. I've been using yadm for a while and couldn't find a way to set strict permissions (like `0600` or `0700`) to files or directories, and git doesn't track permissions.

TheLocehiliosan:
yadm will set private permissions for ssh, gnupg, and any encrypted archive. 

Additionally, the encrypted archive retains the original permissions, so if those are private, they remain private.

vaygr:
@TheLocehiliosan thanks, that clears things up. Though I think mentioning that distinction is important: if I want to set `0700` for a directory or `0600` for a file that's not encrypted, but needs to be private for any reason, yadm wouldn't be able to handle it.

TheLocehiliosan:
You can define permissions inside a bootstrap script run by yadm. Many people do that.

vaygr:
Yes, but well, if we think about it this way, it means any dotfiles manager that supports bootstrapping or hooks supports at least private files. But in the table only chezmoi and yadm are indicated of doing so.

- dotbot for example has `mode` param support: https://github.com/anishathalye/dotbot#example-1
- rcm has hooks support: https://thoughtbot.github.io/rcm/rcup.1.html#k
- vcsh has hooks support: https://github.com/RichiH/vcsh/blob/main/doc/hooks

while in fact only chezmoi has _native_ support for private files.

TheLocehiliosan:
Typing chmod in front of some file patterns seems like a very simple configuration to me. 

It sounds like this distinction is very important to you. So label it however you want.

Someone tagged me for my opinion, so I gave it.

Cheers

halostatue:
I think that changing this from ✅ to ❌ is incorrect given the conversation above, and ⁉️ is an overstatement given that the effort is not "significant", per @TheLocehiliosan.

**If** we’re going to change this from ✅, it should either be a new symbol (maybe❓or ❗️) and legend entry or we consider adding footnotes, because the answer is "it’s complicated and *can* be done.” IMO, the chezmoi way is *more* straightforward, but it takes getting used to.

vaygr:
@halostatue,

- I agree that changing this from :white_check_mark: to :x: is incorrect.
- I'd argue that it should be changed for all 4 (yadm, dotbot, rcm, vcsh), since the effort is equal for each if it's an acceptable solution for yadm.

twpayne:
Thanks @vaygr for the PR and thanks all for the discussion.

There is an existing `⁉️` symbol used for where the functionality can be achieved but only "with significant manual effort", but I don't think that adding a `chmod` line to a script qualifies as this.

Therefore, I'm going to leave the comparison table as-is, i.e. with a ✅ for yadm's support for private files.

vaygr:
@twpayne thanks, what about other dotfiles managers? Do they qualify?

twpayne:
To be honest, it's all very subjective. I really appreciate the author of yadm contributing here. The table is accurate as far as my knowledge goes. I will leave it to the authors of the other dotfile managers in the table to correct any errors about their software.

PR #3112, opened 2023-07-23
feat: Add support for Azure Key Vault secrets.

catay:


assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
```diff
@@ -0,0 +1,25 @@
...
+
+A secret value can be retrieved with the `azureKeyVault` template function, for
+example:
+
+```
+exampleSecret = {{ azureKeyVault "my-secret-name" }}
+```
+
+The Azure Key Vault URL can be specified in the chezmoi's config file with
+`azureKeyVault.vaultUrl.
```
halostatue, line 13: This should *probably* be a second optional parameter in the event that there are two different projects that need configuring which use different key vaults. `azureKeyVault.vaultUrl` would be used if otherwise unspecified.

I don’t know what an Azure Key Vault URL looks like, but for the suggestion below, I’m going to assume it’s an HTTPS URL. For the case where there may be multiple key vaults referenced, we may want to make it so that the function accepts either `{{ azureKeyVault "my-secret-name" ["vault-url" | "vault-url-alias"] }}` where the aliases would be defined in the config file under `azureKeyVault.vaultUrls` (plural), which would be a `map[string]string`:

```yaml
azureKeyVault:
  vaultUrls:
    default: https://my-default-key-vault-url/
    sekrit_project: https://my-sekrit-project-vault-url/
```

So `{{ azureKeyVault "my-secret-name" }}` would use the `default` URL and `{{ azureKeyVault "my-secret-name" "sekrit_project" }}` would use the `sekret_project` URL.

assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
```diff
@@ -0,0 +1,25 @@
...
+
+Ensure [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli) is installed and
+[log in](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#azureCLI) with
+a user having the right key vault RBAC permissions.
```
halostatue, line 17: Am I understanding correctly that the Azure CLI is only required for session authentication and that chezmoi will use that session authentication?

I’m not sure how long Azure CLI sessions remain active, so this might be something worth pointing out.

pkg/cmd/azurekeyvaulttemplatefuncs.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type azureKeyVaultConfig struct {
+	VaultURL string `json:"vaultUrl" mapstructure:"vaultUrl" yaml:"vaultUrl"`
+	client   *azsecrets.Client
+	cache    map[string]string
+}
+
+func (c *Config) azureKeyVaultTemplateFunc(name string) string {
+	if secret, ok := c.AzureKeyVault.cache[name]; ok {
```
halostatue, line 17: The cache key *should* be set based on the secret name and the vault URL, especially if the suggestion is accepted to allow multiple vault URLs.

halostatue (commented):
This seems like a reasonably good start.

You should add some tests for this, but I’m not entirely sure how you would do so because the closest analogue to what you’ve implemented is `pkg/cmd/awssecretsmanagertemplatefuncs.go` which also does not have tests.

twpayne:
> You should add some tests for this, but I’m not entirely sure how you would do so because the closest analogue to what you’ve implemented is `pkg/cmd/awssecretsmanagertemplatefuncs.go` which also does not have tests.

I think it's OK to skip tests for this, as the functions are straightforward and it's not easy to mock out the Azure service.

assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
catay: Thanks for the feedback!

> This should _probably_ be a second optional parameter in the event that there are two different projects that need configuring which use different key vaults. `azureKeyVault.vaultUrl` would be used if otherwise unspecified.
> 
> I don’t know what an Azure Key Vault URL looks like, but for the suggestion below, I’m going to assume it’s an HTTPS URL. For the case where there may be multiple key vaults referenced, we may want to make it so that the function accepts either `{{ azureKeyVault "my-secret-name" ["vault-url" | "vault-url-alias"] }}` where the aliases would be defined in the config file under `azureKeyVault.vaultUrls` (plural), which would be a `map[string]string`:

Initially I was also thinking of supporting multiple key vaults but wanted to keep it simple.
I followed the example of the AWS Secrets Manager implementation which also supports only one profile.
Not sure we need this complexity right now as a key vault can hold many secrets.

The Azure Key Vault URL is indeed an HTTPS endpoint like `https://my-personal-key-vault.vault.azure.net/`.
Where the `hostname` is the Azure resource name and needs to be globally unique.

assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
catay: Azure CLI uses OAuth 2.0 and indeed only required for authentication. 
That's the easiest way and most common way to do it.

Alternative ways are also possible with the `DefaultAzureCredential` method which chains all authentication types available.

For more info:

https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#2-authenticate-with-azure

assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
halostatue: I think that this could be trivially extended to support multiple vaults and simplify the general use. Right now, you have to make sure you modify your `config.$FORMAT` file to include a `azureKeyVault.vaultUrl`. If you add a second parameter `{{ azureKeyVault "my-secret-name" ["vault-url"] }}`, then users don’t have to add that configuration before they can do anything else.

We could make `my-personal-key-vault` (no `https://` at the beginning) transform into `https://my-personal-key-vault.vault.azure.net/`, but that feels too magic to begin with and for something that I don’t use.

assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
halostatue: OK. If that‘s the way that it works, that’s the way that it works. It would be nice if there was a way to have the Go SDK prompt for authentication or even open an oauth2 web callback.

pkg/cmd/azurekeyvaulttemplatefuncs.go
halostatue: This is only needed if multiple vaults are permitted.

assets/chezmoi.io/docs/user-guide/password-managers/azure-key-vault.md
halostatue: I’ll leave it up to you as to whether you want to change this to support a second parameter, but I think it would be a good idea (of course, I think that the AWS implementation should *also* be changed, but that’s an entirely different request).

halostatue (approved):
I have things that I would *like* to see, but do not think that we *need* to have for a first version of this.

I think we can merge this tomorrow afternoon if you don’t make the changes I have suggested for multiple vaults, because there’s nothing *wrong* with single vault support.

catay:
Thanks.  Give me until the end of the week to look into the multiple vault support. 
The main complexity I see is with the template function and optional parameters but not impossible.

halostatue:
> Thanks. Give me until the end of the week to look into the multiple vault support.

I’ll ping this on Friday afternoon if you haven’t update this before then with the plan of merging it Friday afternoon. If you prefer, we can merge this and add on the multiple vault support later.

Like I said, nice (because I’m a bit of a completionist for this sort of thing), but not *necessary*.

twpayne:
Could I request a change to the structure? Instead of:



> ```yaml
> azureKeyVault:
>   vaultUrls:
>     default: https://my-default-key-vault-url/
>     sekrit_project: https://my-sekrit-project-vault-url/
> ```

Could you make it:

```
azureKeyVault:
  vaults:
  - url: https://my-default-key-vault-url/
  - name: sekrit_project
    url: https://my-sekrit-project-vault-url/
```

`name` defaults to the empty string, and if no vault is specified in the call to the `azureKeyVault` template function, then the first vault with an empty name should be used.

For handling optional arguments to template functions, see for, example [how they are handled in the `include` template function](https://github.com/twpayne/chezmoi/blob/1f9a7bbadd54358fab7b4f9203a62c7913265108/pkg/cmd/templatefuncs.go#L216-L225).

catay:
Thanks for the input.  Will implement like you suggested.
I was playing with the same approach for the structure and was also looking into using a variadic function for optional parameters.

catay:
After implementing the approach discussed I believe there is more elegant way on how to implement this. 
See below for the rational.

**Note**: 
- I only committed the code for now,  if considered ok i'll fix the documentation tomorrow.
   -  [2950bfe](https://github.com/twpayne/chezmoi/pull/3112/commits/2950bfecc3faab65ef6cc4fb7221b8ae71f25a19) option 1 with the initially proposal.
   - [355a435](https://github.com/twpayne/chezmoi/pull/3112/commits/355a435bbd4a92c11a3e6fc79760eb53f9a64878) with this proposal.
- Should I rebase to a single commit & force push or will you squash merge to master?

An Azure Key Vault comes with a __globally unique hostname__ under the Microsoft owned domain __vault.azure.net__.

The __hostname__ is the name of the resource in the Azure portal and cannot be created if not globally unique.

Examples:

```
kv-weu-t-sec-01.vault.azure.net
kv-random-name-01.vault.azure.net
```

This means we have enough with only hostname / vault name to construct the URL. 

The proposal is using the below structure in the config.

```
azureKeyVault:
  vaults:
  - name: my-vault-01
  - name: my-vault-02
    alias: vault2
```

The parameters for the template function would look like this:

```
{{ azureKeyVault "my-secret-name" ["vault-name" | "vault-alias"] }}
```

Example of different template function implementations.

```
A secret {{ azureKeyVault my-secret }} using default my-vault-01 (no alias set).
A secret {{ azureKeyVault my-secret vault2 }} using my-vault-02.
A secret {{ azureKeyVault my-secret my-vault-03 }} using my-vault-03 vault. 
```

The `name` is the __hostname__ of the vault,  the `alias` is a custom ref that can be used in the template function.  If the `alias` is not set, this will be the default vault in case no vault name or alias is provided in the template function.

If the `alias` is set in the template function it will lookup the correct vault `name`.
If the `alias` set in the template function is not found, it will assume it's a vault `name`.

twpayne:
Thank you, your design is excellent. I really like it.

The main / test-website check failure is unrelated to this PR.

I think it's OK to remove the aliasing functionality as the vault names are short strings and the aliasing can be achieved with other template functions.

halostatue:
I agree with @twpayne about the aliasing not being necessary since there’s a fixed name. I’ll put some comments in-line that I think could be used to simplify this.

pkg/cmd/azurekeyvaulttemplatefuncs.go
```diff
@@ -0,0 +1,97 @@
...
+
+type azureKeyVault struct {
+	Name   string `json:"name"  mapstructure:"name"  yaml:"name"`
+	Alias  string `json:"alias" mapstructure:"alias" yaml:"alias"`
+	client *azsecrets.Client
+	cache  map[string]string
+}
+
+func (v *azureKeyVault) URL() string {
+	return "https://" + v.Name + "." + azureVaultDomain + "/"
```
halostatue, line 21: This could be simplified to `return fmt.Sprintf("https://%s.vault.azure.net/", v.Name)`.

pkg/cmd/azurekeyvaulttemplatefuncs.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type azureKeyVaultConfig struct {
+	Vaults []*azureKeyVault `json:"vaults" mapstructure:"vaults" yaml:"vaults"`
+	cred   *azidentity.DefaultAzureCredential
```
halostatue, line 26: I’ll go a step further than @twpayne about the removal of `alias:`, I would change this to:

```go
type azureKeyVaultConfig struct {
  DefaultVault String `json:"defaultVault" toml:"defaultVault" yaml:"defaultVault"`
  vaults map[string]*azureKeyVault
  cred *azidentity.DefaultAzureCredential
```

`azureKeyVault` can then be simplified to `name`, `client`, and `cache` with no serialization instructions. The `vaults` map is initialized with an entry for `DefaultVault`.

pkg/cmd/azurekeyvaulttemplatefuncs.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (a *azureKeyVaultConfig) GetVaultNameFromAlias(alias string) string {
+	for _, vault := range a.Vaults {
```
halostatue, line 30: Switching to `map[string]*azureKeyVault` means that we can do named lookup rather than having to loop over and append. It’s a small efficiency, but I think worthwhile.

pkg/cmd/azurekeyvaulttemplatefuncs.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (c *Config) azureKeyVaultTemplateFunc(args ...string) string {
+	var secretName, vaultName string
+
+	switch len(args) {
+	case 1:
+		secretName, vaultName = args[0], c.AzureKeyVault.GetVaultNameFromAlias("")
+	case 2:
+		secretName, vaultName = args[0], c.AzureKeyVault.GetVaultNameFromAlias(args[1])
+	default:
```
halostatue, line 92: This would be changed to:

```go
switch len(args) {
case 1:
  secretName, vaultName = args[0], c.AzureKeyVault.DefaultVault
case 2:
  secretName, vaultName = args[0], args[1]
default:
  panic(…)
}
```

halostatue (commented):
This looks amazing.

If we drop aliasing, we lose the value of having a list of defined vaults — and that’s OK, because you’ve confirmed that there’s one `vault.azure.net` base domain for all instances.

My recommendation is to change the exposed configuration to `DefaultVault String` (so `azureKeyVault.defaultVault = "my-default-key-vault"`), but mostly *keep* your `azureKeyVault` structure and functions because they do make most of it much easier.

I’m happy to make a follow-on PR if you’re wanting to be done with this originally simple change.

catay:
The rational behind the alias was the following assuming you use multiple vaults:

Some users do not want to have the real vault name stored in version control and reference the alias defined in the configuration file.  Security through obscurity basically. 

Also trivial,  but if you want migrate to another vault and use aliases in all your templates, you only need to change the name in the configuration. 

But maybe that's just overkill. :)

If we decide to drop the aliases we probably also can simplify the structure back to this.

```
azureKeyVault:
  defaultVault: my-vault-02
```

We only have to specify a default and support for another vault would be the second parameter of the template function.

My preference is to stick with the layout containing the alias as it leaves more flexibility for later if you need to add an additional option for a specific vault.

halostatue:
I have no strong feelings either way on this, and your rationale is sound about the flexibility:

> My preference is to stick with the layout containing the alias as it leaves more flexibility for later if you need to add an additional option for a specific vault.

If, like with AWS, you are encouraged to have more than one Azure account or profile for separation of concerns, then I could see *later* adding a configuration option to identify the account or credential identifier. If this is *possible*, then we should keep the current structure even if we just drop `alias` support.

With respect to aliasing, it is possible to do that with `data` references:

```yaml
azureKeyVault:
  defaultVault: my-default-vault

data:
  akv:
    innocuous: take-over-the-world
```

```
{{ azureKeyVault "system-password" .akv.innocuous }}
```

This is *sort* of what I do with some of my secrets, as [home/private_dot_config/gh/private_dot_hosts.yml.tmpl](https://github.com/halostatue/dotfiles/blob/main/home/private_dot_config/gh/private_hosts.yml.tmpl) shows.

pkg/cmd/azurekeyvaulttemplatefuncs.go
catay: Completed.

pkg/cmd/azurekeyvaulttemplatefuncs.go
catay: Implemented.

pkg/cmd/azurekeyvaulttemplatefuncs.go
catay: Removed this function as not required anymore.

pkg/cmd/azurekeyvaulttemplatefuncs.go
catay: Implemented, but added an additional check in case no default is provided in the config.

catay:
Please check out the latest commit. 

The summary is that I basically removed the alias approach and simplified it more based on the provided feedback.
Also all the documentation is now updated to the latest design. 

I think it's ready to get  merged to main now.

Thanks for the really good feedback and patience.

assets/chezmoi.io/docs/reference/templates/azure-key-vault-functions/azureKeyVault.md
```diff
@@ -0,0 +1,15 @@
...
+The mandatory `secret name` argument specifies the *name of the secret* to
+retrieve.
+
+The optional `vault name` argument specifies the *name of the vault*, if not set,
+the default vault name will be used.
+
+!!! warning
+
+    The current implementation will always return the latest version of the secret.
+    Defining a specific version of a secret is not supported.
```
halostatue, line 15: I would use `Obtaining` or `Retrieving` instead of `Defining`.

pkg/cmd/azurekeyvaulttemplatefuncs.go
```diff
@@ -0,0 +1,90 @@
...
+	return *resp.Value
+}
+
+func (c *Config) azureKeyVaultTemplateFunc(args ...string) string {
+	var secretName, vaultName string
+
+	switch len(args) {
+	case 1:
+		if c.AzureKeyVault.DefaultVault == "" {
+			panic(fmt.Errorf("no default vault set"))
```
halostatue, line 80: Maybe `"no value set in azureKeyVault.defaultVault"`?

halostatue (approved):
This is very good and you’ve been quite patient with a picky code review.

I have suggested a couple of small wording changes that aren’t urgent or required, but may improve the documentation / UX.

assets/chezmoi.io/docs/reference/templates/azure-key-vault-functions/azureKeyVault.md
catay: Completed.

pkg/cmd/azurekeyvaulttemplatefuncs.go
catay: Completed.

twpayne:
Thank you very much for this excellent code @catay and for the thorough reviews @halostatue!

The main / test-website build error is https://github.com/mkdocs/mkdocs/issues/3310 / https://github.com/fralau/mkdocs-mermaid2-plugin/issues/80 and unrelated to this PR.

I'll merge this when the remaining checks are complete.

catay:
Thanks.

I just pushed a PR for the workflow fix but it seems you beat me to it  :)

halostatue:
@catay I’m not sure what the "This branch cannot be rebased due to too many changes", but I *suspect* that if you were to rebase this and squash the various commits together, we can have one good commit that can be merged/rebase-merged cleanly.

catay:
Rebase branch on master done.

PR #3119, opened 2023-07-27
chore: Pin mkdocs-material version to avoid incompatibility in latest

twpayne:


twpayne:
Closing in favor of #3120.

PR #3128, opened 2023-07-29
fix: Make detection of mackup config dir more robust

twpayne:
@eugenesvk this should resolve https://github.com/twpayne/chezmoi/issues/885#issuecomment-1656728570.

### DIFF
--- a/pkg/cmd/mackupcmd_darwin.go
+++ b/pkg/cmd/mackupcmd_darwin.go
@@ -16,11 +16,10 @@ import (
 )
 
 var (
-	mackupCommentRx           = regexp.MustCompile(`\A#.*\z`)
-	mackupKeyValueRx          = regexp.MustCompile(`\A(\w+)\s*=\s*(.*)\z`)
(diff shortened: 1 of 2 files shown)


halostatue:
I know that this isn’t documented, but it is probably worth noting *somewhere* that it only works with homebrew-installed Mackup.

I haven’t had time to play with it for various reasons, but I was looking at creating a Python package that could install Mackup in a venv and use Mackup as a library to get the configuration values, since Mackup is broken by default if you’re not using the atrocious Dropbox Desktop application (I use Maestral).

It might also be possible to use an archive of the mackup repo as a source of truth.

twpayne:
Thanks for the ideas. Ideally Mackup would provide a command to expose its database, but the Mackup developers didn't seem very interested in collaborating when I asked them.

As Mackup doesn't have a huge number of users, and only a handful of users will use both chezmoi and mackup, I want to keep the integration as simple as possible. For reference, according to `brew info`, Mackup had 304 install-on-demands in the last 30 days, compared to chezmoi's 2818.

Issue #3132, opened 2023-07-31
[feature request] Doppler support


equals03:
## Is your feature request related to a problem? Please describe.

chezmoi includes support for HCP Vault Secrets using the `vlt` CLI to expose data through the `hcpVaultSecret` and `hcpVaultSecretJson` template functions. It does not, however, support [Doppler](https://www.doppler.com/). The feature set and implementation is _almost_  1:1 between `vlt` and the [Doppler CLI](https://docs.doppler.com/docs/cli).  

## Describe the solution you'd like

The Doppler CLI to be implemented.

## Describe alternatives you've considered

Using `vlt` but I like Doppler :smile: 

## Additional context

I would be willing to give implementation a stab (after all it is basically a copy and paste of the `vlt` implementation) time permitting and should this get enough interest/align with the project.

twpayne:
Please do implement this :)

https://github.com/twpayne/chezmoi/pull/3067/files should give you an idea of what files need to be added/modified.

PR #3134, opened 2023-07-31
feat: Add support for Doppler

equals03:
fixes https://github.com/twpayne/chezmoi/issues/3132

Should be good for a review.


assets/chezmoi.io/docs/reference/templates/doppler/dopplerJson.md
```diff
@@ -0,0 +1,15 @@
+# `dopplerJson` *key* [*application-name* [*project-id* [*organization-id*]]]
+
+`dopplerJson` returns the plaintext secret from [Doppler](https://www.doppler.com) using `doppler
+secrets get --plain`, however it is a requirement that this secret be structured as `json`.
```
halostatue, line 4: I don’t use Doppler, so I can’t be 100% sure that this is right, but I would change this function as described in the [secrets access guide](https://docs.doppler.com/docs/accessing-secrets#getting-values) for getting multiple secrets: `doppler secrets get DOPPLER_PROJECT DOPPLER_CONFIG --json --no-file` or `doppler secrets download`.

Otherwise, I think we should drop `dopplerJson`, because it could be replaced with `{{ doppler key project organization | fromJson }}`.

internal/cmd/dopplertemplatefuncs.go
```diff
@@ -0,0 +1,132 @@
...
+	outputCache map[string][]byte
+}
+
+var dopplerMinVersion = semver.Version{
+	Major: 0,
+	Minor: 0,
+	Patch: 0,
+} // no specific version required (as of yet)
+
+func toUpperSnakeCase(s string) string {
```
halostatue, line 31: This feels too magic to me and should either (a) not do any transformations (can any doppler secret names be lowercase? if not, I think that’s a bug on their side) or (b) only do uppercase transformations. We should not treat `"foo bar"`, `"foo!bar"`, and `"foo_bar"` the same, and since `foo_bar` is a valid shell variable name, we probably shouldn’t treat it the same as `FOO_BAR`.

halostatue (commented):
This looks like it would work, but I think that there are two design choices that should be changed.

twpayne:
Thank you @equals03, this is fantastic! Thank you for such an excellent pull request!

+1 on @halostatue's comments. For info, the `passwordManagerJson` template functions were either added before the `fromJson` template function was added, or still make sense when the JSON output contains significantly different data to the plain text output. If my understanding is correct, `fromJson` does the job of `dopplerJson` here. Please correct me if this is not the case.

We can also drop the `dopplerMinVersion` variable for now and re-add it later if needed.

halostatue:
> If my understanding is correct, `fromJson` does the job of `dopplerJson` here. Please correct me if this is not the case.

That's my understanding of the implementation, too. When I was looking at the Doppler documentation, it appears that *unlike* `vlt`, there is a way to get all of the secrets for a given project / config at once, so `{{ dopplerJson [project [config]] }}` could be used for a more efficient project configuration retrieval as well.

It might be better to call that `dopplerProjectJson [project [config]]`, but I think that it is potentially very useful. Using a JSON-shaped value inside that JSON will require the application of `| fromJson`, but I think that’s fine.

And just to be clear, I really appreciate the work that @equals03 put into this and it’s really good work.

assets/chezmoi.io/docs/reference/templates/doppler/dopplerJson.md
equals03: sounds good :smile:  
I'll implement it so that `dopplerJson` pulls all the secrets for the project and give it a few tests. Obviously this would change the template API to:  
```{{ (dopplerJson "project" "config").PASSWORD }}```

File: internal/cmd/dopplertemplatefuncs.go (hunk 2 above)
Comment (equals03) [reply internal/cmd/dopplertemplatefuncs.go:31]: I went backwards and forwards on this one a bit. Doppler, itself, does not support any project secrets not in `UPPER_SNAKE_CASE`. I believe its because they have different formats they want to adhere to (most notably `.env`)

```console
> doppler secrets get --project chezmoi --config dev --plain "not upper snake case"
Unable to fetch secrets
Doppler Error: Secret name "not upper snake case" may only contain uppercase letters, numbers, and underscores.
```

I built this in as a convenience. I can remove if its going to cause too much confusion?

equals03:
Thanks for all the positive feedback @halostatue @twpayne! :+1: 

Loving the idea of switching `dopplerJson` to `dopplerProjectJson [project [config]]`. I'll make it happen.
As for the `minVersion` I did notice there was a breaking change in release [3.25.0](https://github.com/DopplerHQ/cli/releases/tag/3.25.0) that I thought we may have to peg it to, but I don't think that will be required. I'll remove and we can always add it back later if needs be.

Also, side  note: I wrote the windows tests blind as I don't have access to a windows machine atm. I'm going to try a subtle change (when I update for the new template funcs) and see if this one passes. If not, I may need someone to help on those. :smile:

assets/chezmoi.io/docs/reference/templates/doppler/dopplerJson.md
equals03: UPDATE: I'll change it to `dopplerProjectJson` as mentioned below [here](https://github.com/twpayne/chezmoi/pull/3134#issuecomment-1659663407)

assets/chezmoi.io/docs/reference/templates/doppler/dopplerProjectJson.md
```diff
@@ -0,0 +1,15 @@
+# `dopplerProjectJson` [*project* [*config*]]
+
+`dopplerProjectJson` returns all the secrets for the specified project and configuration
+from [Doppler](https://www.doppler.com) using `doppler secrets download --json --no-file`
+as `json` structured data.
+
+If any of *project*, or *config* are empty or
```
halostatue, line 7: I prefer:

> If either of *project* or *config* are empty or omitted, then…

internal/cmd/dopplertemplatefuncs.go
halostatue: I think that it will be better to remove it, especially with the `dopplerProjectJson` function returning the JSON of all secrets in a project and therefore only supporting `UPPER_SNAKE_CASE`.

assets/chezmoi.io/docs/reference/templates/doppler/doppler.md
```diff
@@ -0,0 +1,15 @@
+# `doppler` *key* [*project* [*config*]]
+
+`doppler` returns the plaintext secret from [Doppler](https://www.doppler.com) using `doppler
+secrets get --plain`.
+
+If any of *project*, or *config* are empty or
+omitted, then chezmoi will use the value from the
+`doppler.project`, and
+`doppler.config` config variables if they are set and not empty.
```
halostatue, line 9: I mentioned the wording on `dopplerProjectJson`, but the same pattern would be used here.

assets/chezmoi.io/docs/reference/templates/doppler/doppler.md
```diff
@@ -0,0 +1,15 @@
...
+
+!!! example
+
+    ```
+    {{ doppler "USERNAME" }}
+    ```
```
halostatue, line 15: I think that it could be *worthwhile* mentioning JSON-structured secrets further decoding with `fromJson`. I don’t know that it’s required, but this is one of those complexities of chezmoi in that it is essentially endlessly composable and such compositions may not be always visible.

halostatue (changes requested):
This looks very good overall and is I think very close to complete.

assets/chezmoi.io/docs/reference/templates/doppler/doppler.md
equals03: I do mention it in the [user-guide](https://github.com/twpayne/chezmoi/pull/3134/files#diff-fa42ea18298385828ab9d687dc4866fdf6f5018e74ba5b4cd821b6e46edd4a24R50). I didn't think it would be the right place to mention this here - on each function. Otherwise where do you draw that line? Would all new integrations have to list any potential compositions on all function references?  

Thoughts?

halostatue: I had somehow missed that. (small voice)never mind(/small voice)

equals03:
@halostatue I've made some updates :smile:  

Most notably, while implementing the feedback, I did a bit of performance testing and realised there was no good reason not to "unify" the pipelines for both commands. The now use the same mechanism to get the secrets and can make use of the output cache.

In my testing this cuts down on individual calls, dramatically, has quite a performance boost and ensures that the results from both functions are consistent.

As such, I took another stab at rewriting the docs for the user guide. 

Also, I have removed that "magic" as you suggested.

halostatue (approved):

halostatue:
> Most notably, while implementing the feedback, I did a bit of performance testing and realised there was no good reason not to "unify" the pipelines for both commands. The now use the same mechanism to get the secrets and can make use of the output cache.
> 
> In my testing this cuts down on individual calls, dramatically, has quite a performance boost and ensures that the results from both functions are consistent.

I *think* that this is OK, but it is worth noting that from a "least privilege" security stance, you are now downloading and caching in memory secrets that may not be used in the chezmoi process by using `download` all the time. I’m not sure that this requires changing anything, but it is worth noting.

> As such, I took another stab at rewriting the docs for the user guide.

I didn’t comment on the lines, but there is an extra comma after `doppler.project` in both command documents; it’s a small thing.

internal/cmd/testdata/scripts/doppler.txtar
```diff
@@ -0,0 +1,151 @@
...
+echo.  "PASSWORD": "default-project-password"
+echo.}
+exit /b 0
+
+:download-other-project
+echo.{
+echo.  "DOPPLER_CONFIG": "default-config",
+echo.  "DOPPLER_ENVIRONMENT": "default-config",
+echo.  "DOPPLER_PROJECT": "other-project",
+echo.  "PASSWORD": "default-other-password"
```
twpayne, line 143: This needs to be `other-project-password`.

twpayne (commented):
Thank you, this is excellent.

A small change is needed for the tests to pass on Windows.

Once you've fixed that, please could you squash all the commits together. Then it's ready to ship :)

internal/cmd/testdata/scripts/doppler.txtar
equals03: rats. i had a feeling i missed one! i need to get windows set up for things like this...

equals03:
@twpayne should be fixed now and commits squashed.

@halostatue 
> I _think_ that this is OK, but it is worth noting that from a "least privilege" security stance, you are now downloading and caching in memory secrets that may not be used in the chezmoi process by using `download` all the time. I’m not sure that this requires changing anything, but it is worth noting.

Honestly, I went through the same thought process BUT I landed on: "this is OK" because of the following:
1. we already allow the download of all the secrets through `dopplerProjectJson`
2. as far as "attack vectors" go, the CLI is already available on the machine. It would probably just be easier to exploit it via the CLI
3. from a more practical point of view, you should probably only keep the secrets you are wanting to use in chezmoi in the same project - no more & no less. :smile: essentially, they should be there because you are going to use them.

halostatue:
Thank you very much for your contribution, @equals03!

Issue #3141, opened 2023-08-02
`lookPath` should accept an optional `path` override


arran4:
## Is your feature request related to a problem? Please describe.

As part of my dotfiles script I'm building up a list of paths that will replace the existing paths as a result I want not just the "current state" of paths, but if something is found in the paths that I'm building up. One way to do this would be to provide an optional argument to `lookPath` where I could specify the value of `$PATH` that I'm building up

## Describe the solution you'd like

```
{{ if lookPath "diff-so-fancy" }}
# diff-so-fancy is in $PATH
{{ end }}
```

Would work along with:

```
{{ if lookPath "diff-so-fancy" data.futurepath }}
# diff-so-fancy is in future $PATH
{{ end }}
```

## Describe alternatives you've considered

I guess I could loop through it and use `stat` to build up a map of what's in there for a quick lookup later. But it seems slower.

twpayne:
Two threads, maybe not solving this issue:

1. Can template logic work? For example, here's [how I add paths](https://github.com/twpayne/dotfiles/blob/f25b14407ec63b79b8c84a3a3159431fc4c02c24/home/dot_zshrc.tmpl#L9-L19).
2. [chezmoi caches path look-ups](https://github.com/twpayne/chezmoi/blob/547de5b86f0c0da185dd663ded727d246e5e6a91/internal/chezmoi/lookpath.go). Maybe looking up a path is not expensive?

arran4:
I am doing something similar to:
https://github.com/twpayne/dotfiles/blob/f25b14407ec63b79b8c84a3a3159431fc4c02c24/home/dot_zshrc.tmpl#L9-L19

That's exactly what I imagined it to look like:

https://github.com/twpayne/chezmoi/blob/547de5b86f0c0da185dd663ded727d246e5e6a91/internal/chezmoi/lookpath.go#L15-L28

My issue is one step further; I would like to pick tools based on what's now available in paths. For instance installing on Solaris often puts tools in `/opt/` which wouldn't be picked up with the current `lookPath` once I have modified path..

So:

```text/template

{{- $paths := list }}
{{- $homeDir := .chezmoi.homeDir }}
{{- range $_, $relPath := list "bin" "go/bin" ".cargo/bin" ".local/bin" }}
{{    $path := joinPath $homeDir $relPath }}
{{-   if stat $path }}
{{-     $paths = mustAppend $paths $path }}
{{-   end }}
{{- end }}
{{- if $paths }}
[...]

PR #3148, opened 2023-08-03
Adds `lookPathIn`

arran4:
Resolves: https://github.com/twpayne/chezmoi/issues/3141

As per: https://github.com/twpayne/chezmoi/issues/3141

Adds `lookPathIn`

Not sure the testing procedure. Followed: https://www.chezmoi.io/developer/contributing-changes/ 

Microcommiters hate: https://www.conventionalcommits.org/en/v1.0.0/#summary -_-

### DIFF
--- a/internal/cmd/testdata/scripts/templatefuncs_windows.txtar
+++ b/internal/cmd/testdata/scripts/templatefuncs_windows.txtar
@@ -1,0 +1,14 @@
+[!windows] skip 'Windows only'
+
+# Couldn't figure out why this works locally but not in github actions
+# # test lookPathIn template function to find in specified script - success with extension
+# exec chezmoi execute-template '{{ lookPathIn "git.exe" "c:\\windows\\system32;c:\\windows\\system64;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0;C:\\Program Files\\Git\\cmd" }}'
+# stdout '^C:\\Program Files\\Git\\cmd\\git.exe$'
+#
+# # test lookPathIn template function to find in specified script - success without extension
+# exec chezmoi execute-template '{{ lookPathIn "git" "c:\\windows\\system32;c:\\windows\\system64;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0;C:\\Program Files\\Git\\cmd" }}'
+# stdout '^C:\\Program Files\\Git\\cmd\\git.exe$'
+#
+# # test lookPathIn template function to find in specified script - failure
+# exec chezmoi execute-template '{{ lookPathIn "asdf" "c:\\windows\\system32;c:\\windows\\system64;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0;C:\\Program Files\\Git\\cmd" }}'
+# stdout '^$'
--- a/assets/chezmoi.io/docs/reference/templates/functions/lookPathIn.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/lookPathIn.md
@@ -1,0 +1,39 @@
+# `lookPathIn` *file* *paths*
+
+`lookPathIn` searches for an executable named *file* in the directories provided by
+the `paths` parameter using the standard OS way of separating the PATH environment
+variable. The result may be an absolute path or a path relative to the current directory.
+If *file* is not found, `lookPathIn` returns an empty string.
+
+If the OS is Windows `lookPathIn` will either: if there is an extension, check to see if
+the extension is specified in the `PathExt` environment variable. If there isn't an
+extension it will try each of the extensions specified in the `PathExt` environment
+variable in the order provided until it finds one. In either case if it doesn't `lookPathIn`
+moves onto the next path provided in the `paths` parameter.
+
+`lookPathIn` is provided as an alternative to `lookPath` so that you interrogate the
+paths as you would have them.
+
+Each successful lookup is cached based on the full path, and evaluated in the correct
+order each time to reduce `File Stat` operations.
+
+!!! example
+
+    ```
+        {{- $paths := list }}
+        {{- $homeDir := .chezmoi.homeDir }}
+        {{- range $_, $relPath := list "bin" "go/bin" ".cargo/bin" ".local/bin" }}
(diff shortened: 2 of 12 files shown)


arran4:
@twpayne Need to figure out how to make this work on windows too.

arran4:
@twpayne Please approve running now.

internal/cmd/testdata/scripts/templatefuncs_windows.txtar
```diff
@@ -0,0 +1,9 @@
+[!windows] skip 'Windows only'
+
+# test lookPathIn template function to find in specified script - success
+exec chezmoi execute-template '{{ lookPathIn "winver" "c:\windows" }}'
+stdout ^/bin/echo
+
+# test lookPathIn template function to find in specified script - failure
+exec chezmoi execute-template '{{ lookPathIn "winver" "c:\users" }}'
```
twpayne, line 8: This probably also needs another backslash.

arran4: Yep you're right. I went in blind. I have identified an actual issue too with this test.

arran4:
@twpayne Okay I did what I could and I'm probably going to abandon efforts to get the windows test working. The unit tests show that it does work, there is just something odd about the txtar which is causing an issue.

halostatue:
This is good work, and I love the *idea* of this function, and I’m more than happy to give a detailed review, but… Right now I don’t love *this function*, and many of the comments that I would make would be covered by the concerns from this. In particular, I don’t love the fact that it’s expecting an OS-specific PATH list. (As an aside, that's a `{,ba,z,k}sh`-specific thing. In Fish, `$PATH` is a list, not a colon-separated list.)

```
{{ $found := lookPathIn "something" "foo:bar:baz" }}
{{/* On Windows */}
{{ $found := lookPathIn "something" "foo;bar;baz" }}
```

I think that we should *either* embrace a variadic form (`lookPathInTemplateFunc(needle string, ...haystack string) string`), an explicit array form (`lookPathInTemplateFunc(needle string, haystack string[]) string`), or something more dynamic and in-between (`lookPathInTemplateFunc(needle string, ...haystack any) string`).

These would allow:

```
{{ $found := lookPathIn "something" "foo" "bar" "baz" }}{{/* variadic, dynamic */}}
{{ $found := lookPathIn "something" (list "foo" "bar" "baz") }}{{/* array, dynamic */}}
```

The hard part about the dynamic form would be the error handling if given a non-string or non-string-array. Other questions about the dynamic form:

- would we allow mixed strings and string-arrays?
- would we base our assumption of what the rest of the values are based on the first value?

With the way it’s implemented now, I think that it’s really confusing — and you’re going to end up calling `file.splitList` *anyway*, so why force people to join it in a platform-specific way just to split it apart again later? The list character is also not exposed in any way to chezmoi templates, so that becomes knowledge that people Just Have To Know (that *might* be a useful function `joinOsList` or something, but…).

My preference is for dynamic, either all strings or one array (which still requires `...haystack any`), because that allows me to build a list and then look, or to specify a list as parameters. If we don’t want the complexity of dynamic resolution, then I would make it explicitly two parameters (`lookInPath "needle" (list "straw" "straw" "straw")`) for an array.

What do you guys think?

arran4:
I think if you're not assuming a platform standard it should potentially be something else. The value add are the assumptions as you take those assumptions away the work involved to use it increases. While it's a good point it doesn't help with `fish` I'm not sure what we can do in the scope of this function. I feel to match yours a separate feature set (if it doesn't already exist) is required to add functional programming like operators and expose `SplitList` if it hasn't already

halostatue:
> I think if you're not assuming a platform standard it should potentially be something else. The value add are the assumptions as you take those assumptions away the work involved to use it increases. While it's a good point it doesn't help with `fish` I'm not sure what we can do in the scope of this function. I feel to match yours a separate feature set (if it doesn't already exist) is required to add functional programming like operators and expose `SplitList` if it hasn't already

But this function *isn’t* using PowerShell, bash, fish, or cmd to find executables, it’s using a customized copy of a Go builtin to search a list of paths…that we have to split because we are requiring that it be joined in the first place. It’s better to think of this outside of the OS box and instead think of it in terms of what we want to do and how we want to use this. I, for one, do not want to use this by having to remember to join with `":"` or `";`" depending on platform, and it feels like a usability regression over running `chezmoi` twice at that point.

- The *easiest* thing for simple use would be to accept `...haystack string`, because it’s then `{{ lookPathIn "needle" "straw" "straw" "straw" }}`.
- The *easiest* thing for template reuse would be to accept `haystack []string`, because it’s then `{{ lookPathIn "needle" $haystack }}` (where `$haystack` is a previously created list).

We *can* do both, but that slightly increases the `lookPathInTemplateFunc` definition complexity:

```go
func lookPathInTemplateFunc(needle string, haystack ...any) string {
  if len(haystack) == 0 {
    panic(fmt.Errorf("expected at least one haystack parameter"))
  }

  switch haystack[0].(type) {
    case string:
      return chezmoi.lookPathIn(needle, haystack)
    case []string:
      return chezmoi.lookPathIn(needle, haystack[0])
    default:
      panic(fmt.Errorf("invalid haystack type, must be string or []string"))
  }
}
```

At that point, there’s no need for joining, no need to expose what the list character is, and no need for `chezmoi.lookPathIn` to split the characters. It’s also completely shell and platform independent from the template user perspective, which is the main thing that I think matters.

twpayne:
I agree that it makes more sense for the haystack to be a list rather than a string. chezmoi already has a [`splitList` template function](http://masterminds.github.io/sprig/string_slice.html). Accepting exactly one haystack parameter is also fine: there are several template functions for combining lists if needed.

arran4:
I keep forgetting the sprig reexport exists. I will create new commits in new PRs for the broken up functions. The name of the function would not be appropriate as it's really just a minimal modification of the [`lookPath`](https://www.chezmoi.io/reference/templates/functions/lookPath/) [re-exported](https://cs.opensource.google/go/go/+/refs/tags/go1.20.7:src/os/exec/lp_unix.go;l=52) function allowing you to provide your own. So I will choose a different name. I am thinking; `findExecutableByName` or something along those lines. Let me know if you have any opinions.

Issue #2130, opened 2022-06-12
Add Debian package


twpayne:
## Is your feature request related to a problem? Please describe.

It would be great if chezmoi were included in Debian's packaging system as this would make chezmoi more easily installable for users of Debian-based distributions, particularly those in corporations who use Debian-based distributions but with strict IT departments who do not let users run non-allowlisted binaries on their machines.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012721 tracks the request for packaging in Debian.

## Describe the solution you'd like

I'd like to be able to run

```console
$ apt-get install chezmoi
```

on my Debian-based distribution and have it install chezmoi.

## Describe alternatives you've considered

* Installing chezmoi via any of the existing installation methods.
* Using `chezmoi archive` to generate my dotfiles for my corporate-controlled machine (#1410).

twpayne:
A quick update on this as this issue is now over a year old.

I would love to see chezmoi in Debian. Sadly, my understanding is that Debian's approach to packaging Go binaries (currently [documented here](https://go-team.pages.debian.net/packaging.html)) means that Debian will never be able to support modern software like Kubernetes or chezmoi which ship a statically-linked binary with updates moving faster than Debian itself.

So, with a heavy heart, I'm closing this.

twpayne:
Re-opening this as there is now a [discussion on the debian-go list](https://lists.debian.org/debian-go/2023/08/msg00000.html).

twpayne:
cc @ryanakca 

Following the [discussion on the debian-go list](https://lists.debian.org/debian-go/2023/08/maillist.html#00000), it seems that chezmoi could possibly be included with a mix of vendored and system packages. #3151 adds vendored packages, but I'll only merge this if it looks like the Debian package will proceed.

twpayne:
With no response from Debian, I'll close this.

PR #3151, opened 2023-08-04
chore: Vendor dependencies

twpayne:
Refs #2130.


twpayne:
I have absolutely no idea why Github Actions is running the `govulncheck / misspell` job. This makes no sense. govulncheck and misspell are in separate workflows and there is no link between them that I can determine.

Otherwise, I think the code in this PR is correct.

twpayne:
I'll close this for now, as there's been no further response from the Debian Go maintainers. We can re-open this PR if/when needed.

PR #3152, opened 2023-08-05
fix: Implement some logic for `isExecutable` for windows

arran4:
This is based on:
* https://www.nextofwindows.com/what-is-pathext-environment-variable-in-windows
* It goes back to at least Windows XP: https://ss64.com/nt/syntax-variables.html potentially further

fix: Provide some logic for `isExecutable` under windows for determining if it's an executable

BREAKING CHANGE: `isExecutable` no longer always returns false and tries to determine execution status by file extension

### DIFF
--- a/internal/chezmoi/chezmoi_windows.go
+++ b/internal/chezmoi/chezmoi_windows.go
@@ -2,13 +2,37 @@ package chezmoi
 
 import (
 	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
 )
 
 const nativeLineEnding = "\r\n"
 
-// isExecutable returns false on Windows.
+var pathExt []string = nil
+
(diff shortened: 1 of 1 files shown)


halostatue (commented):
This looks good to me. I’m going to defer approval to @twpayne because it is a breaking change, but I think that it’s a *sensible* change. I do not expect that anything is depending on `isExecutable` returning false on Windows, but [Hyrum’s law](https://www.hyrumslaw.com) always applies.

arran4:
I think it is primarily used for one of the source control systems. In which case it might need to be split.

> Defining the correct level of abstraction is a completely separate discussion (see Mythical Man-Month)

I don't remember that being in the book, but means I can refer to it a lot more. The 70s man..

twpayne:
Thanks, this makes sense. There are few minor changes I'd like to do to make the control flow more compact, but I'll do this in a separate PR.

arran4:
Thanks I will read.

PR #3153, opened 2023-08-05
feature: Exposes `isExecutable` to the user

arran4:
Exposes `isExecutable` to the user. I'm not sure if this is consistent as there are other functions that should be exported, and `fileExists` is expected to be done with `stat` from what I can tell.


halostatue (approved):
This also looks good to me, although I still think that `findExecutableIn needle haystack` would be Good To Have, although it becomes a fairly simple extension over the exposed `isExecutable` template function.

arran4:
I did some testing, it is possible for us to bridge: https://pkg.go.dev/golang.org/x/exp/slices into it. Given: https://github.com/golang/go/issues/57433

assets/chezmoi.io/docs/reference/templates/functions/isExecutable.md
```diff
@@ -0,0 +1,11 @@
+# `isExecutable` *file*
+
+`isExecutable` returns true if a file is executable.
+
+!!! example
+
+    ```
+        {{ if isExecutable "/bin/echo" }}
+            `sh` is executable
```
twpayne, line 9: This should be `echo is executable`. You should also remove the backticks around sh as this is text or shell, not Markdown.

twpayne:
> I did some testing, it is possible for us to bridge: https://pkg.go.dev/golang.org/x/exp/slices into it. Given: [golang/go#57433](https://github.com/golang/go/issues/57433)

I don't think this is possible as `text/template` does not support functions as values (assuming you were expecting to use `slices.ContainsFunc`).

twpayne:
Superseded by #3157.

arran4:
I have a sample project with it where I'm able to do that. I haven't extensively tested it. But I got it working. However I suspect it will have significant issues with generics until the base libraries such as `text/template` are updated to support it. (I am assuming they have not been.)

Was the reason for re-authoring the commit in a new PR for speed and to keep it to a single commit?

assets/chezmoi.io/docs/reference/templates/functions/isExecutable.md
arran4: Yes you're correct on both accounts. I would fix it but I think I missed the boat on that.

twpayne:
> Was the reason for re-authoring the commit in a new PR for speed and to keep it to a single commit?

Mostly speed. I have some vacation coming up and want to get a new version of chezmoi out before I leave. You've been very responsive, but it's also been quite a lot of discussion over the last few days and we're in very different timezones.

arran4:
Having issues sleeping helped with the responsiveness and timezone not with the attention to detail in communication or code. 

Enjoy your vacation. If I put through another PR I'm not expecting a speedy response.

PR #3156, opened 2023-08-06
chore: Tidy up implementation if isExecutable on Windows

twpayne:
Refs #3152, cc @arran4.

Rationale for changes:
* Computing the slice of extensions is cheap, so we don't need to do it lazily when first required. Instead, we can do it on startup. This also avoids any potential races. chezmoi executes templates concurrently when reading the source state.
* Windows environment variables are case-insensitive, so use the all-uppercase name as that seems to be the canonical form.
* We can simplify the logic by returning early as soon as we know that the file is not an executable.

### DIFF
--- a/internal/chezmoi/chezmoi_windows.go
+++ b/internal/chezmoi/chezmoi_windows.go
@@ -5,34 +5,28 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+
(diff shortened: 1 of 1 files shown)


arran4:
Yep that's fair. Sorry about that all stuff I should have done.

PR #3157, opened 2023-08-06
feat: Add isExecutable template function

twpayne:
Supersedes #3153. Note author is still credited as @arran4.

Changes compared to #3153 
* Includes #3156.
* Minor doc fix.
* Make code follow chezmoi's style more closely.
* Makes commit messages follow chezmoi's guidelines, as described in [the contributing guide](https://www.chezmoi.io/developer/contributing-changes/).

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/isExecutable.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/isExecutable.md
@@ -1,0 +1,11 @@
+# `isExecutable` *file*
+
+`isExecutable` returns true if a file is executable.
+
+!!! example
+
+    ```
+    {{ if isExecutable "/bin/echo" }}
+    # echo is executable
+    {{ end }}
+    ```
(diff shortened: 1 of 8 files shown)


twpayne:
txtar pretty much only cuts up files between the `-- filename --` markers. Everything before the first marker is a comment (which testcript interprets as a script to run). That's pretty much it.

A common problem on Windows is that the txtar file will contain UNIX line endings, whereas many Windows programs want Windows line endings. There's a `unix2dos` command that fixes the line endings after.

twpayne:
BTW, it's pretty horrible trying to debug Windows problems if you don't have access to a Windows machine. The round trip time of going via chezmoi's Github Actions is very painful. I do have a Windows machine with me and will have a look.

arran4:
Thanks. I wonder if it was just the `unix2dos`. Could `.gitattributes` be used on files matching `*_windows.txtar` or some such? Ie `_windows.txtar text eol=crlf` although we have yet to verify that is in-fact the issue.

PR #3158, opened 2023-08-06
chore: Fix isExecutable tests on Windows

twpayne:
Refs #3157. cc @arran4 

The issue was that the `PATHEXT` environment variable was not set in the tests. In the end, it wasn't a line ending issue because the contents of the files are never used.


arran4:
Ah that makes sense. Didn't know those were managed, would have never dug deep enough to find out.

PR #3164, opened 2023-08-09
feat: Add gitHubReleases and gitHubTags

halostatue:
Adds a feature requested in discussion #3161.

- This reworks `gitHubLatestTag` to use `gitHubTags` and share the caching between the two functions.

- Pagination for both (both size and page number) is not configurable. This will always return the first page of items, where the size of the page is the default specified by GitHub. That is currently thirty items, but is subject to change.

### DIFF
--- a/internal/cmd/githubtemplatefuncs.go
+++ b/internal/cmd/githubtemplatefuncs.go
@@ -25,15 +25,21 @@ type gitHubLatestReleaseState struct {
 	Release     *github.RepositoryRelease `json:"release"     yaml:"release"`
 }
 
-type gitHubLatestTagState struct {
-	RequestedAt time.Time             `json:"requestedAt" yaml:"requestedAt"`
-	Tag         *github.RepositoryTag `json:"tag"         yaml:"tag"`
+type gitHubReleasesState struct {
+	RequestedAt time.Time                   `json:"requestedAt" yaml:"requestedAt"`
+	Releases    []*github.RepositoryRelease `json:"releases"    yaml:"releases"`
+}
+
+type gitHubTagsState struct {
+	RequestedAt time.Time               `json:"requestedAt" yaml:"requestedAt"`
+	Tags        []*github.RepositoryTag `json:"tags"        yaml:"tags"`
 }
 
 var (
 	gitHubKeysStateBucket          = []byte("gitHubLatestKeysState")
 	gitHubLatestReleaseStateBucket = []byte("gitHubLatestReleaseState")
-	gitHubLatestTagStateBucket     = []byte("gitHubLatestTagState")
+	gitHubReleasesStateBucket      = []byte("gitHubReleasesState")
+	gitHubTagsStateBucket          = []byte("gitHubTagsState")
 )
 
 type gitHubData struct {
@@ -41,7 +47,8 @@ type gitHubData struct {
 	clientErr          error
 	keysCache          map[string][]*github.Key
 	latestReleaseCache map[string]map[string]*github.RepositoryRelease
-	latestTagCache     map[string]map[string]*github.RepositoryTag
+	releasesCache      map[string]map[string][]*github.RepositoryRelease
+	tagsCache          map[string]map[string][]*github.RepositoryTag
 }
 
 func (c *Config) gitHubKeysTemplateFunc(user string) []*github.Key {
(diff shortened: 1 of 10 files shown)


halostatue:
I’m not sure how to fix the lint issue, because the functions are *not* duplicates, but they sure do rhyme really hard. They are not, however, compatible on return types and cache types.

twpayne:
This looks great!

> * do we actually want them?

Yes, I think these are worth adding.

> * are we good with these names, or do we want them to be `gitHubListReleases` and `gitHubListTags`

I think the current names (without `List`) are good.

> * do we want to allow the passing of pagination, or is this always the latest (up to) 30 releases?
> * is 30 the right number, or should we go smaller (10 is probably sufficient in most cases)
> * if we allow pagination parameters, how do we want to handle the caching, since it's now very different?

I think the approach in this PR is correct. Handling pagination is already fiddly, and would be especially difficult inside a template. We obviously don't want to list all releases.

The only change I would make would be to remove the `gitHubListOptions` variable and pass `nil` as the options instead, so the page sizes remain at the GitHub default (currently 30, but may change). The documentation would then say "returns the first page of GitHub releases" with a link to the GitHub API. This defers the default page size decision to GitHub so we behave exactly as GitHub does. I can't imagine a scenario where someone needs to go back more than 30 releases.

> * is the `output "curl"` or `output "gh" "api"` approach sufficient to where we do not need these?

For now, yes.

halostatue:
I may end up changing the `gitHubLatestTag` function so that it’s reusing and caching the results of the `gitHubTags` values, since there’s a lot of similarities there, and requesting 1 is as expensive as requesting (default page size).

bradenhilton:
`gitHubReleasesFunc` -> `gitHubReleasesTemplateFunc`

`gitHubTagsFunc` -> `gitHubTagsTemplateFunc`

halostatue:
> `gitHubReleasesFunc` -> `gitHubReleasesTemplateFunc`
> 
> `gitHubTagsFunc` -> `gitHubTagsTemplateFunc`

Yeah. I have changed those but not pushed it because of other things I was looking at.

twpayne:
> I’m not sure how to fix the lint issue, because the functions are _not_ duplicates, but they sure do rhyme really hard. They are not, however, compatible on return types and cache types.

You can use a `//nolint:` directive.

halostatue:
The functions should now be ready for empirical testing with `go run`, but I still need to write tests and documentation before this is ready to step out of draft mode.

There are two things of interest here:

- I changed the implementation of `gitHubLatestTagTemplateFunc` and `gitHubTagsTemplateFunc` to depend on a third function that shares state.
- I removed the old tag persistent state values. I’m not sure whether that would be considered safe or not.

internal/cmd/githubtemplatefuncs.go
```diff
@@ -25,23 +25,30 @@ type gitHubLatestReleaseState struct {
 	Release     *github.RepositoryRelease `json:"release"     yaml:"release"`
 }
 
-type gitHubLatestTagState struct {
-	RequestedAt time.Time             `json:"requestedAt" yaml:"requestedAt"`
-	Tag         *github.RepositoryTag `json:"tag"         yaml:"tag"`
+type gitHubReleasesState struct {
+	RequestedAt time.Time                   `json:"requestedAt" yaml:"requestedAt"`
+	Releases    []*github.RepositoryRelease `json:"release"     yaml:"release"`
```
twpayne, line 30: Nit: the tags should be `json:"releases" yaml:"releases"` (plural, not singular).

internal/cmd/githubtemplatefuncs.go
```diff
@@ -25,23 +25,30 @@ type gitHubLatestReleaseState struct {
...
+}
+
+type gitHubTagsState struct {
+	RequestedAt time.Time               `json:"requestedAt" yaml:"requestedAt"`
+	Tags        []*github.RepositoryTag `json:"tag"         yaml:"tag"`
```
twpayne, line 35: Nit: the tags should be `json:"tags" yaml:"tags"` (plural, not singular).

twpayne (commented):
Thank you, this looks excellent.

I did some tests with:

```console
$ go run . execute-template '{{ gitHubReleases "twpayne/chezmoi" | toJson }}'
$ go run . execute-template '{{ gitHubTags "twpayne/chezmoi" | toJson }}'
```

and also set `github.refreshPeriod` and inspected changes to the state with `chezmoi state dump`. Everything worked perfectly.

halostatue:
Ok. I’m on vacation for a couple more days and don’t expect to have time until Wednesday to finish this, but I’m having fun with it.

halostatue:
OK. I have the tests in place, but will be waiting to un-draft this when I squash everything having seen the successful run. Probably tomorrow evening.

twpayne (approved):
This is fantastic, thank you! Please merge :)

PR #3176, opened 2023-08-15
feat: Support nushell scripts on Windows

Sacro:
I've started writing scripts in [Nushell](https://www.nushell.sh/), then they can run on Windows, Linux, MacOS at once.

This just adds the mapping so Windows knows what to do with the file extension.

### DIFF
--- a/assets/chezmoi.io/docs/reference/target-types.md
+++ b/assets/chezmoi.io/docs/reference/target-types.md
@@ -109,6 +109,7 @@ The default script interpreters are:
 
 | Extension | Command      | Arguments |
 | --------- | ------------ | --------- |
+| `.nu`     | `nu`         | *none*    |
 | `.pl`     | `perl`       | *none*    |
 | `.py`     | `python3`    | *none*    |
 | `.ps1`    | `powershell` | `-NoLogo` |
--- a/internal/cmd/util_windows.go
+++ b/internal/cmd/util_windows.go
@@ -17,6 +17,9 @@ var defaultInterpreters = map[string]*chezmoi.Interpreter{
 	"cmd": {},
 	"com": {},
 	"exe": {},
+	"nu": {
+		Command: "nu",
+	},
 	"pl": {
 		Command: "perl",
 	},


assets/chezmoi.io/docs/reference/target-types.md
```diff
@@ -108,7 +108,8 @@ extensions. Other extensions require an interpreter, which must be in your
 The default script interpreters are:
 
 | Extension | Command      | Arguments |
-| --------- | ------------ | --------- |
+| --------- | ------------ |-----------|
+| `.nu`     | `nushell`    | *none*    |
```
twpayne, line 112: Is the command `nu` or `nushell`? In this line the command is described as `nushell` but in `util_windows.go` the default is set to `nu`.

internal/cmd/util_windows.go
```diff
@@ -17,6 +17,9 @@ var defaultInterpreters = map[string]*chezmoi.Interpreter{
 	"cmd": {},
 	"com": {},
 	"exe": {},
+	"nu": {
+		Command: "nu",
```
twpayne, line 21: Should this be `nu` or `nushell`?

assets/chezmoi.io/docs/reference/target-types.md
```diff
@@ -108,7 +108,8 @@ extensions. Other extensions require an interpreter, which must be in your
 The default script interpreters are:
 
 | Extension | Command      | Arguments |
-| --------- | ------------ | --------- |
+| --------- | ------------ |-----------|
```
twpayne, line 111: A couple of spaces got replaced with hyphens in the last column here.

twpayne (changes requested):
Looks reasonable, but is inconsistent. See comments.

assets/chezmoi.io/docs/reference/target-types.md
Sacro: The command is `nu`, I was thrown by `powershell`, which changed to `pwsh` in later versions.

internal/cmd/util_windows.go
Sacro: It should be `nu`, I've tested it locally and it runs the scripts fine.

twpayne (approved):
Wonderful, thank you!

PR #3184, opened 2023-08-18
docs: Linuxbrew -> Homebrew

osalbahr:
[Homebrew on Linux was formerly referred to as Linuxbrew](https://docs.brew.sh/Homebrew-on-Linux). The two projects got merged.

### DIFF
--- a/assets/chezmoi.io/docs/install.md.tmpl
+++ b/assets/chezmoi.io/docs/install.md.tmpl
@@ -107,7 +107,7 @@ chezmoi is available in many cross-platform package managers:
     asdf plugin add chezmoi && asdf install chezmoi {{ $version }}
     ```
 
-=== "Linuxbrew"
+=== "Homebrew"
 
     ```sh
     brew install chezmoi


halostatue:
The change here looks fine, but it looks a little weird to have two headings `Homebrew`. I’ll leave it up to @twpayne to make the decision, but I think that we should do one of two things:

- remove `Homebrew` from the `macOS` section
- rename this section from `Homebrew` to `Homebrew on Linux`

My personal preference is for the latter, but it’s a weak preference.

osalbahr:
I lean more towards keeping it as just Homebrew for brevity reasons and being consistent with having two Nix headings. I’m not sure about, say, “Nix on non-NixOS” either.

I’d like to note that I can see my PR comment being a little misleading. To clarify, “Homebrew on Linux” isn’t the name of a separate package manager that works across Linux systems. It is the same package manager, Homebrew, works on both macOS and Linux. Rather, that phrase is just the heading of a technical documentation webpage.

bradenhilton:
I'm of the opinion that `Homebrew` should be moved from the `macOS` section to the cross-platform section.

I'd personally also swap the sections so the cross-platform section comes before the platform-specific section, but I feel that may be more contentious.

halostatue:
I just looked at the way it shows up on the web page, and this is a special `mkdocs` format which creates tabs and nested tabs. I think that this is fine as it is for the current structure, because it corrects the project name from `Linuxbrew` to `Homebrew`, and it does nothing else.

halostatue (approved):

PR #3199, opened 2023-08-28
feat: Add age command

twpayne:
Refs #3198.

@arrrgi is this what you had in mind?

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/age.md
+++ b/assets/chezmoi.io/docs/reference/commands/age.md
@@ -1,0 +1,18 @@
+# `age`
+
+Interact with age's passphrase-based encryption.
+
+!!! hint
+
+    To get a full list of subcommands run:
+
+    ```console
+    $ chezmoi age help
+    ```
+
+!!! example
+
+    ```console
+    $ chezmoi age encrypt --passphrase plaintext.txt > ciphertext.txt
+    $ chezmoi age decrypt --passphrase ciphertext.txt > decrypted-ciphertext.txt
+    ```
--- a/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
+++ b/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/encryption.md
@@ -44,7 +44,7 @@ $ cat > run_once_before_decrypt-private-key.sh.tmpl <<EOF
 #!/bin/sh
 
 if [ ! -f "${HOME}/key.txt" ]; then
-    age --decrypt --output "${HOME}/key.txt" "{{ .chezmoi.sourceDir }}/key.txt.age"
+    chezmoi age decrypt --output "${HOME}/key.txt" --passphrase "{{ .chezmoi.sourceDir }}/key.txt.age"
     chmod 600 "${HOME}/key.txt"
 fi
 EOF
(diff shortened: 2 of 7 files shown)


arrrgi:
Can you see a possibility of a race condition occurring by calling chezmoi within chezmoi (within chezmoi ad-infinitum)? 

Maybe there should be a check that if the parent process is chezmoi, only a subset of commands is available to prevent the scenario discussed in #3198 

> Running chezmoi from chezmoi is generally OK as long as the nested chezmoi does not modify the source or target states. Specifically, chezmoi decrypt should be safe to run from chezmoi.

Otherwise, LGTM 👍

arrrgi:
One other question, does this change or require a change to the `--use-builtin-age` switch? [(Reference)](https://www.chezmoi.io/reference/command-line-flags/global/#-use-builtin-age-value)

halostatue:
> One other question, does this change or require a change to the `--use-builtin-age` switch? [(Reference)](https://www.chezmoi.io/reference/command-line-flags/global/#-use-builtin-age-value)

That’s a good catch as `--use-builtin-age` for `chezmoi decrypt` and similar operations now supports passphrase entry, it looks like.

arrrgi:
How do we configure this in .chezmoi.<format>.tmpl under the `encryption` and `age` config keys? 

I'm making a fairly big assumption as I don't properly understand the code, that the default **chezmoi** configuration of `useBuiltinAge: auto` would handle this situation automatically. I'm not sure if it does right now so this is an additional consideration.

halostatue:
> How do we configure this in .chezmoi..tmpl under the `encryption` and `age` config keys?
> 
> I'm making a fairly big assumption as I don't properly understand the code, that the default **chezmoi** configuration of `useBuiltinAge: auto` would handle this situation automatically. I'm not sure if it does right now so this is an additional consideration.

`useBuiltinAge: auto` only works if you do not have `age` installed. But there are two parts to this PR:

1. Adding passphrase entry support (that’s `lazyscryptidentity`). As far as I can tell, this will work *automatically* with `useBuiltinAge: true`.
2. Adding a `chezmoi age` subcommand with a reasonable (small) subset of the `age` command functionality, making it so that you can call `chezmoi age` in a `run_before_` script.

So the only change to the docs mentioned would be to remove the mention that there is no support for passphrase entry. For normal `age` operations where `useBuiltinAge: true` or `age` is not installed…there is no need to tell `chezmoi` to use `chezmoi age`, because it uses the `age` *code* from within the operations requiring decryption.

twpayne:
> But there are two parts to this PR:
> 
> 1. Adding passphrase entry support (that’s `lazyscryptidentity`). As far as I can tell, this will work _automatically_ with `useBuiltinAge: true`.
> 2. Adding a `chezmoi age` subcommand with a reasonable (small) subset of the `age` command functionality, making it so that you can call `chezmoi age` in a `run_before_` script.

This isn't quite right.

This PR adds the `chezmoi age encrypt` and `chezmoi age decrypt` commands that only work with passphrase-based encryption. It then updates the user guide to describe how to use these commands to only have to enter your passphrase when you run `chezmoi init` on a new machine for the first time.

Fundamentally, the `chezmoi age` commands are separate from chezmoi's builtin age encryption (although, of course, a lot of the code is shared). This separation is deliberate.

Passphrases are, by nature, interactive, so you don't want to use them to encrypt files in chezmoi, so chezmoi's builtin age encryption doesn't support them.

However, passphrases are very useful for one-off encryption/decryption tasks, such as decrypting your private key on a new machine, so the `chezmoi age` commands support exactly only passphrase-based encryption for exactly this purpose.

halostatue (approved):

PR #3201, opened 2023-08-29
docs: Minor expansion of application order docs

halostatue:
The discussion in #3198 suggests some confusion about the order of application. This adds a little more context.


arrrgi (approved):
LGTM and definitely clears up the use case we've discussed under #3198

twpayne (approved):
Thank you!

Issue #3220, opened 2023-09-05
A tar external with AppleDouble files extracts them instead of applying the contents as attributes (or ignoring them)


halostatue:
## Describe the bug

This was originally found as an issue where several `txtar` tests fail because of the presence of [AppleDouble](https://en.wikipedia.org/wiki/AppleSingle_and_AppleDouble_formats) files (that is, an archive with `file` and `._file` present), resulting in extra output:

```
    --- FAIL: TestScript/external (0.32s)
        testscript.go:534: # test that chezmoi diff includes external files by default (0.051s)
            # test that chezmoi diff --exclude=externals excludes diffs from external files (0.027s)
            # test that chezmoi reads external files from .chezmoiexternal.toml (0.059s)
            # test that chezmoi reads executable external files from .chezmoiexternal.toml (0.072s)
            # test that chezmoi managed --include=externals lists external targets (0.047s)
            > exec chezmoi managed --include=externals
            [stdout]
            .dir
            .dir/._dir
            .dir/dir
            .dir/dir/._file
            .dir/dir/._symlink
            .dir/dir/file
            .dir/dir/symlink
            > cmp stdout golden/managed
            diff stdout golden/managed
            --- stdout
            +++ golden/managed
            @@ -1,7 +1,4 @@
             .dir
            -.dir/._dir
             .dir/dir
            -.dir/dir/._file
            -.dir/dir/._symlink
             .dir/dir/file
             .dir/dir/symlink

            FAIL: testdata/scripts/external.txtar:30: stdout and golden/managed differ
```

I have been able to reproduce this using the attached tarball: [test.tar.gz](https://github.com/twpayne/chezmoi/files/12518531/test.tar.gz)

## To reproduce

1. Put the attached tarball in `~/tmp`. It has been deliberately created with one extended attribute, but has additional attributes automatically added by macOS Ventura (`com.apple.provenance`).
2. Serve `~/tmp` as a website: `cd ~/tmp && python3 -m http.server 8080`
3. Add this tarball as an external:
   ```yaml
   test:
     type: archive
     url: http://localhost:8080/test.tar.gz
   ```
4. Run `chezmoi managed --include=externals`:

   ```console
   $ chezmoi managed --include=externals
   test
   test/._test
   test/test
   test/test/._file
   test/test/file
   ```

This *only* happens with tarballs and not with zipfiles according to my testing.

## Expected behavior

[...]

PR #3221, opened 2023-09-05
fix: Prevent AppleDouble file creation in .txtar

halostatue:
Related to #3220 and fixes it for tests run on macOS (but does not fix the underlying issue).

Any time we create an archive with `exec tar c…` in a `txtar` test file, particularly where the golden contents will be compared, we need to have a separate command for `darwin` platforms:

```txtar
[!darwin] tar cfz archive.tar.gz archive
[darwin] tar cfz archive.tar.gz --no-mac-metadata archive
```

Otherwise, the actual output will be incorrect compared to the golden version on macOS because of the presence of
[AppleDouble](https://en.wikipedia.org/wiki/AppleSingle_and_AppleDouble_formats) files. As a developer on macOS, this is helpful because it will give me an improved test cycle before submitting for CI.

### DIFF
--- a/internal/cmd/testdata/scripts/external.txtar
+++ b/internal/cmd/testdata/scripts/external.txtar
@@ -1,5 +1,6 @@
 symlink archive/dir/symlink -> file
-exec tar czf www/archive.tar.gz archive
+[!darwin] exec tar czf www/archive.tar.gz archive
+[darwin] exec tar czf www/archive.tar.gz --no-mac-metadata archive
 
 httpd www
 
--- a/internal/cmd/testdata/scripts/externalarchiveinclude.txtar
+++ b/internal/cmd/testdata/scripts/externalarchiveinclude.txtar
@@ -1,7 +1,8 @@
 symlink archive/symlink1 -> file1
 symlink archive/symlink2 -> file2
 mkdir www
-exec tar czf www/archive.tar.gz archive
+[!darwin] exec tar czf www/archive.tar.gz archive
+[darwin] exec tar czf www/archive.tar.gz --no-mac-metadata archive
 
 httpd www
 
--- a/internal/cmd/testdata/scripts/externalencrypted.txtar
+++ b/internal/cmd/testdata/scripts/externalencrypted.txtar
@@ -4,7 +4,8 @@
 mkgpgconfig
 
 # use chezmoi's encryption to encrypt a file and an archive
-exec tar czf $HOME/archive.tar.gz archive
+[!darwin] exec tar czf $HOME/archive.tar.gz archive
+[darwin] exec tar czf $HOME/archive.tar.gz --no-mac-metadata archive
 exec chezmoi add --encrypt $HOME${/}.file $HOME${/}archive.tar.gz
 mkdir www
 cp $CHEZMOISOURCEDIR/encrypted_dot_file.asc www/.file.asc
--- a/internal/cmd/testdata/scripts/import.txtar
+++ b/internal/cmd/testdata/scripts/import.txtar
@@ -1,6 +1,7 @@
 mkhomedir
 symlink archive/.dir/.symlink -> .file
-exec tar czf archive.tar.gz archive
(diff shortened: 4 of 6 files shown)


twpayne:
> Otherwise, the actual output will be incorrect compared to the golden version on macOS because of the presence of
> [AppleDouble](https://en.wikipedia.org/wiki/AppleSingle_and_AppleDouble_formats) files.

Is this really necessary? The tests already pass on macOS without the `--no-mac-metadata` flag. We have CI that runs on macOS, and I regularly develop on macOS (Ventura) and have not seen any of the problems described.

halostatue:
> > Otherwise, the actual output will be incorrect compared to the golden version on macOS because of the presence of
> > [AppleDouble](https://en.wikipedia.org/wiki/AppleSingle_and_AppleDouble_formats) files.
> 
> Is this really necessary? The tests already pass on macOS without the `--no-mac-metadata` flag. We have CI that runs on macOS, and I regularly develop on macOS (Ventura) and have not seen any of the problems described.

For some reason, almost every file I have in my `chezmoi` source tree has a `com.apple.provenance` attribute *that cannot be removed* (this appears to be [related to newest Gatekeeper](https://eclecticlight.co/2023/03/16/what-is-macos-ventura-doing-tracking-provenance/)) and may be due to my use of iTerm2 as my preferred terminal. It could also be my use of Homebrew `fish`. It could also be related to Apple Silicon. I’m not sure.

Every time I run `make test`, I get:

```
    --- FAIL: TestScript/external (0.32s)
        testscript.go:534: # test that chezmoi diff includes external files by default (0.051s)
            # test that chezmoi diff --exclude=externals excludes diffs from external files (0.027s)
            # test that chezmoi reads external files from .chezmoiexternal.toml (0.059s)
            # test that chezmoi reads executable external files from .chezmoiexternal.toml (0.072s)
            # test that chezmoi managed --include=externals lists external targets (0.047s)
            > exec chezmoi managed --include=externals
            [stdout]
            .dir
            .dir/._dir
            .dir/dir
            .dir/dir/._file
            .dir/dir/._symlink
            .dir/dir/file
            .dir/dir/symlink
            > cmp stdout golden/managed
            diff stdout golden/managed
            --- stdout
            +++ golden/managed
            @@ -1,7 +1,4 @@
             .dir
            -.dir/._dir
             .dir/dir
            -.dir/dir/._file
            -.dir/dir/._symlink
             .dir/dir/file
             .dir/dir/symlink

            FAIL: testdata/scripts/external.txtar:30: stdout and golden/managed differ
```

I also get failures on 3–4 other `txtar` files, which means that I cannot trust the output of `make test` on my macOS system.

I can try cloning the chezmoi tree again, but even the test files that I created for #3220 demonstrate this issue:

```console
$ xattr -r ~/tmp/test
/Users/austin/tmp/test/file: com.apple.provenance
/Users/austin/tmp/test/file: io.chezmoi.test
/Users/austin/tmp/test: com.apple.provenance
```

The presence of `com.apple.provenance` seems *unlikely* to show up on the macOS runners, but the presence of any single xattr is sufficient to cause the `txtar` tests to break.

twpayne:
Thank you for the further investigation!

Firstly, I'd really like to understand why these attributes are getting added on your machine but not on mine. I also use iTerm2 on Apple Silicon (MacBook Air M1, 2020, Ventura 13.5.1). I'll see if I can find another Mac to test on.

Secondly, it's interesting to see that libarchive filters out such files. It's reassuring to see that the filtering is both on the filename and the first eight bytes of its content, which means that if chezmoi does filter out such files, it is much less likely to filter out false positives.

Thirdly, if chezmoi does end up automatically filtering out AppleDouble files from tar archives, then this PR will not be needed, as the tests should pass without the need to add the `--no-mac-metadata` flag when creating the archives.

But, I'd really like to understand where the difference in behavior is coming from first.

halostatue:
I’ve set this to be closed from #3262 on merge as it will no longer be necessary.

PR #3222, opened 2023-09-05
docs: Improve developer experience under fish

halostatue:
The added documentation serves as the rationale for this documentation change:

If you use `fish` as your primary shell, you may get warnings from Fish during the `.txtar` tests:

```
error: can not save history
warning-path: Unable to locate data directory derived from $HOME: '/home/user/.local/share/fish'.
warning-path: The error was 'Operation not supported'.
warning-path: Please set $HOME to a directory where you have write access.
```

These can be avoided with by running tests with `SHELL=bash` or `SHELL=zsh`:

```console
$ SHELL=bash make test
$ SHELL=zsh make smoketest
$ SHELL=bash go test ./...
```

---

As a note, I tried to make `make test` re-execute when under `fish`, but it requires some GNU make magic that may not work on all systems:

```Makefile
.PHONY: test
test:
ifeq ($(shell echo $$SHELL | grep -c fish$$),1)
	SHELL=$(SHELL) $(MAKE) test
else
	${GO} test -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o022" ./...
	${GO} test -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o002" ./...
endif
```

This can be added to this PR or made as a separate PR, if desired.

### DIFF
--- a/assets/chezmoi.io/docs/developer/developing-locally.md
+++ b/assets/chezmoi.io/docs/developer/developing-locally.md
@@ -26,7 +26,6 @@ $ go build
     package github.com/twpayne/chezmoi/v2: build constraints exclude all Go files in /home/twp/src/github.com/twpayne/chezmoi
     ```
 
-
 Run all tests:
 
 ```console
@@ -49,3 +48,23 @@ Run a set of smoketests, including cross-compilation, tests, and linting:
 ```console
 $ make smoketest
 ```
+
+!!! hint
+
+    If you use `fish` as your primary shell, you may get warnings from Fish
+    during tests:
+
+    ```
+    error: can not save history
+    warning-path: Unable to locate data directory derived from $HOME: '/home/user/.local/share/fish'.
+    warning-path: The error was 'Operation not supported'.
+    warning-path: Please set $HOME to a directory where you have write access.
+    ```
+
+    These can be avoided with by running tests with `SHELL=bash` or `SHELL=zsh`:
+
+    ```console
+    $ SHELL=bash make test
+    $ SHELL=zsh make smoketest
+    $ SHELL=bash go test ./...
+    ```


twpayne:
Hmm, this feels like a bug in fish to me:

1. `$HOME` is not always writable, for example when you use `su` or `sudo` to change user.

   You definitely should not write files to `$HOME` if you're not the owner of the directory. Otherwise, using `sudo` or `su` will result in root-owned files in your home directory, which cannot be modified by you and are a pain to remove.

2.  Running a shell non-interactively should not generate user-visible messages.

Further investigation indicates that this is the same issue as https://github.com/fish-shell/fish-shell/issues/6630, fixed with https://github.com/fish-shell/fish-shell/commit/ee943a0821cd1e0846cf6ea52d2259374ebe02c3. It looks there's some sort of `warning-path` option that can be set instead.

halostatue:
> Further investigation indicates that this is the same issue as [fish-shell/fish-shell#6630](https://github.com/fish-shell/fish-shell/issues/6630), fixed with [fish-shell/fish-shell@ee943a0](https://github.com/fish-shell/fish-shell/commit/ee943a0821cd1e0846cf6ea52d2259374ebe02c3). It looks there's some sort of `warning-path` option that can be set instead.

As far as I can tell, that is only settable from `fish --debug=-warning-path` (that is, from the initial shell execution). I can’t find any evidence of the setting from inside of fish itself or to do so "permanently" (not necessarily what I want to do in any case). Since `go test…` is calling `$SHELL` automatically with different `$HOME` values, there’s no opportunity to disable these warnings.

A comment was made in [fish/fish-shell#4573](https://github.com/fish-shell/fish-shell/issues/4573#issuecomment-348925470) that fish is just more verbose on this matter than other shells. It would be nice to not have those warnings show up during the running of `make test` while developing for chezmoi, since it is noise. Although I figured out what would be necessary to make this work *safely* in the Makefile, I figured the documentation update is the least intrusive.

twpayne:
Thanks, that makes sense.

twpayne (approved):

PR #3230, opened 2023-09-12
chore: Add test case for issue #3229

twpayne:
Refs #3229.


twpayne:
No longer needed as #3229 did not get a response from the reporter and there doesn't seem to be an issue with chezmoi here.

PR #3235, opened 2023-09-13
feat: Add plugin support

twpayne:
cc @felipecrs who previously requested plugin support (sorry I can't find the link)

### DIFF
--- a/assets/chezmoi.io/docs/reference/plugins.md
+++ b/assets/chezmoi.io/docs/reference/plugins.md
@@ -1,0 +1,8 @@
+# Plugins
+
+chezmoi supports plugins, similar to git.
+
+If you run `chezmoi command` where *command* is not a builtin chezmoi command
+then chezmoi will look for a binary called `chezmoi-command` in your `$PATH`. If
+such a binary is found then chezmoi will execute it. Otherwise, chezmoi will
+report an unknown command error.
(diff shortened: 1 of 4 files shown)


felipecrs:
That's super cool, can't wait to see what people will build around it.

For example, I could transform my `rootmoi` thing into a `chezmoi-root` plugin.

halostatue (approved):
Looks great.

halostatue:
Two things that would be useful.

1. Expose the `$CHEZMOI_` environment variables.
2. Figure out a way to expose (possibly through a named pipe, FIFO, or something similar) the chezmoi context (`.data`).

The first should be pretty easy. The second will take some thinking and design, so should not be part of this PR.

felipecrs:
The second can be passed through an environment variable like CHEZMOI_DATA='{"this": "is-json"}'

halostatue:
> The second can be passed through an environment variable like CHEZMOI_DATA='{"this":"is-json"}'

Maybe, if the data is small (and I suspect it would be for most folks). I believe that some platforms (Windows?) have a maximum size on environment variable values. It’s ridiculously big (4k or 8k), but it is still limited.

halostatue:
> 1. Expose the `$CHEZMOI_` environment variables.

Looking at the implementation again, I’m not sure that it would be easy. Both of these options should probably be follow-ups. If `.data` or other output is required, then `chezmoi execute-template` or `chezmoi data` can be called.

I think that the only thing we should *probably* do is set something like setting `CHEZMOI=1` and `CHEZMOI_PLUGIN=1`.

twpayne:
Thanks for the reviews and comments. Let's :shipit: as-is.

PR #3241, opened 2023-09-15
chore: Add test and tweak error message in cd command

twpayne:


halostatue (approved):
I didn't even think of checking for tests, and I had gone back and forth on adding the target to the output—what you have is more succinct than the versions I had come up with.

PR #3252, opened 2023-09-18
chore: use json.MarshalIndent in toPrettyJson

bradenhilton:
Recently ran across [`MarshalIndent`](https://pkg.go.dev/encoding/json#MarshalIndent) in the docs.

The final newline is needed for backwards compatibility.


bradenhilton:
I think for now I will satisfy the linter rather than use another nolint directive or similar. I don't have a strong opinion either way though.

halostatue (approved):
Nice.

twpayne (approved):

PR #3258, opened 2023-09-21
fix: Only invoke diff pager if command modifies filesystem

### LINKED ISSUE #3257: chezmoi 2.40's always outputs a blank page when using `PAGER`
## Describe the bug

Sorry, I could not think of a better title ATM. Feel free to retitle it. And I believe a video is worth more than words:

https://github.com/twpayne/chezmoi/assets/29582865/5e63706e-72d7-4302-8195-a529984de071

This is happening to pretty much all chezmoi commands, like update, apply, init, status. Note that I have `PAGER=less` in my rc files.

PS: I'm using [tea](https://tea.xyz/) to test different versions of chezmoi more easily.

## To reproduce

Supposedly it should be as simple as:

```
PAGER=less chezmoi status
```

## Expected behavior

The output to be given just like before.

No extra logs are shown.

## Output of `chezmoi doctor`

I can't see the output.

image

## Additional context

- Probably related to https://github.com/twpayne/chezmoi/pull/3243

twpayne:
Fixes #3257.

cc @felipecrs. Thank you for your help on this. This still needs a bit of testing and thinking about edge cases, so this PR is draft for now. If you could test this PR, that would be awesome.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -1992,7 +1992,9 @@ func (c *Config) persistentPreRunRootE(cmd *cobra.Command, args []string) error 
 		c.sourceSystem = chezmoi.NewDryRunSystem(c.sourceSystem)
 		c.destSystem = chezmoi.NewDryRunSystem(c.destSystem)
 	}
-	if c.Verbose || annotations.hasTag(outputsDiff) {
+	if annotations.hasTag(outputsDiff) ||
+		c.Verbose &&
+			(annotations.hasTag(modifiesDestinationDirectory) || annotations.hasTag(modifiesSourceDirectory)) {
 		// If the user has configured a diff pager, then start it as a process.
 		// Otherwise, write the diff output directly to stdout.
 		var writer io.Writer
--- a/internal/cmd/statuscmd.go
+++ b/internal/cmd/statuscmd.go
@@ -28,7 +28,6 @@ func (c *Config) newStatusCmd() *cobra.Command {
 		RunE:              c.runStatusCmd,
 		Annotations: newAnnotations(
 			dryRun,
-			modifiesDestinationDirectory,
 			persistentStateModeReadMockWrite,
 			requiresSourceDirectory,
 		),
--- a/internal/cmd/testdata/scripts/issue3257.txtar
+++ b/internal/cmd/testdata/scripts/issue3257.txtar
@@ -1,0 +1,26 @@
+[windows] skip 'UNIX only'
+
+chmod 755 bin/custom-pager
+
+# test that chezmoi add invokes the pager when verbose is set
+exec chezmoi add $HOME${/}.file
+stdout custom-pager
+
+# test chat chezmoi chattr invokes the pager when verbose is set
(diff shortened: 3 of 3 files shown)


internal/cmd/config.go
```diff
@@ -1986,7 +1986,9 @@ func (c *Config) persistentPreRunRootE(cmd *cobra.Command, args []string) error
 		c.sourceSystem = chezmoi.NewDryRunSystem(c.sourceSystem)
 		c.destSystem = chezmoi.NewDryRunSystem(c.destSystem)
 	}
-	if c.Verbose || annotations.hasTag(outputsDiff) {
+	if annotations.hasTag(outputsDiff) ||
+		c.Verbose &&
+			(annotations.hasTag(modifiesDestinationDirectory) || annotations.hasTag(modifiesSourceDirectory)) {
```
felipecrs, line 1991: You probably thought about this already, but let me suggest in case you didn't: have you thought about adding the `outputsDiff` tag to the command when applicable with Verbose? This way you could keep the if as `if annotations.hasTag(outputsDiff) {` only.

twpayne: The annotations are properties of the command, which are set before the flags are parsed.

felipecrs:
Ok, I tested it, and it fixes the bug. Although, I need to tell you, I really disliked calling `PAGER` automatically on `apply --verbose` or `update --verbose`. It's really weird for example having to answer questions within the pager itself. Not to mention it clears my whole screen for several seconds until something actually pops up (this is a side effect of my dotfiles repository being so heavy, but calling PAGER makes it appear a little worse).

https://github.com/twpayne/chezmoi/assets/29582865/ecab3ae2-dd6d-4103-aa17-6d8dad595fbb

If you really want to push this feature, maybe you should make it opt-in through a config and wait to make it default on chezmoi v3. Maybe something like:

```yaml
# .chezmoi.yaml.tmpl

apply:
  usePagerOnVerbose: true
```

twpayne:
Um, isn't this an issue with the chosen pager, not chezmoi? The pager's job is to present long output one page at a time. This should be lightweight. If your pager takes several seconds to display the output then could the issue be within the chosen pager?

felipecrs:
Oh, no. That's not the reason why it's taking so long to show something. That's because chezmoi opens the pager very early but only outputs things to it after some seconds (because my dotfiles are heavy).

The same thing happens when not using a pager:


https://github.com/twpayne/chezmoi/assets/29582865/b06cadcd-1967-44e2-9881-e18f5b359d05

twpayne:
@felipecrs I think this PR fixes a genuine problem so I've merged it, but the issue of chezmoi's behavior when you have heavy dotfiles is a different one. Please could you open a new issue with the heavy dotfiles problem.

felipecrs:
You are right. And it is ok, I have [disabled pager for chezmoi](https://github.com/felipecrs/dotfiles/commit/58a11da640e79c9c87eb89616f18e22ceb9dd430) ever since, and I can live with it. Thank you.

PR #3259, opened 2023-09-21
docs: Fix cross-function link references

halostatue:
No files specified


twpayne (approved):
Thank you very much for fixing these :)

Issue #3256, opened 2023-09-20
Look for one of several executables in `$PATH`


halostatue:
## Is your feature request related to a problem? Please describe.

I found myself writing the following template code, which is somewhat unreadable:

```
{{ $program := or (lookPath "eza") (lookPath "exa") | base }}
```

This returns the *first* instance of `eza` or `exa` and takes the basename.

## Describe the solution you'd like

I think that this could be easier with a small complexity increase to `internal/chezmoi/lookpath.go`. With a little more, we can extend `lookPath` to automatically take the `basename`. that is, the above code now becomes:

```
{{ $program := lookPath (list "era" "exa") | base }}
```

I have a ~70% code change ready for this, but wanted thoughts before I went further. This could *also* be done as a separate template function `lookPathList`, but the any/conversion approach I’ve taken results in better caching, IMO.

## Describe alternatives you've considered

The `or` code above, which gets *really* ugly if you have one of *several* possible options. See https://github.com/halostatue/dotfiles/blob/main/home/private_dot_config/git/attributes.tmpl#L68-L70 and https://github.com/halostatue/dotfiles/blob/main/home/private_dot_config/git/diff-merge.tmpl#L108-L120.

PR #3260, opened 2023-09-21
feat: find[One]Executable in user-supplied paths

halostatue:
Implement `findExecutable` with strong caching. The main change from
#3162 is that it now also accepts either a `file` or `file-list` for
searching, so that one of several different options can be searched
simultaneously for roughly equivalent implementations. See #3256 for the
inspiration for this particular change.

We _can_ switch this to two template functions, but I think that the
core implementation would best remain the way that it is.

Resolves: #3141
Closes: #3162
Co-authored-by: Arran Ubels <arran4@gmail.com>


internal/chezmoi/findexecutable_darwin_test.go
```diff
@@ -0,0 +1,90 @@
+//go:build darwin
```
twpayne, line 1: I _think_ this line is not needed as presence of `darwin` in the filename means that Go will only build it on macOS.

internal/chezmoi/findexecutable_darwin_test.go
```diff
@@ -0,0 +1,90 @@
...
+		name := fmt.Sprintf("FindExecutable %v in %#v as %v", test.file, test.paths, test.expected)
+		t.Run(name, func(t *testing.T) {
+			actual, err := FindExecutable(test.file, test.paths)
+			if err != nil {
+				t.Errorf("FindExecutable() error = %v, expected = nil", err)
+			}
+
+			if actual != test.expected {
+				t.Errorf("FindExecutable() actual = %v, expected = %v", actual, test.expected)
+			}
```
twpayne, line 43: Instead of writing these assertions out in Go, the rest of the code uses github.com/alecthomas/assert/v2 which means this block can be written as:

```go
actual, err := FindExecutable(test.file, test.paths)
assert.NoError(t, err)
assert.Equal(t, test.expected, actual)
```

This same comment applies to the other tests in this PR.

twpayne (approved):
LGTM, thank you.

Feel free to ignore the two minor nits if you want to merge this now.

internal/chezmoi/findexecutable_darwin_test.go
halostatue: I’ll give it a try as its own commit to see if there are build issues. That makes it easy to revert if there are.

internal/chezmoi/findexecutable_darwin_test.go
halostatue: Done. Let’s see what the automated build says for both of these changes.

halostatue:
One question before I merge this, @twpayne. Do you prefer `findExecutable *file | file-list* *path-list*` (one template function) or `findExecutable *file* *path-list*` and `findOneExecutable *file-list* *path-list*` (multiple template functions). This will affect the basic implementation I have for resolving #3256.

I think that the underlying implementation (`needle, haystack any` -> `needle any, haystack []string`) is the right one, but I can be explicit about `needle []string, haystack []string` in the underlying implementation and do the appropriate conversion in the template function(s).

twpayne:
I prefer multiple template functions. This gives stronger type checking, which means bugs are found earlier.

twpayne (approved):

PR #3262, opened 2023-09-23
feat: Ignore AppleDouble files in externals

halostatue:
This is somewhere between a fix and a feature because it is a potential bug if an archive contains AppleDouble data, as such data would automatically be written out as parallel `._FILE` files and not reintegrated. I've marked it as a feature because it requires new configuration options.

With this change, we are still not reintegrating the data, but we are explicitly _ignoring_ the data unless `EXTERNAL.archive.extractAppleDoubleFiles = true`. There are tools such as https://github.com/SiviourBla/appledouble that can theoretically reintegrate this using `copyfile(3)` on macOS.

This checks that the extracted file name will be `._*` before trying to determine whether the file is an AppleDouble file.

Resolves: #3220
Closes: #3221

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -55,6 +55,10 @@ var (
 	)
 	templateDirectiveRx             = regexp.MustCompile(`(?m)^.*?chezmoi:template:(.*)$(?:\r?\n)?`)
 	templateDirectiveKeyValuePairRx = regexp.MustCompile(`\s*(\S+)=("(?:[^"]|\\")*"|\S+)`)
+
+	// AppleDouble constants.
+	appleDoubleMagicCode = []byte{0x00, 0x05, 0x16, 0x07}
+	appleDoubleVersion   = []byte{0x00, 0x02, 0x00, 0x00}
 )
 
 // An External is an external source.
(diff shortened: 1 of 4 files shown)


halostatue:
This uses the logic at https://github.com/libarchive/libarchive/blob/67a20ce8f0ba22f06fb74119b4459ac69ade643b/libarchive/archive_write_disk_posix.c#L4453-L4458 and verifies the filename begins with `._` before that.

We have the test file in #3220 that would be ideal to test with, but I don't know how or if we can include a `.tar.gz` in the repo for testing (it's 546 bytes, so not a big file), so I made the `txtar` test work only on macOS, where you can *create* the AppleDouble files. If we can include the test `.tar.gz`, then we can test `chezmoi managed --include=externals` on *any* platform for extracting the AppleDouble files.

I have made it so that the default behaviour is to *not* extract them as they are usually accidental unless you are guaranteed going between macOS and macOS and really need the extended attributes &c.…but I don’t see that being a big deal for chezmoi as an additional step would be required in any case.

internal/chezmoi/sourcestate.go
```diff
@@ -2338,6 +2345,17 @@ func (s *SourceState) readExternalArchive(
...
+			if strings.HasPrefix(path.Base(name), "._") && !external.Archive.ExtractAppleDoubleFiles {
+				if bytes.Equal(appleDoubleMagicCode, contents[0:4]) {
+					return nil
+				}
+
+				if bytes.Equal(appleDoubleVersion, contents[4:8]) {
+					return nil
+				}
+			}
+
```
twpayne, line 2358: These lines are identical to the lines in `SourceState.readExternalArchiveFile`, so I'd factor them out into a `isAppleDoubleFile(name string, contents []byte) bool` function.

twpayne (commented):
LGTM, one nit. Please merge when the function is factored out.

twpayne (approved):

internal/chezmoi/sourcestate.go
```diff
@@ -2815,3 +2832,11 @@ func allEquivalentDirs(sourceStateEntries []SourceStateEntry) bool {
 	}
 	return true
 }
+
+// isAppleDoubleFile returns true if the file looks like and has the
+// expected signature of an AppleDouble file.
+func isAppleDoubleFile(name string, contents []byte) bool {
+	return strings.HasPrefix(path.Base(name), "._") &&
+		bytes.Equal(appleDoubleMagicCode, contents[0:4]) &&
+		bytes.Equal(appleDoubleVersion, contents[4:8])
```
twpayne, line 2841: This needs a `len(contents) >= 8` check to avoid panicking. Sorry I just realized this.

halostatue:
Pushed up that last change. You can merge if you’re up later, or I’ll get it in the morning my time.

PR #3264, opened 2023-09-26
docs: Add examples for string init functions

N3WK1D:
I struggled for quite a while to setup this function in my config template. For someone like me with no knowledge of Go, this would have been super helpful to have.


halostatue (approved):

halostatue:
This looks very good from my perspective.

twpayne:
Thank you!

PR #3270, opened 2023-10-01
feat: Add Path Separator and Path List Separator to provided template values

arran4:
This change exports a provided variable from Go which practically makes it free in terms of implementation costs.

This functionality would be implemented by hand by most people without variance when it is provided to Chezmoi by Go.

References: https://github.com/twpayne/chezmoi/pull/3162

There seems to be reference to this here too: 
https://github.com/twpayne/chezmoi/pull/3261#issuecomment-1734136510

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -249,27 +249,29 @@ type Config struct {
 }
 
 type templateData struct {
-	arch           string
-	args           []string
-	cacheDir       chezmoi.AbsPath
-	command        string
-	config         map[string]any
-	configFile     chezmoi.AbsPath
-	executable     chezmoi.AbsPath
-	fqdnHostname   string
-	gid            string
-	group          string
-	homeDir        chezmoi.AbsPath
-	hostname       string
-	kernel         map[string]any
-	os             string
-	osRelease      map[string]any
-	sourceDir      chezmoi.AbsPath
-	uid            string
-	username       string
-	version        map[string]any
-	windowsVersion map[string]any
-	workingTree    chezmoi.AbsPath
+	arch              string
+	args              []string
+	cacheDir          chezmoi.AbsPath
+	command           string
+	config            map[string]any
+	configFile        chezmoi.AbsPath
+	executable        chezmoi.AbsPath
+	fqdnHostname      string
+	gid               string
+	group             string
+	homeDir           chezmoi.AbsPath
+	hostname          string
+	kernel            map[string]any
+	os                string
+	osRelease         map[string]any
+	pathListSeparator string
+	pathSeparator     string
+	sourceDir         chezmoi.AbsPath
+	uid               string
+	username          string
+	version           map[string]any
+	windowsVersion    map[string]any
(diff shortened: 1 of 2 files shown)


arran4:
Okay, I was hoping I could get one in without making a mistake. Spelling failure!

internal/cmd/config.go
```diff
@@ -2263,24 +2267,26 @@ func (c *Config) newTemplateData(cmd *cobra.Command) *templateData {
...
+		fqdnHostname:      fqdnHostname,
+		gid:               gid,
+		group:             group,
+		homeDir:           c.homeDirAbsPath,
+		hostname:          hostname,
+		kernel:            kernel,
+		os:                runtime.GOOS,
+		osRelease:         osRelease,
+		pathListSeparator: string(os.PathListSeparator),
+		pathSeparator:     string(os.PathSeparator),
```
twpayne, line 2286: The typecasts to `string` are not needed here as the values are already strings.

arran4: In go 1.21 they are defined as runes.

https://github.com/golang/go/blob/3fb1d95149fa280343581a48547c3c3f70dac5fb/src/os/path_unix.go#L5-L13

They test as int32 too:

```golang
package psstr

import (
	"os"
	"reflect"
	"testing"
)

func TestName(t *testing.T) {
	to := reflect.TypeOf(os.PathSeparator)
	expected := "string"
	if to.String() != expected {
		t.Fatalf("PathSeparator is %s not %s", to.String(), expected)
	}
}
```

```console
=== RUN   TestOne
    main_test.go:13: PathSeparator is int32 not string
--- FAIL: TestOne (0.00s)

```

I can't test direct comparison:
```golang
func TestTwo(t *testing.T) {
	ps := os.PathSeparator
	expected := "/"
	if ps != expected {
		t.Fatalf("PathSeparator %s is not equal to %s", ps, expected)
	}
}
```
I get
```console
# psstr [psstr.test]
./main_test.go:20:11: invalid operation: ps != expected (mismatched types rune and string)

```

arran4: It doesn't look like the `os` package is being shadowed either, it's definitely being imported:
https://github.com/twpayne/chezmoi/blob/d637ddddaa8484622aa36e39eb6fe33d779bbfb3/internal/cmd/config.go#L12-L14

twpayne: I'm sorry, I was wrong. For some reason I was sure that they were constant strings, but you're right, they're runes.

twpayne:
Thank you!

internal/cmd/config.go
arran4: Yeah I've made the same mistake in the past. I guess it's a safe assumption that there aren't any multi-rune path separator OSes in common usage. I feel there is some weirdism in go around conversion between "primitive" / "native" types. Given everywhere else you have to be rather explicit

arran4:
No problem. Thanks for reviewing and merging.

Issue #3282, opened 2023-10-04
Error when adding a remove attribute


drod3763:
## Describe the bug

It throws an error when trying to add a remove attribute to a dotfile

## To reproduce

1. Open terminal in home folder
2. `chezmoi add ~/.zshrc` if you haven't already done so
3. `chezmoi chattr remove ~/.zshrc`

## Expected behavior

Adds the `.zshrc` file with the remove attribute

## Output of command with the `--verbose` flag

```console
$ chezmoi chattr --verbose remove ~/.zshrc
panic: 1: unknown source file type

goroutine 1 [running]:
github.com/twpayne/chezmoi/v2/internal/cmd.(*modifier).modifyFileAttr(0x1400059e0c0, {{0x1400077e1b8, 0x6}, 0x1, {0x0, 0x0}, 0x0, 0x0, 0x0, 0x0, ...})
        github.com/twpayne/chezmoi/v2/internal/cmd/chattrcmd.go:564 +0x54c
github.com/twpayne/chezmoi/v2/internal/cmd.(*Config).runChattrCmd(0x14000412000, 0x14000325b00?, {0x1400007cab0, 0x2, 0x3}, 0x140003f4f00)
        github.com/twpayne/chezmoi/v2/internal/cmd/chattrcmd.go:204 +0x3a4
github.com/twpayne/chezmoi/v2/internal/cmd.(*Config).newChattrCmd.(*Config).makeRunEWithSourceState.func2(0x1400080ee00?, {0x1400007cab0, 0x2, 0x3})
        github.com/twpayne/chezmoi/v2/internal/cmd/config.go:1496 +0x88
github.com/spf13/cobra.(*Command).execute(0x14000325b00, {0x1400007c990, 0x3, 0x3})
        github.com/spf13/cobra@v1.7.0/command.go:940 +0x658
github.com/spf13/cobra.(*Command).ExecuteC(0x14000395800)
        github.com/spf13/cobra@v1.7.0/command.go:1068 +0x320
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.7.0/command.go:992
github.com/twpayne/chezmoi/v2/internal/cmd.(*Config).execute(0x1400071fd08?, {0x140001c4010, 0x4, 0x4})
        github.com/twpayne/chezmoi/v2/internal/cmd/config.go:1185 +0x70
[...]

PR #3283, opened 2023-10-04
fix: Add chattr +remove support for files

halostatue:
Resolves: #3282

### DIFF
--- a/internal/cmd/testdata/scripts/chattr.txtar
+++ b/internal/cmd/testdata/scripts/chattr.txtar
@@ -6,6 +6,16 @@ exec chezmoi chattr empty $HOME${/}.file
 ! exists $CHEZMOISOURCEDIR/dot_file
 exists $CHEZMOISOURCEDIR/empty_dot_file
 
+# test that chezmoi chattr remove sets the remove attribute on a file
+exec chezmoi chattr remove $HOME${/}.file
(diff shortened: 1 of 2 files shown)


halostatue:
I should probably modify this PR and the commit, because `+remove` is already supported for directories, but not files.

internal/cmd/testdata/scripts/chattr.txtar
```diff
@@ -6,6 +6,16 @@ exec chezmoi chattr empty $HOME${/}.file
 ! exists $CHEZMOISOURCEDIR/dot_file
 exists $CHEZMOISOURCEDIR/empty_dot_file
 
+# test that chezmoi chattr remove sets the remove attribute on a file
+exec chezmoi chattr remove $HOME${/}.file
+! exists $CHEZMOISOURCEDIR/empty_dot_file
+exists $CHEZMOISOURCEDIR/remove_dot_file
+
+# test that chezmoi chattr noremove sets the remove attribute on a file
```
twpayne, line 14: nit: s/sets the remove attribute/removes the remove attribute/

twpayne (approved):
Thank you for fixing this :)

internal/cmd/testdata/scripts/chattr.txtar
halostatue: I’ll fix this this evening.

Issue #3101, opened 2023-07-18
Add dynamic scriptEnv


NavyD:
## Is your feature request related to a problem? Please describe.

When using py scripts, it is usually executed in the venv environment. it is necessary to configure the scriptEnv PATH environment variable for the py script to run in venv. Now, the PATH variable must be determined during chezmoi init. Once the venv environment changes, it must be reinit to take effect

## Describe the solution you'd like

The scriptEnv in the configuration file is processed as a template at runtime, such as

```toml
[scriptEnv]
PATH = '{{`path/to/venv/bin:{{env "PATH"}}`}}'
```

## Describe alternatives you've considered

Currently, the env variable PATH can only be fixed during init

```toml
[scriptEnv]
# {{- $venv_home := joinPath .chezmoi.sourceDir ".venv" }}
PATH = '{{$venv_home}}/bin{{ternary ";" ":" $is_win}}{{env "PATH"}}'
VIRTUAL_ENV = '{{$venv_home}}'
PYTHONHOME = ''
```

twpayne:
I don't fully understand what you want here.

If I understand correctly:
1. You have a Python virtual env which is in the `.venv` directory of your source directory, e.g. `~/.local/share/chezmoi/.venv`.
2. You want the `bin` subdirectory of this `.venv` directory, i.e. `~/.local/share/chezmoi/.venv/bin` to be the first element of `$PATH` when chezmoi runs scripts.

What I don't understand is:

> Once the venv environment changes, it must be reinit to take effect

What exactly do you mean by "the venv environment changes"? What changes exactly?

NavyD:
I apologize for the confusion caused by the wrong description

I am using a venv created with [poetry](https://python-poetry.org/) in the `~/.local/share/chezmoi/.venv` directory, but simply relying on chezmoi itself cannot make the `.chezmoiscripts/*.py` scripts run in the .venv environment. They can only run properly after activating .venv and then running `chezmoi apply`. The `interpreters.py=.venv/bin/python3` [interpreters](https://www.chezmoi.io/reference/target-types/#scripts-on-windows) configuration provided by chezmoi is ineffective on Linux. 

It is convenient to activate venv using `scriptEnv` in chezmoi without external handling for `.chezmoiscripts/*.py`. Otherwise, it may be necessary to rely on shell encapsulation of chezmoi, such as configuring alias, shim, etc.

[...]

PR #3293, opened 2023-10-13
feat: Add expand env vars in scriptEnv config

NavyD:
This PR References #3101 add expand env vars in scriptEnv config

```toml
[scriptEnv]
VIRTUAL_ENV = '/path/to/venv'
PATH = '$VIRTUAL_ENV/bin:$PATH'
```


twpayne:
Thank you for the PR but this will not work as the order of keys in a maps is not preserved.

NavyD:
Sorry, I am not very familiar with the golang and cannot find a way to preserve the keys configuration order. Please ignore this PR.

twpayne:
Indeed, there is no way to preserve key order without re-writing the TOML/JSON/YAML parser. Thanks for trying anyhow!

PR #3304, opened 2023-10-27
feat: adding `passage` password manager template functions

seanhagen:
Adds support for the [passage](https://github.com/FiloSottile/passage) password manager, a fork of the [pass](https://www.passwordstore.org/) password manager that uses [age](https://age-encryption.org/) instead of GnuPG.

This adds two new template functions: `passage` and `passageRaw`, which function the same as the `pass` and `passRaw` template functions.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/passage-functions/index.md
+++ b/assets/chezmoi.io/docs/reference/templates/passage-functions/index.md
@@ -1,0 +1,6 @@
+# passage functions
+
+The `passage` template functions return passwords stored in
+[passage](https://github.com/FiloSottile/passage) using the passage CLI
+(`passage`). `passage` is a fork of `pass` that uses the
+[age](https://age-encryption.org) encryption instead of GnuPG.
--- a/internal/cmd/passagetemplatefuncs.go
+++ b/internal/cmd/passagetemplatefuncs.go
@@ -1,0 +1,119 @@
+package cmd
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"regexp"
+
+	"github.com/coreos/go-semver/semver"
+
+	"github.com/twpayne/chezmoi/v2/internal/chezmoilog"
+)
+
+var (
+	// just requiring the current ( as of 2023-10-27 ) version of passage.
+	passageMinVersion  = semver.Version{Major: 1, Minor: 7, Patch: 4}
+	passageShowArgs    = []string{"show"}
+	passageVersionArgs = []string{"version"}
+	passageVersionRx   = regexp.MustCompile(`=\s+v(\d+\.\d+\.\d+)`)
+)
+
+type passageConfig struct {
+	Command   string `json:"command" mapstructure:"command" yaml:"command"`
+	versionOK bool
+	cache     map[string]string
+	rawCache  map[string][]byte
+}
+
+func (c *Config) passageTemplateFunc(id string) string {
+	c.passageValid()
+
+	if password, ok := c.Passage.cache[id]; ok {
+		return password
+	}
+
+	output := c.fetchPassage(id)
+	passwordBytes, _, _ := bytes.Cut(output, []byte{'\n'})
+	password := string(passwordBytes)
+
+	if c.Passage.cache == nil {
+		c.Passage.cache = map[string]string{}
+	}
+	c.Passage.cache[id] = password
+
+	return password
+}
+
+func (c *Config) passageRawTemplateFunc(id string) string {
+	c.passageValid()
+
+	if output, ok := c.Passage.rawCache[id]; ok {
+		return string(output)
+	}
+
+	output := c.fetchPassage(id)
+	if c.Passage.rawCache == nil {
+		c.Passage.rawCache = map[string][]byte{}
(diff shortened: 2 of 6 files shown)


assets/chezmoi.io/docs/reference/templates/passage-functions/index.md
```diff
@@ -0,0 +1,6 @@
+# passage functions
+
+The `passage` template functions return passwords stored in
+[passage](https://github.com/FiloSottile/passage) using the passage CLI
+(`passage`). `passage` is a fork of `pass` that uses the
```
halostatue, line 5: Remove `the` from this line.

halostatue (changes requested):
The changes here look reasonable, but there are no tests. For this PR to be considered, there must be passing tests for new functionality *when possible*. You will need to make a new test script `internal/cmd/testdata/scripts/passage.txtar` that is more or less a copy of `internal/cmd/testdata/scripts/pass.txtar`.

If we accept this, I believe that it would be good to add `passageFields` as there is `passFields`.

internal/cmd/passagetemplatefuncs.go
```diff
@@ -0,0 +1,119 @@
...
+	args = append(args, passageShowArgs...)
+	args = append(args, id)
+	output, err := c.passageOutput(args...)
+	if err != nil {
+		panic(err)
+	}
+	return output
+}
+
+func (c *Config) passageValid() {
```
halostatue, line 75: I do not believe that this function or the version check is necessary, especially since the tag version of the latest release differs from the version reported (`1.7.4a1` tag vs `1.7.4` reported).

Please remove it and `passageVersionCheck` as well as any supporting variables.

halostatue:
I have some *thoughts* about this PR that do not reflect on the code, but are more philosophical and may perhaps continue a discussion myself, @twpayne, @bradenhilton, and anyone else interested, particularly given Tom's [recent statement](https://github.com/twpayne/chezmoi/pull/3261#issuecomment-1734683428):

> chezmoi has already accumulated multiple template functions that, with hindsight, are not necessary. Most of the JSON variants of the password manager functions are not necessary because the user can add `| fromJson` to their template to achieve the same effect. With hindsight, I would love to remove a lot of chezmoi's existing template functions.

In short, I am not convinced that we need these new functions, since `passage` (and `gopass`) are *implicitly* supported by the `pass*` template functions when `pass.command` is set to either `passage` or `gopass`, both of which claim to be command and output compatible with `pass`. I have verified that this is the case with `passage`.

The only reason to consider having multiple variant functions is if more than one is being run for separate password stores. Even then, `{{ output "passage" "show" "foo" }}` returns the exact same data as `{{ passRaw "foo" }}` when `pass.command` is set to `passage`. The `pass` function is only slightly more difficult: `{{ output "passage" "show" "foo" | splitList "\n" | mustFirst }}`. Much more complex, *but still possible* is a `passFields` equivalent:

```
{{- $result := dict -}}
{{- range $line := (output "passage" "show" "test" | trim "\n" | splitList "\n") -}}
{{-   $parts := splitn ": " 2 $line -}}
{{-   $result = mustMergeOverwrite $result (dict $parts._0 $parts._1) -}}
{{- end -}}
{{- $result -}}
```

Yes, that's the same output. If this were to be added *now*, I would consider adding a `fromFields` mapping function similar to `fromJson` and I would have `{{ pass "foo" | fromFields }}` work as `passFields "foo"` (and I would have made `pass` work like `passRaw`).

As a side note, I believe that `passRaw`, `gopassRaw`, and `passageRaw` template functions are returning incorrect data because they include the terminal newline written by the `echo` output for `pass` and `passage` (and the API equivalence for `gopassRaw`, I suspect. We should probably document that it is desirable to use `| trimSuffix "\n"` when using the raw functions as I do not believe that changing the output would be correct for Chezmoi 2.x (although I would *strongly* suggest it for Chezmoi 3.x).

---

Once the changes requested are made (especially the missing tests) for this PR, we *can* merge this as the implementation is clean. What I don't know is whether we *should* merge this given that `passage` is supported via `pass.command` or `output "passage" "show" "<NAME>"`.

My personal preference would be to add an admonition to each of `pass`, `passRaw`, and `passFields` that indicate that either `passage` or `gopass` can be supported by setting `pass.command`.

halostatue (changes requested):

halostatue:
*IF* we merge this, a test for the binary should be added to `doctorcmd.go`:

```go
&binaryCheck{
  name:        "passage-command",
  binaryname:  c.Passage.Command,
  ifNotSet:    checkResultWarning,
  ifNotExist:  checkResultInfo,
  versionArgs: []string{"version"},
  versionRx:   regex.MustCompile(`(?m)=\s*v(\d+\.\d+\.\d+)`),
},
```

twpayne:
Thanks for this PR. If passage is a drop-in replacement for pass, then I think the functionality is better achieved by setting `pass.command` to `passage`, as @halostatue suggests.

Would you be able to modify the PR to update the documentation instead?

seanhagen:
For sure! I didn't even realize that was a thing I could do.

Issue #3311, opened 2023-11-03
Add support for a distinct `merge-all` command


ElanHasson:
I'd like to be able to specify a different command for `merge-all`.

I see it's really just looping through the files to be merged and calling `merge` for each one. 

https://github.com/twpayne/chezmoi/blob/692a1154a5ea7891776f0a520550b41508e5bdb5/internal/cmd/mergeallcmd.go#L70-L75

I am making a PR for this.

twpayne:
What command, exactly, do you want to run instead?

ElanHasson:
We can close this if you like @twpayne per the comment on the pr.

twpayne:
Awesome, thank you very much for the detailed follow-up in https://github.com/twpayne/chezmoi/pull/3313#issuecomment-1793092902. Much appreciated!

PR #3313, opened 2023-11-03
feat: Add support for custom merge command for `merge-all`

ElanHasson:
Resolves https://github.com/twpayne/chezmoi/issues/3311

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/merge-all.md
+++ b/assets/chezmoi.io/docs/reference/commands/merge-all.md
@@ -3,8 +3,21 @@
 Perform a three-way merge for file whose actual state does not match its target
 state. The merge is performed with `chezmoi merge`.
 
+A custom merge command may be configured by specifying an optional `merge`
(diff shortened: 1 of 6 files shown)


ElanHasson:
@twpayne, I tried running the tests and they all panic. I see CI is green, so I'm guessing maybe it has to do with my code spaces environment I am using? Trying `make test-docker`...

twpayne:
Before reviewing, what is the justification for this change? Specifically, what command do you want to run for `chezmoi merge-all` that is different to the command for `chezmoi merge`?

ElanHasson:
@twpayne i'm going to close this for now. I found a way to do it using the existing config.

I want to use vscode to merge and i was looking to have a new instance startup on the first file, and then re-use that window.

I was able to kinda wire it together based on `chezmoi` having the same pid in the loop through the files to merge. Then i was digging into the vscode repo and saw `--reuse window` will use the last active window and if there isn't one, start a new one, and then continue to use that.

Issue #3321, opened 2023-11-05
Variables supported by commitMessageTemplate


lgersman:
## Is your feature request related to a problem? Please describe.

At first : I love chezmoi since I first used it - it's a great piece of software ! 

I would like to use the commitMessageTemplate feature of chezmoi to have the machine name (or whatever available as environment variable) in the commit message. This is useful to detect on which machine what commit was created.    

I could not find any hint if this is possible right know. 

## Describe the solution you'd like

Support the environment variable references in commitMessageTemplate

twpayne:
Thanks for the kind words!

You can use the [`env` template function](https://masterminds.github.io/sprig/os.html) in your `git.commitMessageTemplate` string, for example:

```toml
[git]
    commitMessageTemplate = "commit with HOME={{ env \"HOME\" }}"
```

I'll add some functionality to improve the configurability of the commit message.

lgersman:
Thank you for your fast response!

Please consider updating the docs (at least with an example `git.commitMessageTemplate` using the Sprig `env` function or a reference to the supported Sprig features). 
I was not aware that I can already use env variables in `git.commitMessageTemplate` :+1: 

One more question : I am not (yet) a Go developer, but I have seen in the Sprig documentation that there exists a [getHostByName template function](https://masterminds.github.io/sprig/network.html). 

Can I use this function also in `git.commitMessageTemplate` to automatically inject the host name in the commit message (thats actually my use case) ?   

Like so ? 

```yaml
[git]
    commitMessageTemplate = "commit from host {{ getHostByName \"127.0.0.1\" }}"
```

twpayne:
> Please consider updating the docs (at least with an example `git.commitMessageTemplate` using the Sprig `env` function or a reference to the supported Sprig features).
> I was not aware that I can already use env variables in `git.commitMessageTemplate` 👍

The documentation states that `git.commitMessageTemplate` is a template. There is no indication that this template is any different from any other template in chezmoi, so I don't think such an example is necessary [^1].

[^1]: In fact, commit message templates were a bit special, but #3322 makes them much more like normal templates in chezmoi.

> One more question : I am not (yet) a Go developer, but I have seen in the Sprig documentation that there exists a [getHostByName template function](https://masterminds.github.io/sprig/network.html).
> 
> Can I use this function also in `git.commitMessageTemplate` to automatically inject the host name in the commit message (thats actually my use case) ?

[...]

PR #3322, opened 2023-11-06
feat: Improve configurability of automatic git commit messages

twpayne:
Refs #3321.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/generate.md
+++ b/assets/chezmoi.io/docs/reference/commands/generate.md
@@ -1,13 +1,15 @@
-# `generate` *file*
+# `generate` *output*
 
-Generates *file* for use with chezmoi. The currently supported *file*s are:
+Generates *output* for use with chezmoi. The currently supported *output*s are:
 
-| File         | Description                                                |
-| ------------ | ---------------------------------------------------------- |
-| `install.sh` | An install script, suitable for use with Github Codespaces |
+| Output               | Description                                                           |
+| -------------------- | --------------------------------------------------------------------- |
+| `git-commit-message` | A git commit message, describing the changes to the source directory. |
+| `install.sh`         | An install script, suitable for use with Github Codespaces            |
 
 !!! example
 
     ```console
     $ chezmoi generate install.sh > install.sh
+    $ chezmoi git commit -m "$(chezmoi generate git-commit-message)"
     ```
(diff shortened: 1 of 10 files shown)


halostatue (approved):

halostatue:
Looks good.

It would be nice if either of these options provided the template for `chezmoi git commit` to override the `commit-message` configuration, but that's a feature request.

halostatue (approved):

twpayne:
Thanks for review @halostatue!

> It would be nice if either of these options provided the template for `chezmoi git commit` to override the `commit-message` configuration, but that's a feature request.

Currently `chezmoi git` passes all following arguments directly through to `git`. Adding special behavior for `chezmoi git commit` would require chezmoi to parse git arguments exactly as git does, which is a very slippery slope. However, this is a great idea so I've added `chezmoi generate git-commit-message` to generate the git commit message so you can do:

```console
$ chezmoi git commit -m "$(chezmoi generate git-commit-message)"
```

halostatue:
Yeah. I thought that was a very clever way to do this and think it's appropriate.

lgersman:
> Thanks for review @halostatue!
> 
> > It would be nice if either of these options provided the template for `chezmoi git commit` to override the `commit-message` configuration, but that's a feature request.
> 
> Currently `chezmoi git` passes all following arguments directly through to `git`. Adding special behavior for `chezmoi git commit` would require chezmoi to parse git arguments exactly as git does, which is a very slippery slope. However, this is a great idea so I've added `chezmoi generate git-commit-message` to generate the git commit message so you can do:
> 
> ```
> $ chezmoi git commit -m "$(chezmoi generate git-commit-message)"
> ```

Looks good for me ❤️. 

BTW: Can I also provide a individual commit message when chezmoi autoCommit/autoPush is enabled ? 

```
[git]
  autoPush = true
  autoCommit = true
```

When I do a `chezmoi re-add` commit and push is done automagically by chezmoi in this case.

twpayne:
> BTW: Can I also provide a individual commit message when chezmoi autoCommit/autoPush is enabled ?

Yes. Use the `git.commitMessageTemplate` or `git.commitMessageTemplateFile` configuration variables to achieve this.

lgersman:
No, I mean not the template but an individual commit message. 

Lets say i have changed a file and want the commit message *not* generated from the `git.commitMessageTemplateFile` but individual for exactly this change. 

But with configured autoCommit  it's not possible to provide a individual commit message, is'nt it ?

twpayne:
You can use the `promptString` template function in your commit message template.

PR #3323, opened 2023-11-08
chore: Remove password manager version checks

twpayne:
When chezmoi initially added support for various password managers the password managers were changing rapidly with functionality being added in minor versions.

Now the password managers are more stable and it makes sense to assume that everyone has a recent-enough version and to punt the version check to `chezmoi doctor` instead.

Review notes:
1. This is in draft. There are unfixed errors.
2. Huge thanks particularly to @halostatue for handling the 1Password CLI version 1 to 2 transition.
3. Individual commits make more sense than the final diff.


halostatue:
Looks like a good start to me.

halostatue (approved):

PR #3330, opened 2023-11-09
chore: Disable tests on faulty Fedora

twpayne:
`go tool dist list` is broken on Fedora 39. See https://bugzilla.redhat.com/show_bug.cgi?id=2248782.


twpayne:
`govulncheck` is failing because of security problems in Go 1.21.3, which are fixed with #3324. Since multiple external things are failing simultaneously it is not possible to get CI to pass on separate branches. I'll close this and move everything into #3324.

PR #3343, opened 2023-11-15
feat: Add read-source-state hook

twpayne:
Refs #3342.

Work in progress. Small code change, big impact. Feedback appreciated.

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/hooks.md
+++ b/assets/chezmoi.io/docs/reference/configuration-file/hooks.md
@@ -1,29 +1,41 @@
 # Hooks
 
-chezmoi runs hooks before and after each chezmoi command as specified by the
-`hooks` configuration variable.
+Hook commands are executed before and after events. Unlike scripts, hooks are
+always run, even if `--dry-run` is specified. Hooks should be fast and
+idempotent.
 
-Before running *command*, chezmoi runs `hooks.`*command*`.pre.command` with the
-arguments `hooks.`*command*`.pre.args`. Similarly, after running *command*,
-chezmoi runs `hooks.`*command*`.post.command` with
-`hooks.`*command*`.post.args`.
+!!! warning
 
-When running hooks, the `CHEZMOI=1` and `CHEZMOI_*` environment variables will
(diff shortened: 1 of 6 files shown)


halostatue:
Seems like a potentially powerful change, as you say.

twpayne:
@bradenhilton @nandalopes @felipecrs I would appreciate your input on this.

felipecrs:
I will carefully analyze and provide my feedback in some hours.

twpayne:
No stress felipecrs and @halostatue I didn't mention because you're already in the conversation :)

felipecrs:
I believe this is a nice change. It will also be useful to install tools required by `output` template calls, like `curl` or `git`.

halostatue:
The code and documentation looks good. I think my main concern is that the commands need to be specified "full path". I'm not sure that there's a good workaround for this.

bradenhilton:
Would the password manager part benefit from being more general instead? A password manager is just one of many potential prerequisites as @felipecrs pointed out.

Either way, this is great.

halostatue (approved):

twpayne:
Thank you all for the reviews!

> I think my main concern is that the commands need to be specified "full path". I'm not sure that there's a good workaround for this.

Good point. It would be nice if the path to the `read-source-state` hook command was relative to the source directory.

One way I can think of doing this is by overriding the command (i.e. first checking for the command relative to the source state directory, and only afterwards falling back to the user's $PATH), but this falls foul of chezmoi's "the source state consists of only regular files and directories" principle as it would require the hook command to be executable in the source state.

Another way would be to use an explicit interpreter, something like:

```toml
[hooks.read-source-state.pre]
    command = "sh"
    args = ["-c", ".local/share/chezmoi/.install-password-manager.sh"]
```

Thoughts?

> Would the password manager part benefit from being more general instead? A password manager is just one of many potential prerequisites as @felipecrs pointed out.

Yes, absolutely! I used the password manager example as it's a problem that I personally have (on a new machine I need to install and setup 1Password before I can deploy my dotfiles). It would be great to have some more examples of adding prerequisites.

PR #3373, opened 2023-11-29
docs: Clarify prefix order for regular files

sumbach:
Based on the implementation here: https://github.com/twpayne/chezmoi/blob/6effb0ab449d0a22403e3fbf8a8d5b103604f7eb/internal/chezmoi/attr.go#L241-L246

I believe the prefix order for regular files is:
- `encrypted_`
- `private_`
- `readonly_`
- `empty_`
- `executable_`

The drift between the documentation and the implementation looks like it happened in two steps:

- `readonly_` was added to the implementation but excluded from the docs in #1403 
- `empty_` was added to the docs but in the wrong order by #2997 

Thanks for the awesome tool @twpayne !!!


twpayne:
Thank you :)

PR #3378, opened 2023-11-30
feat: add cpu and mem info to default data

accidentaldevelopment:
This adds cpu count and total memory to the default template data. I intended to use this for some configuration file generation, but found it wasn't available.

```
❯ go run . data | jq '.chezmoi | .cpus, .totalMemoryMB'
10
65536
```

CPU count was easy to get with `runtime.NumCPU()`.

Memory information required a [new dependency](https://github.com/pbnjay/memory). It hasn't been updated in a while, but the code seemed fine for every platform I could understand. I don't know enough about Windows to review that 😅.

Wasn't sure if this should be an issue to discuss, but it was simple enough I figured I'd just open the PR and see what happens!

Not attached to the field names at all, so happy to change.

Thank you!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -123,6 +123,7 @@ require (
 	github.com/muesli/cancelreader v0.2.2 // indirect
 	github.com/muesli/reflow v0.3.0 // indirect
 	github.com/olekukonko/tablewriter v0.0.5 // indirect
+	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58 // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pjbgf/sha1cd v0.3.0 // indirect
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
--- a/go.sum
+++ b/go.sum
@@ -268,6 +268,8 @@ github.com/olekukonko/tablewriter v0.0.5 h1:P2Ga83D34wi1o9J6Wh1mRuqd4mF/x/lgBS7N
 github.com/olekukonko/tablewriter v0.0.5/go.mod h1:hPp6KlRPjbx+hW8ykQs1w3UBbZlj6HuIJcUGPhkA7kY=
 github.com/onsi/gomega v1.27.10 h1:naR28SdDFlqrG6kScpT8VWpu1xWY5nJRCF3XaYyBjhI=
 github.com/onsi/gomega v1.27.10/go.mod h1:RsS8tutOdbdgzbPtzzATp12yT7kM5I5aElG3evPbQ0M=
+github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58 h1:onHthvaw9LFnH4t2DcNVpwGmV9E1BkGknEliJkfwQj0=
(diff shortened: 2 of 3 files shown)


halostatue:
I did a bit of digging, and it looks like gopsutil *may* be more flexible overall: https://github.com/shirou/gopsutil.

```go
package main

import (
	"github.com/k0kubun/pp/v3"
	"github.com/shirou/gopsutil/v3/mem"
	"github.com/shirou/gopsutil/v3/process"
	"github.com/shirou/gopsutil/v3/net"
	"github.com/shirou/gopsutil/v3/load"
	"github.com/shirou/gopsutil/v3/host"
	"github.com/shirou/gopsutil/v3/docker"
	"github.com/shirou/gopsutil/v3/disk"
	"github.com/shirou/gopsutil/v3/cpu"
)

func dump(xs ...any) {
	for _, x := range xs {
		pp.Println(x)
	}
}

func main() {
	dump(cpu.Info())
	dump(disk.Partitions(false))
	dump(docker.GetDockerStat())
	dump(host.BootTime())
	dump(host.Uptime())
	dump(host.HostID())
	dump(host.KernelArch())
	dump(host.KernelVersion())
	dump(host.PlatformInformation())
	dump(host.Virtualization())
	dump(host.Info())
	dump(host.Users())
	dump(host.SensorsTemperatures())
	dump(load.Avg())
	dump(load.Misc())
	dump(mem.SwapDevices())
	dump(mem.SwapMemory())
	dump(mem.VirtualMemory())
	dump(net.Interfaces())
	dump(process.Pids())
	dump(process.Processes())
}
```

accidentaldevelopment:
Didn't know about that library! It definitely can provide more information. My question is, is it needed? It's a much larger footprint to pull in just for total memory.

But if we want other data that `gopsutil` provides, happy to use it. I could see arguments for things like swap and disk info.

bradenhilton:
It's not a perfect 1-to-1, but we could considering removing https://github.com/twpayne/chezmoi/blob/c68ddac75c2c0e8964cc6fc30cc61afbf5d7ffba/assets/chezmoi.io/docs/user-guide/machines/general.md#L23-L58 if this is merged.

accidentaldevelopment:
@bradenhilton `NumCPU()` returns logical cpus, so probably shouldn't remove it entirely. I could update it though to show the "new" way with this PR.

 If I switch to `gopsutil` then physical and logical cores would be available, and could potentially remove the whole thing.

accidentaldevelopment:
Tried using `gopsutil`. I do like getting the physical and logical cpu counts, not that it makes much difference on this machine.

```
{
  "logical": 10,
  "physical": 10
}
65536
```

I'm good with either, happy to use whatever the group wants!

twpayne:
> Perhaps a better solution is to determine hardware in `.chezmoi.toml.tmpl` and then it's a one-time cost, until that template changes.

Yes, this is exactly what I'd recommend.

accidentaldevelopment:
Then that's what I'll do. I'll close this PR; thank you for your time!

halostatue:
@twpayne @accidentaldevelopment I *do* think that a reformulated version of the question asked here would be a *very good* addition to the `user-guide/machines/general.md` section for explaining *why* one might want to do this as well as the best place to do it (`.chezmoi.$FORMAT.tmpl` and why putting the discovery there but not in general config).

PR #3380, opened 2023-11-30
fix: Fix fromJson template function deserializing non-objects

### LINKED ISSUE #3379: fromJson regression in 2.42.x (it can no longer unmarshal arrays)
## Describe the bug

`fromJson` can no longer unmarshal arrays.

## To reproduce

```console
$  ./chezmoi2.41.0 execute-template '{{ fromJson "[{\"a\": 1}]" }}'
[map[a:1]]
$ ./chezmoi2.42.0 execute-template '{{ fromJson "[{\"a\": 1}]" }}'
chezmoi: template: arg1:1:3: executing "arg1" at <fromJson "[{\"a\": 1}]">: error calling fromJson: json: cannot unmarshal array into Go value of type map[string]interface {}
```

## Expected behavior

The above should return `[map[a:1]]`.

## Additional context

The problem appears to be https://github.com/twpayne/chezmoi/blob/c68ddac75c2c0e8964cc6fc30cc61afbf5d7ffba/internal/cmd/templatefuncs.go#L183, but replacing that with `any` results in everything falling back to the old (float64-only) behaviour.

I tried this:

```go
func (formatJSON) Unmarshal(data []byte, value any) error {
	var result interface{}

	decoder := json.NewDecoder(bytes.NewReader(data))
	decoder.UseNumber()
	if err := decoder.Decode(&result); err != nil {
		return err
	}

	if _, err := decoder.Token(); !errors.Is(err, io.EOF) {
		return errExpectedEOF
	}

	switch result := result.(type) {
	case *[]any:
		value = replaceJSONNumbersWithNumericValuesSlice(*result)
		return nil
	case *map[string]any:
		value = replaceJSONNumbersWithNumericValuesMap(*result)
		return nil
	default:
		value = result
		return nil
	}
}
```

But that causes other failures that I don't have time to investigate right now.

This breaks my dotfiles.

twpayne:
Fixes #3379.

### DIFF
--- a/internal/cmd/templatefuncs.go
+++ b/internal/cmd/templatefuncs.go
@@ -180,7 +180,7 @@ func (c *Config) fromIniTemplateFunc(s string) map[string]any {
 //
 //nolint:revive,stylecheck
 func (c *Config) fromJsonTemplateFunc(s string) any {
-	var value map[string]any
+	var value any
 	if err := chezmoi.FormatJSON.Unmarshal([]byte(s), &value); err != nil {
 		panic(err)
 	}
@@ -190,7 +190,7 @@ func (c *Config) fromJsonTemplateFunc(s string) any {
 // fromJsoncTemplateFunc parses s as JSONC and returns the result. In contrast
 // to encoding/json, numbers are represented as int64s or float64s if possible.
 func (c *Config) fromJsoncTemplateFunc(s string) any {
-	var value map[string]any
+	var value any
 	if err := chezmoi.FormatJSONC.Unmarshal([]byte(s), &value); err != nil {
 		panic(err)
 	}
@@ -198,7 +198,7 @@ func (c *Config) fromJsoncTemplateFunc(s string) any {
 }
 
 func (c *Config) fromTomlTemplateFunc(s string) any {
-	var value map[string]any
+	var value any
 	if err := chezmoi.FormatTOML.Unmarshal([]byte(s), &value); err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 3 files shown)


halostatue (approved):

internal/cmd/templatefuncs.go
```diff
@@ -190,23 +190,23 @@ func (c *Config) fromJsonTemplateFunc(s string) any {
...
+	var value any
 	if err := chezmoi.FormatJSONC.Unmarshal([]byte(s), &value); err != nil {
 		panic(err)
 	}
 	return value
 }
 
 func (c *Config) fromTomlTemplateFunc(s string) any {
-	var value map[string]any
+	var value any
```
halostatue, line 201: I don't think that TOML allows anything but a dictionary at the root, so this is probably fine as it was.

internal/cmd/templatefuncs.go
```diff
@@ -190,23 +190,23 @@ func (c *Config) fromJsonTemplateFunc(s string) any {
...
 	if err := chezmoi.FormatTOML.Unmarshal([]byte(s), &value); err != nil {
 		panic(err)
 	}
 	return value
 }
 
 func (c *Config) fromYamlTemplateFunc(s string) any {
-	var value map[string]any
+	var value any
```
halostatue, line 209: This is probably a good thing as

```yaml
- 1
- 2
- 3
```

is a thing.

halostatue:
This does fix the test case.

internal/cmd/templatefuncs.go
twpayne: Thanks! Fixed in https://github.com/twpayne/chezmoi/commit/e68328eecd27255aac0a44ca33055f9c02813238.

PR #3397, opened 2023-12-06
docs: add warning for GitHub SSH key deletion

Logicer16:
I (used to) use just the `gitHubKeys` function to populate my `~/.ssh/authorized_keys` file, but as I usually access GitHub over HTTPS, GitHub removed one of my SSH Keys due to it being "[stale](https://docs.github.com/en/authentication/troubleshooting-ssh/deleted-or-missing-ssh-keys)". If I didn't have another key on a different device, I would have lost access to machines which I exclusively interact with over SSH. 

As such, this pr expands upon the existing warning in the documentation to advise users to have other keys alongside the `gitHubKeys` output to prevent the same mistake I made.


twpayne:
Wow, very interesting failure mode. Thanks for the warning!

PR #3408, opened 2023-12-14
Set BINDIR to absolute path in install scripts

frantret:
Dear twpayne,

I was surprised to see that the one-line binary install scripts install the tool in 'bin' (resp '.local/bin') directory, _relative_ to the current working directory. After reading again the doc chezmoi.io/install, the section "One-line binary install", I understood this behaviour is intentional, and the scripts indeed provide a way to install the tool in e.g. '$HOME/.local/bin'.

However, I would assume that most users expect the install scripts' default target to be '$HOME/bin' (resp '$HOME/.local/bin') directory. Provided the scripts keep a way for them to specify a custom one, I would think this is fair.

Apologies if there are mistakes in the code, I tested as much as I could, but I'm no Go nor Powershell programmer.

Thank you for your attention

### DIFF
--- a/assets/chezmoi.io/docs/install.md.tmpl
+++ b/assets/chezmoi.io/docs/install.md.tmpl
@@ -130,7 +130,7 @@ repology.org](https://repology.org/project/chezmoi/versions).
 
 ## One-line binary install
 
-Install the correct binary for your operating system and architecture in `./bin`
+Install the correct binary for your operating system and architecture in `$HOME/bin`
 with a single command:
 
 === "curl"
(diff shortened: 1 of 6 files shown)


twpayne:
Thanks for the PR. However, changing the existing behavior would break a the scripts of people who rely on the existing behavior. The install script has a `-b` flag to set the destination directory so you can pass `-b $HOME/bin` to achieve the behavior you want.

frantret:
Fair! Thank you for taking time to reply.
(Besides, I then realised my commits messages don't comply with the project's contribution policy - sorry for that.)

PR #3413, opened 2023-12-17
fix: key.txt is not written to home, but home/.config

zeidlos:
Found an error in the "encryption" documentation.

Fixed it for you. ;)


twpayne:
Good catch, thanks!

zeidlos:
Sadly it doesn't work as per documentation on my machine. I'll try to find out why and either update the documentation or write a bug report.

PR #3459, opened 2024-01-07
Fix structs alignment

lukaszraczylo:
This commit fixes the alignment of structs for the whole project, improving performance and decreasing memory usage :)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -127,7 +127,6 @@ linters-settings:
     local-prefixes: github.com/twpayne/chezmoi
   govet:
     disable:
-    - fieldalignment
     - shadow
     enable-all: true
   ireturn:
--- a/internal/archivetest/archivetest.go
+++ b/internal/archivetest/archivetest.go
@@ -7,14 +7,14 @@ import (
 
 // A Dir represents a directory.
 type Dir struct {
-	Perm    fs.FileMode
 	Entries map[string]any
+	Perm    fs.FileMode
 }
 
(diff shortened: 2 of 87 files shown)


twpayne:
> improving performance and decreasing memory usage :)

What sort of improvement is expected from this PR? What improvement have you measured?

chezmoi is predominantly I/O bound, and even if it handles thousands of files, saving a few bytes per struct means that total memory savings might be a few kilobytes in a program than handles multiple megabytes of data.

Currently fields are ordered in structs for consistency and readability. There would need to be a measurable improvement from this PR for it to be merged.

twpayne:
I am not questioning whether there is a slight performance improvement or slight reduction in memory usage. Both are likely.

What is not evident is whether this PR has any measurable effect on chezmoi's performance. I believe that it does not, and that it has the negative effect of reducing the readability and maintainability of chezmoi.

If you can provide direct, reproducible, measurements of chezmoi's performance that demonstrate that this PR has a non-trivial effect, then I will accept it. If no such evidence is available then sorry but the negative effects on maintainability mean that I will not accept this PR.

lukaszraczylo:
@twpayne I started using chezmoi two days ago, so I'll get back to you once I familiarise myself with it more and then prepare the performance tests, deal? :)

twpayne:
Sounds good, thanks!

twpayne:
Gently closing this issue. Please re-open when needed.

PR #3471, opened 2024-01-11
fix: Improve robustness of keepassxc-cli integration in open mode

### LINKED ISSUE #3468: keepassxc "open" mode not working in non-english environment
## Describe the bug

After releasing v2.44.0 i try to use the new open mode of keepasscx because i plan to use a lot of secrets and opening/closing the database for every call isn't optimal and keep the database open could maybe lead to an speed increase.
After setting the mode to open `chezmoi apply` hangs. The prozessierst shows the open command but no password prompt is shown.
After quick check the code, `Enter password to unlock` is expected but as my system has
```
LANG=de_DE.UTF-8
LC_COLLATE=C
LANGUAGE=de:en
```
i got `Passwort zum Entsperren von`.

So i can workaround this with `LANGUAGE=en chezmoi apply` but then i got errors for opening the temporary attachment file
```
error calling keepassxcAttachment: open /tmp/chezmoi-keepassxc3644937192/file: no such file or directory
```

I'm not sure if the follow up error is related but i have no idea how to debug this further.

## To reproduce

Set keepasscx mode to open and run `LANGUAGE=de chezmoi apply`.

## Expected behavior

The template should render correctly.

twpayne:
Fixes #3468. Not ready yet.

### DIFF
--- a/internal/cmd/keepassxctemplatefuncs.go
+++ b/internal/cmd/keepassxctemplatefuncs.go
@@ -8,7 +8,9 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/Netflix/go-expect"
 	"github.com/coreos/go-semver/semver"
@@ -72,7 +74,7 @@ func (c *Config) keepassxcAttachmentTemplateFunc(entry, name string) string {
 		if err != nil {
 			panic(err)
 		}
-		tempFilename := tempDir.JoinString(name).String()
+		tempFilename := tempDir.JoinString("attachment-" + strconv.FormatInt(time.Now().UnixNano(), 10)).String()
 		if _, err := c.keepassxcOutputOpen("attachment-export", "--quiet", entry, name, tempFilename); err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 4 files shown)


twpayne:
@halostatue I'd appreciate your input on this.

See https://github.com/twpayne/chezmoi/issues/3468#issuecomment-1890185884 for an explanation of the changes.

internal/cmd/keepassxctemplatefuncs.go
```diff
@@ -72,7 +72,7 @@ func (c *Config) keepassxcAttachmentTemplateFunc(entry, name string) string {
 		if err != nil {
 			panic(err)
 		}
-		tempFilename := tempDir.JoinString(name).String()
+		tempFilename := tempDir.JoinString("attachment").String()
```
halostatue, line 75: I don't think we are using goroutines now to run various resolutions in parallel, but if we ever do, this could become problematic. It might be better to somehow randomize `attachment` during execution, or perhaps make it a timestamp (`attachment-20060102030405`)?

halostatue (approved):

internal/cmd/keepassxctemplatefuncs.go
twpayne: Good idea. Done.

PR #3472, opened 2024-01-12
feat: Scan for secrets with gitleaks when adding files

twpayne:
Prompted by @breisfeld's reason for [wanting files encrypted by default](https://github.com/twpayne/chezmoi/discussions/3469#discussioncomment-8098391):

> I wanted encryption as the default, since I anticipate forgetting that I have sensitive information included in a file I wish to `add`.

On reflection, why not scan for known secrets by default?

This PR implements that, using [`github.com/aquasecurity/trivy`](https://github.com/aquasecurity/trivy)'s [list of known secret patterns](https://github.com/aquasecurity/trivy/blob/ae134a9b3884c5fedaad1400863374e2c8c6897e/pkg/fanal/secret/builtin-rules.go).

This PR is draft right now because trivy [claims to require Go 1.21](https://github.com/aquasecurity/trivy/blob/ae134a9b3884c5fedaad1400863374e2c8c6897e/go.mod#L3), which in turn bumps chezmoi to require Go 1.21 too. However, chezmoi supports the old stable version of Go (currently Go 1.20).

cc @halostatue for his excellent reviews


halostatue:
I’ll note that I use [gitleaks/gitleaks](https://github.com/gitleaks/gitleaks) which only requires go 1.19. I’m not sure whether it is as easily embedded. The nice thing is that there is a *wealth* of options for secret scanning now.

halostatue (commented):
Looks good overall and I think it will make a good addition. Trivy is as good a scanner as any other I have played with.

When we get to *customization*, my recommendation is that the `trivy-secret.yaml` by default be loaded in `$CHEZMOI_WORKING_TREE` rather than `$CHEZMOI_SOURCE_DIR`. Or we have a `.chezmoitrivy.yaml` file that lives in `$CHEZMOI_SOURCE_DIR`. I would *not* make this `.chezmoitrivy.$FORMAT` because Trivy only documents a YAML file and the specific configuration items they list are complex enough that YAML is going to be more readable overall.
Using the Trivy load functions for the chezmoi-trivy config file would be, IMO, the correct direction.

If we don't wish to keep the secrets configuration *there*, then as a sibling to `$CHEZMOI_CONFIG_FILE` would be ideal.

twpayne:
Thanks for the review @halostatue!

I've updated the PR to use gitleaks instead of trivy.

For the moment, I think it's easier to not make the secret scanning configurable. People can use either a chezmoi hook or a git pre-commit hook to add more secret scanning if they want.

halostatue (approved):
This looks really good.

Not so much for this pass, but:

- Should we add a suggestion to use `--template` if secrets are found?
- Should we have a "severity" level where `--template` is automatically added if secrets are found?
- Could we use the finding data to "auto-template" secrets when `--template` is used and secrets are found?

twpayne:
Thanks for the re-review and ideas!

> * Should we add a suggestion to use `--template` if secrets are found?

I don't think this is necessary.

> * Should we have a "severity" level where `--template` is automatically added if secrets are found?

It's still up to the user to make the template substitutions, so I don't think this adds much.

> * Could we use the finding data to "auto-template" secrets when `--template` is used and secrets are found?

I don't think this level of automation is possible. It would require chezmoi to know how the user wants to store the secret, and for chezmoi to be able to write to that storage.

PR #3489, opened 2024-01-16
fix: Change init to default to --config if --config-path is absent

### LINKED ISSUE #3127: The init command should default to the global --config when no --config-path is provided
## Describe the bug

I have two chezmoi repositories under the same account. The first repository uses the default config file in `~/.config/chezmoi/chezmoi.toml`. The second repository uses a custom config file. To run any command for the second repository I have to run chezmoi with the global flag `--config <path-to-custom-config-file>`. As far as I can tell, this works for all commands except for the `init` command. Whenever I modify the `.chezmoi.toml.tmpl` file, chezmoi tells me that I have to rerun the `init` command. However, if I do so without providing the `init` flag `--config-path <path-to-custom-config-file>`, it fails to write to the custom config file and instead writes to `~/.config/chezmoi/chezmoi.toml` and corrupts my first chezmoi repository.

## To reproduce

The attached file contains the full script displayed below.

```console
$ rm -rfI test-chezmoi || exit
$ mkdir test-chezmoi
$ cd test-chezmoi
$ home=$PWD
$ mkdir config
$ git init athome
Initialized empty Git repository in /Users/paltherr/test-chezmoi/athome/.git/
$ echo "sourceDir = \"$home/athome\"" >> athome/.chezmoi.toml.tmpl
$ echo "destDir = \"$home\"" >> athome/.chezmoi.toml.tmpl
$ chezmoi --config $home/config/athome.toml --source $home/athome init --config-path $home/config/athome.toml
$ chezmoi --config $home/config/athome.toml status
$ chezmoi --config $home/config/athome.toml doctor
```

<details>

```console
ok        version              v2.36.0, commit be6e90307f9b5c8fa5d356183418a16d7bba69bf, built at 2023-07-28T17:59:29Z, built by Homebrew
ok        latest-version       v2.36.0
ok        os-arch              darwin/amd64
ok        uname                Darwin maloja 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:51:50 PDT 2023; root:xnu-8796.121.2~5/RELEASE_X86_64 x86_64
ok        go-version           go1.20.6 (gc)
ok        executable           /usr/local/bin/chezmoi
ok        upgrade-method       replace-executable
failed    config-file          found ~/.config/chezmoi/chezmoi.toml, expected ~/test-chezmoi/config/athome.toml
warning   source-dir           ~/test-chezmoi/athome is a git working tree (dirty)
ok        suspicious-entries   no suspicious entries
warning   working-tree         ~/test-chezmoi/athome is a git working tree (dirty)
ok        dest-dir             ~/test-chezmoi is a directory
ok        umask                022
ok        cd-command           found /bin/zsh
ok        cd-args              /bin/zsh
ok        edit-command         found /usr/bin/vi
ok        edit-args            /usr/bin/vi
ok        git-command          found /usr/local/bin/git, version 2.41.0
ok        merge-command        found /usr/bin/vimdiff
ok        shell-command        found /bin/zsh
ok        shell-args           /bin/zsh
info      gpg-command          gpg not found in $PATH
info      dashlane-command     dcli not found in $PATH
info      keepassxc-command    keepassxc-cli not found in $PATH
info      keeper-command       keeper not found in $PATH
info      passhole-command     ph not found in $PATH
info      rbw-command          rbw not found in $PATH
info      vlt-command          vlt not found in $PATH
```

</details>

```console
$ ls -la config
total 72
drwxr-xr-x@ 4 paltherr  staff    128 Jul 29 03:46 .
drwxr-xr-x@ 4 paltherr  staff    128 Jul 29 03:46 ..
-rw-------@ 1 paltherr  staff     91 Jul 29 03:46 athome.toml
-rw-------@ 1 paltherr  staff  32768 Jul 29 03:46 chezmoistate.boltdb
$ cat config/athome.toml
sourceDir = "/Users/paltherr/test-chezmoi/athome"
destDir = "/Users/paltherr/test-chezmoi"
```

At this point everything is OK and I can use my second chezmoi repository as expected. I just have to include `--config $home/config/athome.toml` in each command. Interestingly the `doctor` command complains about `config-file` (see details above) but nevertheless everything works as expected.

If I now modify the `.chezmoi.toml.tmpl` file, chezmoi tells me to rerun `init`:

```console
$ echo "# foo" >> athome/.chezmoi.toml.tmpl
$ chezmoi --config $home/config/athome.toml status
chezmoi: warning: config file template has changed, run chezmoi init to regenerate config file
```

If I do so with a `--config-path` then the custom config file gets updated and everything gets back to normal (but `doctor` still complains about `config-file`).

```console
$ chezmoi --config $home/config/athome.toml --source $home/athome init --config-path $home/config/athome.toml
$ chezmoi --config $home/config/athome.toml status
$ chezmoi --config $home/config/athome.toml doctor
```

<details>

```console
ok        version              v2.36.0, commit be6e90307f9b5c8fa5d356183418a16d7bba69bf, built at 2023-07-28T17:59:29Z, built by Homebrew
ok        latest-version       v2.36.0
ok        os-arch              darwin/amd64
ok        uname                Darwin maloja 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:51:50 PDT 2023; root:xnu-8796.121.2~5/RELEASE_X86_64 x86_64
ok        go-version           go1.20.6 (gc)
ok        executable           /usr/local/bin/chezmoi
ok        upgrade-method       replace-executable
failed    config-file          found ~/.config/chezmoi/chezmoi.toml, expected ~/test-chezmoi/config/athome.toml
warning   source-dir           ~/test-chezmoi/athome is a git working tree (dirty)
ok        suspicious-entries   no suspicious entries
warning   working-tree         ~/test-chezmoi/athome is a git working tree (dirty)
ok        dest-dir             ~/test-chezmoi is a directory
ok        umask                022
ok        cd-command           found /bin/zsh
ok        cd-args              /bin/zsh
ok        edit-command         found /usr/bin/vi
ok        edit-args            /usr/bin/vi
ok        git-command          found /usr/local/bin/git, version 2.41.0
ok        merge-command        found /usr/bin/vimdiff
ok        shell-command        found /bin/zsh
ok        shell-args           /bin/zsh
info      gpg-command          gpg not found in $PATH
info      dashlane-command     dcli not found in $PATH
info      keepassxc-command    keepassxc-cli not found in $PATH
info      keeper-command       keeper not found in $PATH
info      passhole-command     ph not found in $PATH
info      rbw-command          rbw not found in $PATH
info      vlt-command          vlt not found in $PATH
```
</details>

```console
$ ls -la config
total 72
drwxr-xr-x@ 4 paltherr  staff    128 Jul 29 03:46 .
drwxr-xr-x@ 4 paltherr  staff    128 Jul 29 03:46 ..
-rw-------@ 1 paltherr  staff     97 Jul 29 03:46 athome.toml
-rw-------@ 1 paltherr  staff  32768 Jul 29 03:46 chezmoistate.boltdb
$ cat config/athome.toml
sourceDir = "/Users/paltherr/test-chezmoi/athome"
destDir = "/Users/paltherr/test-chezmoi"
# foo
```

If I modify again the `.chezmoi.toml.tmpl` file but rerun `init` with no `--config-path` then chezmoi fails to update the custom config file. Interestingly it still updates the `chezmoistate.boltdb` file next to custom config file (which can't be seen in the output here but I could observe that its timestamp was updated when I ran the command manually). It's also interesting to note that afterwards the `status` command no longer asks to rerun `init`.

```console
$ echo "# bar" >> athome/.chezmoi.toml.tmpl
$ chezmoi --config $home/config/athome.toml status
chezmoi: warning: config file template has changed, run chezmoi init to regenerate config file
$ chezmoi --config $home/config/athome.toml --source $home/athome init
$ chezmoi --config $home/config/athome.toml status
$ chezmoi --config $home/config/athome.toml doctor
```

<details>

```console
ok        version              v2.36.0, commit be6e90307f9b5c8fa5d356183418a16d7bba69bf, built at 2023-07-28T17:59:29Z, built by Homebrew
ok        latest-version       v2.36.0
ok        os-arch              darwin/amd64
ok        uname                Darwin maloja 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:51:50 PDT 2023; root:xnu-8796.121.2~5/RELEASE_X86_64 x86_64
ok        go-version           go1.20.6 (gc)
ok        executable           /usr/local/bin/chezmoi
ok        upgrade-method       replace-executable
failed    config-file          found ~/.config/chezmoi/chezmoi.toml, expected ~/test-chezmoi/config/athome.toml
warning   source-dir           ~/test-chezmoi/athome is a git working tree (dirty)
ok        suspicious-entries   no suspicious entries
warning   working-tree         ~/test-chezmoi/athome is a git working tree (dirty)
ok        dest-dir             ~/test-chezmoi is a directory
ok        umask                022
ok        cd-command           found /bin/zsh
ok        cd-args              /bin/zsh
ok        edit-command         found /usr/bin/vi
ok        edit-args            /usr/bin/vi
ok        git-command          found /usr/local/bin/git, version 2.41.0
ok        merge-command        found /usr/bin/vimdiff
ok        shell-command        found /bin/zsh
ok        shell-args           /bin/zsh
info      gpg-command          gpg not found in $PATH
info      dashlane-command     dcli not found in $PATH
info      keepassxc-command    keepassxc-cli not found in $PATH
info      keeper-command       keeper not found in $PATH
info      passhole-command     ph not found in $PATH
info      rbw-command          rbw not found in $PATH
info      vlt-command          vlt not found in $PATH
```

</details>

```console
$ ls -la config
total 72
drwxr-xr-x@ 4 paltherr  staff    128 Jul 29 03:46 .
drwxr-xr-x@ 4 paltherr  staff    128 Jul 29 03:46 ..
-rw-------@ 1 paltherr  staff     97 Jul 29 03:46 athome.toml
-rw-------@ 1 paltherr  staff  32768 Jul 29 03:46 chezmoistate.boltdb
$ cat config/athome.toml
sourceDir = "/Users/paltherr/test-chezmoi/athome"
destDir = "/Users/paltherr/test-chezmoi"
# foo
```

Instead of updating the custom config file, the `init` command updated the default config file `~/.config/chezmoi/chezmoi.toml`:

```console
$ cat $HOME/.config/chezmoi/chezmoi.toml
sourceDir = "/Users/paltherr/test-chezmoi/athome"
destDir = "/Users/paltherr/test-chezmoi"
# foo
# bar
```

## Expected behavior

The `init` command should default `--config-path` to `--config` if the latter was provided but not the former.

[test-chezmoi.txt](https://github.com/twpayne/chezmoi/files/12206108/test-chezmoi.txt)

paltherr:
Here is an updated version of my previous PR. I have checked that it does what's expected and that it doesn't break any existing tests.

It should still be considered as a proof of concept rather than something ready to be submitted. My claim is that it contains the required logic to achieve the desired goal without breaking any existing behavior but that's about it. I haven't tried to add tests, yet. More importantly some changes, like in particular the replacement of `configFileAbsPath` with `configFileAbsPathDefault` and `configFileAbsPathExplicit`, look ugly. Note also that I have no experience programming in Go.

The main benefit of this change is that it becomes possible to use `chezmoi` with an alternate config simply by defining an alias like the following one:

`alias chezmoi2="chezmoi --config=<path-to-alternate-config>"`

This works for all commands, including the initial `init` command.

Fixes #3127.

### DIFF
--- a/internal/cmd/addcmd.go
+++ b/internal/cmd/addcmd.go
@@ -208,7 +208,7 @@ func (c *Config) runAddCmd(cmd *cobra.Command, args []string, sourceState *chezm
 			ProtectedAbsPaths: []chezmoi.AbsPath{
 				c.CacheDirAbsPath,
 				c.WorkingTreeAbsPath,
-				c.configFileAbsPath,
+				c.getConfigFileAbsPath(),
 				persistentStateFileAbsPath,
 				c.sourceDirAbsPath,
 			},
--- a/internal/cmd/catconfigcmd.go
+++ b/internal/cmd/catconfigcmd.go
@@ -20,7 +20,7 @@ func (c *Config) newCatConfigCmd() *cobra.Command {
 }
 
 func (c *Config) runCatConfigCmd(cmd *cobra.Command, args []string) error {
-	data, err := c.baseSystem.ReadFile(c.configFileAbsPath)
+	data, err := c.baseSystem.ReadFile(c.getConfigFileAbsPath())
 	if err != nil {
 		return err
 	}
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -215,16 +215,17 @@ type Config struct {
 	versionStr  string
 
 	// Configuration.
-	fileSystem           vfs.FS
-	bds                  *xdg.BaseDirectorySpecification
-	configFileAbsPath    chezmoi.AbsPath
-	configFileAbsPathErr error
-	baseSystem           chezmoi.System
-	sourceSystem         chezmoi.System
-	destSystem           chezmoi.System
-	persistentState      chezmoi.PersistentState
-	httpClient           *http.Client
-	logger               *zerolog.Logger
+	fileSystem                  vfs.FS
+	bds                         *xdg.BaseDirectorySpecification
+	defaultConfigFileAbsPath    chezmoi.AbsPath
+	defaultConfigFileAbsPathErr error
+	customConfigFileAbsPath     chezmoi.AbsPath
+	baseSystem                  chezmoi.System
+	sourceSystem                chezmoi.System
+	destSystem                  chezmoi.System
+	persistentState             chezmoi.PersistentState
+	httpClient                  *http.Client
+	logger                      *zerolog.Logger
 
 	// Computed configuration.
 	commandDirAbsPath   chezmoi.AbsPath
@@ -499,7 +500,7 @@ func newConfig(options ...configOption) (*Config, error) {
 	if err != nil {
 		return nil, err
 	}
-	c.configFileAbsPath, c.configFileAbsPathErr = c.defaultConfigFile(c.fileSystem, c.bds)
+	c.defaultConfigFileAbsPath, c.defaultConfigFileAbsPathErr = c.defaultConfigFile(c.fileSystem, c.bds)
 	c.SourceDirAbsPath, err = c.defaultSourceDir(c.fileSystem, c.bds)
 	if err != nil {
 		return nil, err
@@ -508,6 +509,13 @@ func newConfig(options ...configOption) (*Config, error) {
 	c._import.destination = c.homeDirAbsPath
 
 	return c, nil
+}
+
+func (c *Config) getConfigFileAbsPath() chezmoi.AbsPath {
+	if c.customConfigFileAbsPath.Empty() {
+		return c.defaultConfigFileAbsPath
+	}
+	return c.customConfigFileAbsPath
 }
 
 // Close closes resources associated with c.
@@ -748,7 +756,11 @@ func (c *Config) createAndReloadConfigFile(cmd *cobra.Command) error {
 	// Write the config.
 	configPath := c.init.configPath
 	if c.init.configPath.Empty() {
-		configPath = chezmoi.NewAbsPath(c.bds.ConfigHome).Join(chezmoiRelPath, configTemplate.targetRelPath)
+		if c.customConfigFileAbsPath.Empty() {
+			configPath = chezmoi.NewAbsPath(c.bds.ConfigHome).Join(chezmoiRelPath, configTemplate.targetRelPath)
+		} else {
+			configPath = c.customConfigFileAbsPath
+		}
 	}
 	if err := chezmoi.MkdirAll(c.baseSystem, configPath.Dir(), fs.ModePerm); err != nil {
 		return err
@@ -1181,6 +1193,7 @@ func (c *Config) execute(args []string) error {
 		return err
 	}
 	rootCmd.SetArgs(args)
+
 	return rootCmd.Execute()
 }
 
@@ -1552,7 +1565,7 @@ func (c *Config) newRootCmd() (*cobra.Command, error) {
 	persistentFlags.BoolVarP(&c.Verbose, "verbose", "v", c.Verbose, "Make output more verbose")
 	persistentFlags.VarP(&c.WorkingTreeAbsPath, "working-tree", "W", "Set working tree directory")
 
-	persistentFlags.VarP(&c.configFileAbsPath, "config", "c", "Set config file")
+	persistentFlags.VarP(&c.customConfigFileAbsPath, "config", "c", "Set config file")
 	persistentFlags.Var(&c.configFormat, "config-format", "Set config file format")
 	persistentFlags.Var(&c.cpuProfile, "cpu-profile", "Write a CPU profile to path")
 	persistentFlags.BoolVar(&c.debug, "debug", c.debug, "Include debug information in output")
@@ -1749,7 +1762,7 @@ func (c *Config) persistentPostRunRootE(cmd *cobra.Command, args []string) error
 	}
 
 	if annotations.hasTag(modifiesConfigFile) {
-		configFileContents, err := c.baseSystem.ReadFile(c.configFileAbsPath)
+		configFileContents, err := c.baseSystem.ReadFile(c.getConfigFileAbsPath())
 		switch {
 		case errors.Is(err, fs.ErrNotExist):
 			err = nil
@@ -1757,7 +1770,7 @@ func (c *Config) persistentPostRunRootE(cmd *cobra.Command, args []string) error
 			// err is already set, do nothing.
 		default:
 			var format chezmoi.Format
-			if format, err = chezmoi.FormatFromAbsPath(c.configFileAbsPath); err == nil {
+			if format, err = chezmoi.FormatFromAbsPath(c.getConfigFileAbsPath()); err == nil {
 				var config map[string]any
 				if err = format.Unmarshal(configFileContents, &config); err != nil { //nolint:revive
 					// err is already set, do nothing.
@@ -1767,7 +1780,7 @@ func (c *Config) persistentPostRunRootE(cmd *cobra.Command, args []string) error
 			}
 		}
 		if err != nil {
-			c.errorf("warning: %s: %v\n", c.configFileAbsPath, err)
+			c.errorf("warning: %s: %v\n", c.getConfigFileAbsPath(), err)
 		}
 	}
 
@@ -1867,15 +1880,15 @@ func (c *Config) persistentPreRunRootE(cmd *cobra.Command, args []string) error 
 
 	// Read the config file.
 	if annotations.hasTag(doesNotRequireValidConfig) {
-		if c.configFileAbsPathErr == nil {
+		if c.defaultConfigFileAbsPathErr == nil {
 			_ = c.readConfig()
 		}
 	} else {
-		if c.configFileAbsPathErr != nil {
-			return c.configFileAbsPathErr
+		if c.defaultConfigFileAbsPathErr != nil {
+			return c.defaultConfigFileAbsPathErr
 		}
 		if err := c.readConfig(); err != nil {
-			return fmt.Errorf("invalid config: %s: %w", c.configFileAbsPath, err)
+			return fmt.Errorf("invalid config: %s: %w", c.getConfigFileAbsPath(), err)
 		}
 	}
 
@@ -2030,7 +2043,7 @@ func (c *Config) persistentPreRunRootE(cmd *cobra.Command, args []string) error 
 
 	// Create the config directory if needed.
 	if annotations.hasTag(requiresConfigDirectory) {
-		if err := chezmoi.MkdirAll(c.baseSystem, c.configFileAbsPath.Dir(), fs.ModePerm); err != nil {
+		if err := chezmoi.MkdirAll(c.baseSystem, c.getConfigFileAbsPath().Dir(), fs.ModePerm); err != nil {
 			return err
 		}
 	}
@@ -2155,8 +2168,8 @@ func (c *Config) persistentStateFile() (chezmoi.AbsPath, error) {
 	if !c.PersistentStateAbsPath.Empty() {
 		return c.PersistentStateAbsPath, nil
 	}
-	if !c.configFileAbsPath.Empty() {
-		return c.configFileAbsPath.Dir().Join(persistentStateFileRelPath), nil
+	if !c.getConfigFileAbsPath().Empty() {
+		return c.getConfigFileAbsPath().Dir().Join(persistentStateFileRelPath), nil
 	}
 	for _, configDir := range c.bds.ConfigDirs {
 		configDirAbsPath, err := chezmoi.NewAbsPathFromExtPath(configDir, c.homeDirAbsPath)
@@ -2279,7 +2292,7 @@ func (c *Config) newTemplateData(cmd *cobra.Command) *templateData {
 		command:           cmd.Name(),
 		commandDir:        c.commandDirAbsPath,
 		config:            c.ConfigFile.toMap(),
-		configFile:        c.configFileAbsPath,
+		configFile:        c.getConfigFileAbsPath(),
 		executable:        chezmoi.NewAbsPath(executable),
 		fqdnHostname:      fqdnHostname,
 		gid:               gid,
@@ -2307,7 +2320,7 @@ func (c *Config) newTemplateData(cmd *cobra.Command) *templateData {
 
 // readConfig reads the config file, if it exists.
 func (c *Config) readConfig() error {
-	switch err := c.decodeConfigFile(c.configFileAbsPath, &c.ConfigFile); {
+	switch err := c.decodeConfigFile(c.getConfigFileAbsPath(), &c.ConfigFile); {
 	case errors.Is(err, fs.ErrNotExist):
 		return nil
 	default:
--- a/internal/cmd/doctorcmd.go
+++ b/internal/cmd/doctorcmd.go
@@ -191,7 +191,7 @@ func (c *Config) runDoctorCmd(cmd *cobra.Command, args []string) error {
 		&configFileCheck{
 			basename: chezmoiRelPath,
 			bds:      c.bds,
-			expected: c.configFileAbsPath,
+			expected: c.getConfigFileAbsPath(),
 		},
 		&dirCheck{
 			name:    "source-dir",
--- a/internal/cmd/editconfigcmd.go
+++ b/internal/cmd/editconfigcmd.go
@@ -24,5 +24,5 @@ func (c *Config) newEditConfigCmd() *cobra.Command {
 }
 
 func (c *Config) runEditConfigCmd(cmd *cobra.Command, args []string) error {
-	return c.runEditor([]string{c.configFileAbsPath.String()})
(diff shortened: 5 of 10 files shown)


paltherr:
Arguably an error check similar to the newly introduced one could also be done in case there is no `--config `option. This could help in cases where someone switches from one format of `.chezmoi.<format>.tmpl` to another one. Currently, if you do that, `chezmoi status` signals that the config changed and that `init` should be rerun. Rerunning `chezmoi init`, signals NO error and generates a new config file with the new format but without deleting the old one. Thus, the next time you run `chezmoi`, it fails with an error stating that there are multiple config files. With the new error check, `chezmoi init` could signal that there is a conflict.

Btw, another benefit of this PR would be that the `init` option `--config-path` could be deprecated (I think). I remember that when I started using `chezmoi` I found it very confusing that there are two options, `--config` and `--config-path`, seemingly for the same thing. It took me some time to figure out when and how to use each one. I think that it would be a great plus if future users would only have to learn about the former.

paltherr:
The PR still needs work (see points below) but I need your input. First of all to know whether you think this can be brought into a state where the added value is worth the extra complexity/ugliness. And, if yes, what your recommendations are to get there. I'm not very familiar with github and pull requests but I guess that I have to mark the PR as "Ready for review" to get your attention. I will do that after this comment.

Here are points that still need work/answers:
- The new error check shouldn't compare file base names but only file extensions.
- Should the new error check also be done when no `--config` option was given (see previous comment)? If yes,
  - In case of a conflict, should an error be reported? An alternative would be to delete the conflicting config file, i.e., changing the format of `.chemoi.<format>.tmpl` would just work.
  - Should this be done in a separate PR?
- Better names are needed for the Config fields.
  - The fields `configFileAbsPathDefault` and `configFileAbsPathErr` could be renamed to `defaultConfigFileAbsPath` and `defaultConfigFileAbsPath` to match the function name `defaultConfigFile` that computes them.
  - The field `configFileAbsPathExplicit` could be renamed to one of `optionConfigFileAbsPath`, `flagConfigFileAbsPath`, or `explicitConfigFileAbsPath`.
- Should the new function `configFileAbsPath()` be renamed and/or moved?
- A number of scenarios/journeys should be tested. Should I try to squeeze everything into an `issue3127.txtar` or could the tests be split over multiple files?

twpayne:
Firstly, thank you for your work and this excellent PR! To answer some of your questions:

> First of all to know whether you think this can be brought into a state where the added value is worth the extra complexity/ugliness.

Yes, absolutely! Users care about the interface, not the implementation, and this is a key usability improvement for people using multiple chezmoi configs. It's definitely worth the extra complexity, the ugliness is not visible, and as long as we have tests to check the user-facing behavior for common cases we can freely change the implementation to be more beautiful in the future.

> Should the new error check also be done when no `--config` option was given (see previous comment)?

Yes, but this falls foul of the [complexities I described](https://github.com/twpayne/chezmoi/issues/3127#issuecomment-1892948048). Specifically, how do distinguish between "no `--config` option was given" and "a `--config` option was given but was set to the default value". Elsewhere in the codebase I've used a [sentinel value to handle this](https://github.com/twpayne/chezmoi/blob/80717f4f1c35862d512e8b5415058cdc5cdcdc97/internal/cmd/config.go#L61), but I don't know if a similar approach could be used here.

> In case of a conflict, should an error be reported?

Yes. chezmoi's users are highly technical so by default report meaningful, actionable errors when anything is inconsistent.

> Better names are needed for the Config fields.

Yup, the names you propose are better.

> Should the new function `configFileAbsPath()` be renamed and/or moved?

Yes, but this is an internal detail. I would suggest `getConfigFileAbsPath` to make the distinction between a value (e.g. `configFileAbsPath`) and a potentially-expensive function that returns a value (`getConfigFileAbsPath()`).

chezmoi does try to keep functions in some kind of mix of hierarchical and alphabetical order, so no need to move the function right now. At some point, I'll write a linter to check function order, but, for this PR, meaningful diffs are more important than function order so there is no need to move functions.

> * A number of scenarios/journeys should be tested. Should I try to squeeze everything into an `issue3127.txtar` or could the tests be split over multiple files?

Please split the tests over multiple files. This both separates intent and allows the tests to run in less wallclock time (`go test` can run them in parallel). Specifically:

* For changes to behavior that should be obvious defaults - like this, where removing the `--config-path` hack is a clear improvement, please update `internal/cmd/testdata/scripts/init.txtar` with the new behavior.
* For exhaustive testing of the changes, please create a new `initconfig.txtar` script with as many tests as you can think of to this script, especially edge cases.
* For specific issues, like the original report in #3127, please do create an `issue3127.txtar` that tests the reported case directly. These are extremely helpful as high-level behavioral checks.
* Finally, if there are edge cases that you can think of that the current code doesn't pass, please include them anyway, but commented out or prefixed with `[skip]` so they don't run. Such cases are extremely helpful to future developers, and it's OK that the code doesn't pass them now.

Thanks again for this excellent PR!

paltherr:
>
> Yes, absolutely! Users care about the interface, not the implementation,
> and this is a key usability improvement for people using multiple chezmoi
> configs. It's definitely worth the extra complexity, the ugliness is not
> visible, and as long as we have tests to check the user-facing behavior for
> common cases we can freely change the implementation to be more beautiful
> in the future.


Great to hear that :-) I'll first clean up the code and then write the
tests. The former should be relatively quick, the latter probably less so
as I'll first have to understand how txtar files work. I found testscript
<https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript>, which
describes them. Is there anything else I should know about tests?

*Before I proceed, I have one question though: how am I supposed to
contribute the updates?* This is my first PR on github and it's not
entirely clear to me how I should go forward. Here is my current setup:

- I cloned your repo on github.
- I cloned my clone on my local machine.
- I created a commit on my local machine and pushed it to my clone on
github.
- On github I used the commit to create the PR.

My current understanding of how I should go forward is the following:

- Create one or more new commits on my local machine.
- Push the commits to my clone on github. They will auto-magically be added
to the PR.
- Rinse and repeat until we agree on a final state.
- You integrate the PR into your repo.

If this is indeed the correct way, how should I name and describe my
commits? Should I use "chore:", "fix:", or some other prefix? How important
are the descriptions? Are they just for you to understand what I'm
changing/adding to the PR or will they survive the PR and should therefore
also make sense for future readers of your repo? As you can see, one thing
that isn't clear to me is whether the extra commits will survive the PR or
whether they will all be merged into a single commit on your repo.

Yes, but this falls foul of the complexities I described
> <https://github.com/twpayne/chezmoi/issues/3127#issuecomment-1892948048>.
> Specifically, how do distinguish between "no --config option was given"
> and "a --config option was given but was set to the default value".


That should now be possible. That's why I split the old configFileAbsPath field
into two fields, configFileAbsPathDefault and configFileAbsPathExplicit,
and one function configFileAbsPath(). The former field only ever contains
the default value computed by defaultConfigFile(), the latter field only
ever contains the value of the  --config option, or Empty, if none was
given. The function combines the two fields and returns the value that was
previously stored in the configFileAbsPath field.

This assumes that the --config option is the only way to explicitly specify
the config file (to load). My understanding is that the XDG specification
only allows to specify where (in which directories) to look for a config
file but not the config file itself. Did I overlook something?

twpayne:
Firstly, there are answers to most of your questions in [the contributing docs](https://www.chezmoi.io/developer/contributing-changes/). You also see this link when you open a PR.

> Is there anything else I should know about tests?

Please follow the style of the existing tests. Note that you can't easily use a debugger to step through testscript tests, so you might have to do some printf-style debugging. Useful commands for this include:

```
exec find $WORK # find everything in the filesystem

exec chezmoi ...
! stdout . # print whatever the previous command wrote stdout and stop
! stderr . # print whatever the previous command wrote stderr and stop
```

> This assumes that the --config option is the only way to explicitly specify the config file (to load). My understanding is that the XDG specification only allows to specify where (in which directories) to look for a config file but not the config file itself. Did I overlook something?

I think you are correct. It will be great to verify edge cases with tests.

paltherr:
tl;dr: Rather than extend the new error check to more cases I will drop it. The case that it signals is a corner case that isn't much different from similar already existing corner cases that aren't signaled either. I still think that it would make sense to signal these cases but doing so turns out to be complex enough to warrant its own issue and pull request.

### Long Version

The new error check that I added in my first version of the code tries to ensure that the updated config is written into a file of the right format. More precisely, it tries to enforce that if the config template is of format _FORMAT_, then the config is written into a file with a `.FORMAT` extension. The check tests the whole base name of the target file and thus enforces that it's equal to `chezmoi.FORMAT`. That's too strict because the `--config` option allows to use config files with arbitrary names. When I looked into restricting the check to the extension of the file, I figured that even the extension could be arbitrarily chosen, since the format of the config file can be independently specified with the `--config-format` option. Thus, just writing a check that is neither too strict nor too lax is already non-trivial.

The check tries to prevent chezmoi from corrupting its own config. The `init` command unfortunately has that potential. Running it can prevent further chezmoi commands of working, even with a perfectly valid config template. One scenario where this can happen is if the user replaces their config template with a new one, possibly an equivalent one, but in a different format. After such a change, chezmoi prompts the user to rerun the `init` command. Assuming that the config file is in the default location, running `chezmoi init` creates a new config file in the new format in the default location. From then on, all chezmoi commands, including the `init` command, exit with an error stating that there are multiple config files. The fix is easy, delete the config file in the old format, but it may not be so obvious for users who didn't realize that the change of format of the config template wasn't completely innocuous. Therefore detecting and signaling this case  may be worthwhile.

For users that rely on a config file in a custom location, things are a bit different. They have to run all their chezmoi commands with the `--config PATH` option and whenever they run the `init` command they have to run it with the `--config-path PATH` option. If they do that after replacing their config template with a new one in a different format, then all following chezmoi commands fail because they can't parse the updated config file. Here again the fix is easy, rename the updated file. Alternatively the `init` command could have been run with a `--config-path` option with the new file name. In any case, the user has to update their commands/aliases/scripts to use a `--config` option with the new file name. All this may well be obvious to users advanced enough to use the `--config` option but a little warning would certainly not hurt either.

With the change proposed in this PR and assuming that the init option `--config-path` is eventually deprecated, users using a custom config location would be compelled to run `chezmoi -config PATH init` and then rename the updated config file. This illustrates that while detecting and reporting the case might be desirable, turning it into an inescapable error is probably not advisable, as it may make it much more difficult to update the config file in that scenario.

Fortunately the case looks very much like a corner case. The scenario described above leading to it, while possible, is certainly not one happening very often. I'm not aware of other scenarios leading to the same case. I can't exclude that others exist but they too most likely occur only rarely. Thus, there is no urgency in detecting and signaling this case. Furthermore, the problem exists already and the change in the PR doesn't make it significantly worse, the logic to detect the case isn't completely trivial, and finally even how to signal it isn't necessarily completely obvious. For all these reasons, it looks more appropriate to leave the detection and signaling of this case for a separate PR with its own issue.

twpayne:
This PR is looking great. Please can you add some tests. The right place to add them is probably `internal/cmd/testdata/scripts/init.txtar`, although you can also create a new separate testscript for this functionality.

Thanks for the detailed explanations. It's certainly a very messy area with multiple interacting, possibly conflicting, sources of truth (e.g. multiple XDG config directories, `--config` flag, `--config--format` flag, chezmoi's internal defaults).

On the up side, people changing config file formats is a relatively rare thing (although it does happen, see #1430), so I'm OK if chezmoi does not handle all edge cases correctly.

paltherr:
I'm working on the tests. I was pondering on what exactly to include and also how to express it with testscript. They should soon be ready.

twpayne:
Please don't hesitate to ask questions or brainstorm about tests. Generally speaking, the more the better, and it doesn't matter if they're ugly. We can always tidy them up later.

twpayne (commented):
Looks excellent.

Please follow the existing style of testscripts, specifically with respect to removing empty lines between comments and the commands they refer to as they have a meaning to testscript.

paltherr:
The latest error is a bit weird. First I thought that it was because I should have used $/ instead of / but after looking more closely, it looks like that chezmoi prints UNIX-like paths (with /) even when it runs on Windows and the problem is that $CHEZMOICONFIGDIR, $HOME, and $WORK contain Windows-like paths (with \). Is there a way to rewrite variables when they are expanded by testscript?

twpayne:
Sadly it's really tricky to get tests that test paths to pass on Windows. chezmoi uses forward slashes internally on all platforms, but testscript uses the native path separator, i.e. backward slashes on Windows. When you mix chezmoi paths with testscript paths then you get a mix of the two.

In the short term, please add

    [windows] skip 'test requires path separator to be forward slash'

as the first line of the failing testscript tests. I hope to clean this up sometime in the future.

Once that's done, please also squash all commits into a single one using this PR's title as the commit message, and then I think this PR is good to merge :) Thank you for your patience on this one.

twpayne:
This is a fantastic PR that fixes a long-standing bug in chezmoi. Thank you so much!

PR #3494, opened 2024-01-17
feat: Add re-add --recursive flag and make it the default

### LINKED ISSUE #3490: feat: `re-add <dir>` descends into directories

Currently, `chezmoi re-add <dir>` ignores the directory `<dir>`. For example, `chezmoi re-add .` would simply exit. A working `chezmoi re-add .` would fit nicely with the mental model of `git add .`.

It would be great if chezmoi descends into `<dir>` and re-adds the tracked files in that directory. 

- Just re-add the files individually (but I am lazy to type their paths)
- Or `chezmoi re-add --interactive` and select wanted files

## Additional context

I did a shallow dive into the codebase and found that slight modifications of the logic here might suffice:
https://github.com/twpayne/chezmoi/blob/85983f06931e86b157a877287a82d8b6faca8eea/internal/cmd/readdcmd.go#L43-L71

But I don't know golang at all, so it would be a bit difficult for me to submit a PR.

twpayne:
Fixes #3490.

@bryango would you be able to test this? You'll find pre-built chezmoi binaries in the artifacts section of the checks run by GitHub Actions on this PR. Sorry, there's no way to provide a direct link.

Technically, this is a backwards-incompatible change because the default behavior of `chezmoi re-add` changes from not-recursive to recursive. However, I think this is OK because the use case that it breaks is using `chezmoi re-add` on a directory where you want re-add the directory itself but not its contents. I suspect that this use case is sufficiently rare that it's OK to make this change with only a patch version bump, especially as `chezmoi re-add` is likely used interactively rather than scripted.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/re-add.md
+++ b/assets/chezmoi.io/docs/reference/commands/re-add.md
@@ -2,10 +2,14 @@
 
 Re-add modified files in the target state, preserving any `encrypted_`
 attributes. chezmoi will not overwrite templates, and all entries that are not
-files are ignored.
+files are ignored. Directories are recursed into by default.
 
 If no *target*s are specified then all modified files are re-added. If one or
 more *target*s are given then only those targets are re-added.
+
+## `-r`, `--recursive`
+
+Recursively add files in subdirectories.
 
 !!! hint
 
@@ -16,4 +20,5 @@ more *target*s are given then only those targets are re-added.
     ```console
     $ chezmoi re-add
     $ chezmoi re-add ~/.bashrc
+    $ chezmoi re-add --recursive=false ~/.config/git
     ```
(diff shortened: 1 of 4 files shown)


bryango (approved):
Wow, that's fast! Thank you! Tested with the `chezmoi-linux-amd64-musl.zip` artifact. It works!

assets/chezmoi.io/docs/reference/commands/re-add.md
bryango, line 1: Nit: should we add `Directories are recursed by default.` here?

https://github.com/twpayne/chezmoi/pull/3494/files#diff-b956ad1a9a5bf24d3617ed00f5faa29d008fb45155bc5024a3b8df63d0b4fc0bR3-R5

Or maybe it's okay since it's mentioned later in the doc?

halostatue (approved):

assets/chezmoi.io/docs/reference/commands/re-add.md
twpayne: Good idea. Done.

PR #3505, opened 2024-01-23
docs: combine testing and debugging templates

dpprdan:
This combines the two similar sections "testing templates" and "debugging templates".

I've also taken the liberty to link to the section where (I found it) appropriate, but feel free to revert that commit if that isn't your style.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/templating.md
+++ b/assets/chezmoi.io/docs/user-guide/templating.md
@@ -93,20 +93,28 @@ $ chezmoi edit --apply ~/.zshrc
 
 ## Testing templates
 
-Templates can be tested with the `chezmoi execute-template` command which treats
-each of its arguments as a template and executes it. This can be useful for
-testing small fragments of templates, for example:
+Templates can be tested and debugged with `chezmoi execute-template`,
+which treats each of its arguments as a template and executes it.
+The templates are interpreted and the results are output to standard
+output, making it useful for testing small template fragments:
 
 ```console
 $ chezmoi execute-template '{{ .chezmoi.hostname }}'
 ```
 
-If there are no arguments, `chezmoi execute-template` will read the template
-from the standard input. This can be useful for testing whole files, for example:
+Without arguments, `chezmoi execute-template` will read the template from
+standard input, which is useful for testing whole files:
 
 ```console
 $ chezmoi cd
 $ chezmoi execute-template < dot_zshrc.tmpl
+```
+
+If file redirection does not work (as when using PowerShell), the contents
+of a file can be piped into `chezmoi execute-template`:
+
+```console
+$ cat foo.txt | chezmoi execute-template
 ```
 
 ## Template syntax
@@ -159,29 +167,6 @@ HOSTNAME=myhostname
 
 Notice that this will remove any number of tabs, spaces and even newlines and
 carriage returns.
-
-## Debugging templates
-
-If there is a mistake in one of your templates and you want to debug it, chezmoi
-can help you. You can use this subcommand to test and play with the examples in
-these docs as well.
-
-There is a very handy subcommand called `execute-template`. chezmoi will
-interpret any data coming from stdin or at the end of the command. It will then
-interpret all templates and output the result to stdout. For example with the
-command:
-
-```console
-$ chezmoi execute-template '{{ .chezmoi.os }}/{{ .chezmoi.arch }}'
-```
-
-chezmoi will output the current OS and architecture to stdout.
-
-You can also feed the contents of a file to this command by typing:
-
-```console
-$ cat foo.txt | chezmoi execute-template
-```
 
 ## Simple logic
 
--- a/assets/chezmoi.io/docs/reference/commands/execute-template.md
+++ b/assets/chezmoi.io/docs/reference/commands/execute-template.md
@@ -1,6 +1,7 @@
 # `execute-template` [*template*...]
 
-Execute *template*s. This is useful for testing templates or for calling
+Execute *template*s. This is useful for [testing
+templates](../../user-guide/templating.md#testing-templates) or for calling
 chezmoi from other scripts. *templates* are interpreted as literal templates,
 with no whitespace added to the output between arguments. If no templates are
 specified, the template is read from stdin.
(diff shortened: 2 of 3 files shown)


assets/chezmoi.io/docs/user-guide/templating.md
```diff
@@ -93,22 +93,33 @@ $ chezmoi edit --apply ~/.zshrc
...
 
-Templates can be tested with the `chezmoi execute-template` command which treats
-each of its arguments as a template and executes it. This can be useful for
-testing small fragments of templates, for example:
+Templates can be tested and debugged with the `chezmoi execute-template`
+command which treats each of its arguments as a template and executes it.
+It will then interpret the templates and output the result to standard
+output.
+
+This can be useful for testing small fragments of templates, for example:
```
halostatue, line 101: I would make this one paragraph:

```markdown
Templates can be tested and debugged with `chezmoi execute-template`, which treats each of its arguments as a template and executes it. The templates are interpreted and the results are output to standard output, making it useful for testing small template fragments:
```

assets/chezmoi.io/docs/user-guide/templating.md
```diff
@@ -93,22 +93,33 @@ $ chezmoi edit --apply ~/.zshrc
...
 
 ```console
 $ chezmoi execute-template '{{ .chezmoi.hostname }}'
 ```
 
 If there are no arguments, `chezmoi execute-template` will read the template
-from the standard input. This can be useful for testing whole files, for example:
+from the standard input. This can be useful for testing whole files,
+for example:
```
halostatue, line 109:
```markdown
Without arguments, `chezmoi execute-template` will read the template from standard input,
which is useful for testing whole files:
```

assets/chezmoi.io/docs/user-guide/templating.md
```diff
@@ -93,22 +93,33 @@ $ chezmoi edit --apply ~/.zshrc
...
 
 ```console
 $ chezmoi cd
 $ chezmoi execute-template < dot_zshrc.tmpl
 ```
 
+If this does not work (e.g. with Powershell), you can pipe the contents of
```
halostatue, line 116:
````markdown
If file redirection does not work (as when using PowerShell), the contents of a file can be piped
into `chezmoi execute-template`:

```console
$ cat foo.txt | chezmoi execute-template
```
````

**Review by halostatue — APPROVED** [review]
These changes make sense; I have a couple of suggestions to reword some of the text to be less verbose.

**Comment by dpprdan** [comment]
@halostatue Thanks, I applied all your suggestions.

File: assets/chezmoi.io/docs/user-guide/templating.md [hunk 4]
```diff
@@ -93,22 +93,33 @@ $ chezmoi edit --apply ~/.zshrc
...
 
-Templates can be tested with the `chezmoi execute-template` command which treats
-each of its arguments as a template and executes it. This can be useful for
-testing small fragments of templates, for example:
+Templates can be tested and debugged with `chezmoi execute-template`,
+which treats each of its arguments as a template and executes it.
+The templates are interpreted and the results are output to standard
+output, making it useful for testing small template fragments:
+
+This can be useful for testing small fragments of templates, for example:
```
halostatue, line 101: Line 101 should have been removed.

assets/chezmoi.io/docs/user-guide/templating.md
halostatue: This change was not made. The main thing that I care about is `e.g. with PowerShell`; the canonical formatting is `PowerShell` not `Powershell` and `e.g.` is often overused.

Thinking more about this paragraph, I think it would be better (since PowerShell is being used as the example) to use a PowerShell example:

```powershell
Get-Content -Path input.txt -Raw | chezmoi execute-template
```

I’m sure that @bradenhilton will correct me if I got that invocation wrong.

assets/chezmoi.io/docs/user-guide/templating.md
halostatue: I’d like to eliminate the `for example`:, which is part of why I suggested the rewording.

halostatue (changes requested):
Nearly ready, but a couple of text suggestions still, and it looks like some changes I suggested didn't get applied cleanly.

With respect to your commits, I believe that this can all be squashed into a single commit with the PR title as the commit title. I don't think we need a separate commit for adding the links, but I would add it as a comment to the single commit.

dpprdan:
Oh no, how did that happen? I hope it's working now.

assets/chezmoi.io/docs/user-guide/templating.md
dpprdan: > Thinking more about this paragraph, I think it would be better (since PowerShell is being used as the example) to use a PowerShell example:

I don't agree, I am afraid. `cat` works fine in PowerShell (it is [defined as an alias](https://learn.microsoft.com/en-us/powershell/scripting/learn/shell/using-aliases?view=powershell-5.1#compatibility-aliases-in-windows)), but is not limited to it like `Get-Content`.

halostatue: If `cat` works, then that's fine and it does not need changing. I don't use PowerShell, so I was unaware of that default alias.

halostatue:
OK. Please squash these down into one or two commits (commit 3 and 4 should be reordered and squashed into commit 1 if you keep two commits) and we can get this merged. It's a good change.

dpprdan:
> Please squash these down into one or two commits

Oh, don't you have the option to squash and merge the PR? That would be easiest, I think.

image

halostatue:
No; the repository is configured only to accept rebase merges.

twpayne:
> Oh, don't you have the option to squash and merge the PR? That would be easiest, I think. Else I'll squash it tomorrow.

No, because chezmoi generates its user-visible release notes from the individual commit messages, and squashing and merging would collapse multiple user-visible changes in a single PR into a single, less meaningful, commit.

dpprdan:
Understood. 
I think we're good to go.

halostatue:
Thanks!

> What do you use to format markdown?

I *personally* use prettier to format my documents (it works for enough cases that it doesn't get in my way), but the Chezmoi documents are written in a way where I have to remember to turn off my auto-formatting when editing the docs (as an example, I prefer to use `_emphasize_` and `**strong**`, but chezmoi docs use `*emphasize*` and I don't think uses `**strong**`).

If you meant rendering, it's all done with mkdocs material.

halostatue (approved):

PR #3509, opened 2024-01-24
feat: Support rage as an alternative age encryption command

### LINKED ISSUE #3503: Support arbitrary encryption/decryption tools

I run chezmoi mostly-non-interactively with [darkman](https://gitlab.com/WhyNotHugo/darkman/) as part of an automated theme switching workflow. It worked wonderfully with GPG because there's a GUI prompt for the GPG passphrase; the script kicks off in the background and sends me a notification, so when the prompt appears a second later, I know what it's for, enter the PIN, and the rest executes without error. Since switching to age, however, it's much more painful because neither upstream age nor age bundled with chezmoi open that GUI prompt, so the script fails. I have to interactively switch themes from inside a shell because decryption immediately fails when there's no console.

I also can't use the age version bundled with chezmoi because it doesn't integrate with [age-plugin-yubikey](https://github.com/str4d/age-plugin-yubikey/) at all.

I think the _best_ solution would be allowing users to pick whatever encryption/decryption system they prefer; they could just specify the two commands for encryption/decryption with a `$FILEPATH` variable or something that chezmoi fills in before execution.

Alternatively, adding support for just [rage](https://str4d.xyz/rage), which [does have a GUI prompt](https://github.com/str4d/age-plugin-yubikey/issues/154#issuecomment-1904788310) through `pinentry`, would solve my problem.

twpayne:
Fixes #3503.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,12 +13,13 @@ env:
   AGE_VERSION: 1.1.1
   CHOCOLATEY_VERSION: 2.2.2
   EDITORCONFIG_CHECKER_VERSION: 2.7.2
+  FIND_TYPOS_VERSION: 0.0.3
   GO_VERSION: 1.21.6
   GOFUMPT_VERSION: 0.5.0
   GOLANGCI_LINT_VERSION: 1.55.2
   GOLINES_VERSION: 0.11.0
   GOVERSIONINFO_VERSION: 1.4.0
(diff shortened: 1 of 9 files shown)


halostatue (approved):
This looks good.

I wonder if it might not be possible to *autodetect* `rage` or `age` since they are compatible to the command-line level. (The detection order would be `age`, `rage`, fall back to `age` anyway.)

Issue #2814, opened 2023-03-03
Run modify scripts in parallel


VorpalBlade:
## Is your feature request related to a problem? Please describe.

I have a few modify scripts (to deal with KDE):
```console
$ find . -name 'modify_*' | wc -l                           
15
```
and I have noticed the performance of operations like `chezmoi diff` seem to degrade linearly with the number of scripts.

I used strace to figure out what was going on, and to me it looks like chezmoi waits for each script to run before starting the next one. There is no reason modify scripts could not be run in parallel (as they do not affect each other). Even if the scripts are completely CPU bound this would be a big speedup (8x - 16x in my case, depending on the computer).  In practise my scripts perform a mix of IO and CPU, so the speedup could be even bigger (or smaller).

## Describe the solution you'd like

I propose that modify scripts should be executed in parallel by chezmoi. That would interfer with the proposed solution of using `/dev/tty` in #2244 (which I have not yet implemented) but that is a non-standard thing to do and could be solved with the scripts themselves using locking between themselves when they rarely need to interact with the terminal.

## Describe alternatives you've considered

I considered rewriting my python script in Rust. This would undoubtedly speed it up, and I might end up doing this in the end anyway. It is however harder to distribute (need to have x86, x86-64, ARM32 and ARM64 versions for just my own personal needs for example).

My script does not however allow much internal parallelisation, so it would still not be able to make efficient use of all cores. Running multiple modify scripts in parallel would still be beneficial regardless.

## Additional context

It would be interesting to think about if there are any other areas of chezmoi that would benefit from parallelisation as well, but I feel that is the topic of another issue.

twpayne:
This is similar to #2740.

chezmoi runs scripts sequentially in a strict order for transparency and repeatability. Writing scripts that perform correctly when run concurrently is extremely difficult, and running them concurrently is not repeatable.

You can run things concurrently in a single script, for an example see https://github.com/twpayne/chezmoi/issues/2740#issuecomment-1416929835.

VorpalBlade:
> This is similar to #2740.
> 
> chezmoi runs scripts sequentially in a strict order for transparency and repeatability. Writing scripts that perform correctly when run concurrently is extremely difficult, and running them concurrently is not repeatable.

[...]

Issue #2670, opened 2023-01-02
Download externals concurrently


twpayne:
## Is your feature request related to a problem? Please describe.

chezmoi downloads externals serially. Downloading externals takes time, and this can lead to long pauses in chezmoi's progress (#2369). It would be nice if this overall process could be faster.

## Describe the solution you'd like

chezmoi should download externals concurrently, subject to some externally-specified sensible limit (e.g. number of simultaneous concurrent downloads).

## Describe alternatives you've considered

Keeping things as they are.

## Additional context

Programs like [curl](https://curl.se/) and [aria2](https://aria2.github.io/) are much more advanced. Maybe, as well as enabling concurrent downloads, chezmoi could include a download plugin option.

halostatue:
Possible with go channels?

PR #3518, opened 2024-01-27
feat: Set CHEZMOI_SOURCE_FILE env var for scripts

### LINKED ISSUE #2934: modify script executed from weird path (even when modify script is not templated)

I was trying to be smart and inspect the path to the modify script (in order to detect an associated data file located relative to it). My modify script is actually a config file being executed by a rust program:

```bash
#!/path/to/rust/program

some directives here;
ignore blah;
transform quux;
```

The idea was that the rust program would inspect the path to this `modify_` "script" to locate additional data files. However, when chezmoi actually executes this script the path is something like: `/tmp/3306499162.Trolltech.conf`.

This would make sense if the modify script was templated (`modify_blah.tmpl`) as chezmoi would need to replace things in the file. But that is not the case here, it is NOT a template. How do I get the original path to this file? Do I simply have to accept that I can't, and make it a `.tmpl` and insert something like: `source "{{ .chezmoi.sourceDir }}/{{ .chezmoi.sourceFile}}"`?

Either way, I can't find where this behaviour is documented.

Scratching my head at this (it is late local time...).

* [X] I have read [chezmoi's user guide](https://chezmoi.io/user-guide/command-overview/), and not found the answer.
* [X] I have searched [chezmoi's reference guide](https://chezmoi.io/reference/), and not found the answer.

## Output of any commands you've tried with `--verbose` flag

Irrelevant without the whole setup of my rust program. But if you really want to try it I can setup a test case.

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
warning   version              v2.33.1, built at 2023-04-08T07:27:16Z
ok        latest-version       v2.33.1
ok        os-arch              linux/amd64 (Arch Linux)
ok        uname                Linux theseus 6.2.11-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 13 Apr 2023 16:59:24 +0000 x86_64 GNU/Linux
ok        go-version           go1.20.3 (gc)
ok        executable           /usr/bin/chezmoi
ok        config-file          ~/.config/chezmoi/chezmoi.yaml, last modified 2023-04-05T12:45:46+02:00
warning   source-dir           ~/.local/share/chezmoi is a git working tree (dirty)
warning   suspicious-entries   ~/.local/share/chezmoi/.chezmoi_modify_manager.add_hook
warning   working-tree         ~/.local/share/chezmoi is a git working tree (dirty)
ok        dest-dir             ~ is a directory
ok        umask                022
ok        cd-command           found /bin/zsh
ok        cd-args              /bin/zsh
ok        edit-command         found /usr/bin/code
ok        edit-args            code -w
ok        git-command          found /usr/bin/git, version 2.40.0
ok        merge-command        found /usr/bin/bcompare
ok        shell-command        found /bin/zsh
ok        shell-args           /bin/zsh
ok        gpg-command          found /usr/bin/gpg, version 2.2.41
info      dashlane-command     dcli not found in $PATH
ok        keepassxc-command    found /usr/bin/keepassxc-cli, version 2.7.4
info      keeper-command       keeper not found in $PATH
info      passhole-command     ph not found in $PATH
info      rbw-command          rbw not found in $PATH
```

</details>

twpayne:
Fixes #2934. It only took nine months.

@VorpalBlade this sets the `$CHEZMOI_SOURCE_FILE` environment variable for `modify_` scripts (and also normal scripts), which should be useful for your excellent [chezmoi_modify_manager](https://github.com/VorpalBlade/chezmoi_modify_manager).

As background info, this change became possible after 7203d6bfc (in chezmoi v2.40.3).

Would you be able to test this?

### DIFF
--- a/internal/chezmoi/realsystem.go
+++ b/internal/chezmoi/realsystem.go
@@ -123,6 +123,9 @@ func (s *RealSystem) RunScript(scriptname RelPath, dir AbsPath, data []byte, opt
 	if err != nil {
 		return err
 	}
+	cmd.Env = append(os.Environ(),
+		"CHEZMOI_SOURCE_FILE="+options.SourceRelPath.String(),
+	)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -1855,6 +1855,9 @@ func (s *SourceState) newModifyTargetStateEntryFunc(
 
 			// Run the modifier on the current contents.
 			cmd := interpreter.ExecCommand(tempFile.Name())
+			cmd.Env = append(os.Environ(),
+				"CHEZMOI_SOURCE_FILE="+sourceRelPath.String(),
+			)
 			cmd.Stdin = bytes.NewReader(currentContents)
 			cmd.Stderr = os.Stderr
 			contents, err = chezmoilog.LogCmdOutput(cmd)
@@ -1911,6 +1914,7 @@ func (s *SourceState) newScriptTargetStateEntryFunc(
 			sourceAttr: SourceAttr{
 				Condition: fileAttr.Condition,
 			},
+			sourceRelPath: sourceRelPath,
 		}, nil
 	}
 }
--- a/internal/chezmoi/sourcestate_test.go
+++ b/internal/chezmoi/sourcestate_test.go
@@ -1016,6 +1016,7 @@ func TestSourceStateRead(t *testing.T) {
 							sourceAttr: SourceAttr{
 								Condition: ScriptConditionAlways,
 							},
+							sourceRelPath: NewSourceRelPath("run_script"),
 						},
 					},
 				}),
@@ -1046,6 +1047,7 @@ func TestSourceStateRead(t *testing.T) {
 							sourceAttr: SourceAttr{
 								Condition: ScriptConditionAlways,
 							},
+							sourceRelPath: NewSourceRelPath("run_script"),
 						},
 					},
 				}),
--- a/internal/chezmoi/system.go
+++ b/internal/chezmoi/system.go
@@ -14,8 +14,9 @@ import (
 )
 
 type RunScriptOptions struct {
-	Interpreter *Interpreter
-	Condition   ScriptCondition
+	Interpreter   *Interpreter
+	Condition     ScriptCondition
+	SourceRelPath SourceRelPath
 }
 
 // A System reads from and writes to a filesystem, runs scripts, and persists
--- a/internal/chezmoi/targetstateentry.go
+++ b/internal/chezmoi/targetstateentry.go
@@ -52,10 +52,11 @@ type TargetStateRemove struct{}
 // A TargetStateScript represents the state of a script.
 type TargetStateScript struct {
 	*lazyContents
-	name        RelPath
-	interpreter *Interpreter
-	condition   ScriptCondition
-	sourceAttr  SourceAttr
+	name          RelPath
+	interpreter   *Interpreter
+	condition     ScriptCondition
+	sourceAttr    SourceAttr
+	sourceRelPath SourceRelPath
 }
 
 // A TargetStateSymlink represents the state of a symlink in the target state.
@@ -333,8 +334,9 @@ func (t *TargetStateScript) Apply(
 	runAt := time.Now().UTC()
 	if !isEmpty(contents) {
 		if err := system.RunScript(t.name, actualStateEntry.Path().Dir(), contents, RunScriptOptions{
-			Condition:   t.condition,
-			Interpreter: t.interpreter,
+			Condition:     t.condition,
+			Interpreter:   t.interpreter,
+			SourceRelPath: t.sourceRelPath,
 		}); err != nil {
 			return false, err
 		}
--- a/internal/cmd/testdata/scripts/issue2934.txtar
+++ b/internal/cmd/testdata/scripts/issue2934.txtar
@@ -1,10 +1,24 @@
-[windows] skip
+[windows] skip 'UNIX only'
 
 # test that chezmoi sets environment variables for modify_ scripts
-exec chezmoi apply
-grep ^${CHEZMOISOURCEDIR@R}$ $HOME/.modify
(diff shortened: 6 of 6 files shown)


halostatue (approved):
Looks reasonable, but I have not tested it.

VorpalBlade:
I will try: I'm down with a nasty cold at the moment, so it may take a few days before I have energy to deal with this.

It will be interesting to see if it speeds up chezmoi diff (currently a no-op one takes about 85 ms on my desktop iirc, and I have run out of ways to speed it up on my side, a single chezmoi_modify_manager invocation is about 2-3 ms)

twpayne:
No rush at all and wishing you a speedy recovery!

VorpalBlade:
I actually got the energy to give this a try now. The mechanism in itself doesn't really offer any speed-up (as expected).

But this means that every `modify_` script that doesn't otherwise need templating can now be a non-templated one. *That* does offer a speedup. In my own chezmoi source state I could convert 16 (out of 20) modify scripts to no longer be templates.

The runtime of `./chezmoi diff` went from
```console
$ # With lookup based on env vars but still .tmpl files
$ hyperfine "./chezmoi diff"
Benchmark 1: ./chezmoi diff
  Time (mean ± σ):      87.2 ms ±   1.6 ms    [User: 72.5 ms, System: 32.4 ms]
  Range (min … max):    84.9 ms …  92.8 ms    33 runs

$ # After converting to non-.tmpl files
$ hyperfine "./chezmoi diff"
Benchmark 1: ./chezmoi diff
  Time (mean ± σ):      78.5 ms ±   0.8 ms    [User: 65.7 ms, System: 27.8 ms]
  Range (min … max):    76.9 ms …  80.2 ms    37 runs
```

That is ~90% of the original time, not bad.

(The difference between .tmpl with `source "{{ .chezmoi.sourceDir }}/{{ .chezmoi.sourceFile | trimSuffix ".tmpl" | replace "modify_" "" }}.src.ini"` and `source auto` while still a .tmpl file was not measurable let alone statistically significant.)

## Further improvement idea

Do you still always make a copy of the `modify_` scripts into /tmp? I bet avoiding that could save a bunch of time, though it would require setting `chmod +x` on the script in the source directory (this is fine, git tracks +x file attribute).

To support legacy scripts you would need to support copying if required though I guess. Might still be worth detecting the file modes via a `stat/lstat/fstat/fstatat()` syscall. (Or if you want to be fancy and not fetch more than you need I believe there are newer selective syscalls, including allowing you to get the data at directory traversal time) I *assume* that would be quicker than copying the file, but you would need to benchmark to be sure of course.

VorpalBlade:
(Another note, I found it slightly annoying that I need to combine CHEZMOI_SOURCE_DIR and CHEZMOI_SOURCE_FILE to get the path, but if you don't already have that combined on your side for other purposes, it is probably faster to do the string manipulation in the rust code on my side anyway.)

twpayne:
Thank you for testing!

> Do you still always make a copy of the `modify_` scripts into /tmp? I bet avoiding that could save a bunch of time, though it would require setting `chmod +x` on the script in the source directory (this is fine, git tracks +x file attribute).

Yes, the scripts are always written to a temporary directory before being executed. This is always necessary if the script is a template. I agree that if the `modify_` script is not a template and already has the executable attribute then this can be skipped.

Even though git preserves the executable attribute, chezmoi's source state is designed to only use regular files and directories. This is important to provide compatibility with all filesystems and version control systems. In theory, chezmoi should work even if the only way to transfer your dotfiles is on a FAT32-formatted USB stick.

Although milliseconds can be shaved by avoiding the write to a temporary filesystem, I think the real benefits for the user will come from #2814 and #2670, which should improve chezmoi's running time by multiple seconds.



> (Another note, I found it slightly annoying that I need to combine CHEZMOI_SOURCE_DIR and CHEZMOI_SOURCE_FILE to get the path. But if you don't already have that combined on your side for other purposes, it is probably faster to do the string manipulation in the rust code on my side anyway.)

Concatenating two strings should only take fractional nanoseconds.

Thank you again, and hope that you're on your way to recovery. I'll merge this PR.

VorpalBlade:
Two questions:

1. Which version number of chezmoi will the the minimum to support this?
2. Is the output of `chezmoi --version` stable? (I.e. will always contain a substring on the form `v[0-9]+\.[0-9]+\.[0-9]+`, where the version presumably follows semantic versioning?)

twpayne:
> 1. Which version number of chezmoi will the the minimum to support this?

The next one, which will be either 2.46.1 or 2.47.0 depending on what other features and bug fixes land in the next few weeks. However, you can also test for the presence of the `CHEZMOI_SOURCE_FILE` env var directly, which should be more robust.



> 2\. Is the output of `chezmoi --version` stable? (I.e. will always contain a substring on the form `v[0-9]+\.[0-9]+\.[0-9]+`, where the version presumably follows semantic versioning?)

Yes. To be extra robust, look for the first occurrence of this regexp in the output of `chezmoi --version`, which may contain multiple lines.

VorpalBlade:
> > 1. Which version number of chezmoi will the the minimum to support this?
> 
> The next one, which will be either 2.46.1 or 2.47.0 depending on what other features and bug fixes land in the next few weeks. However, you can also test for the presence of the `CHEZMOI_SOURCE_FILE` env var directly, which should be more robust.

Not where I need to: I have a flag `--add` where the user invokes chezmoi_modify_manager directly. This is basically like `chezmoi add` but sets up a template modify script etc, and handles the case of converting managed-by-chezmoi to managed-by-chezmoi_modify_manager. Here is where I need to decide on if the added modify script should be templated or not, etc.

I wrote some documentation on how all of this works (with placeholder chezmoi version numbers):

https://github.com/VorpalBlade/chezmoi_modify_manager/blob/567f1b2069d36fdac5c909714a25421b2f5a95fe/doc/source_specification.md

I cannot just switch over to `CHEZMOI_SOURCE_FILE` if set, since the user may not be following the official naming scheme (it was never documented as unsupported to do so, and as such would be a breaking change).

> 
> > 2. Is the output of `chezmoi --version` stable? (I.e. will always contain a substring on the form `v[0-9]+\.[0-9]+\.[0-9]+`, where the version presumably follows semantic versioning?)
> 
> Yes. To be extra robust, look for the first occurrence of this regexp in the output of `chezmoi --version`, which may contain multiple lines.

Awesome.

Issue #2273, opened 2022-08-19
Rewrite target paths


twpayne:
## Is your feature request related to a problem? Please describe.

Some files are in different directories on different machines, for example user-installed fonts are in `~/.local/share/fonts` on Linux and `~/Library/Fonts` on macOS. It would be nice to have a way to handle these differences that is easier to use than [the current method](https://www.chezmoi.io/user-guide/manage-machine-to-machine-differences/#handle-different-file-locations-on-different-systems-with-the-same-contents).

## Describe the solution you'd like

From [this comment](https://github.com/twpayne/chezmoi/issues/2138#issuecomment-1159313813), roughly it could look something like:

1. In your config file, you specify something like:

    ```toml
    [rewriteTargetPaths]
        "Documents" = "One Drive - {organization}/Documents"
    ```

2. This would tell chezmoi that whenever it finds `Documents` in your source state it should replace it with `One Drive - {organization}/Documents` in the target state.

3. To handle machine-to-machine differences, there's the [existing config file generation mechanism](https://www.chezmoi.io/reference/special-files-and-directories/chezmoi-format-tmpl/) which we can use. So, for example, your config file template would look something like:

    ```
    [rewriteTargetPaths]
        "Documents" = "One Drive - {{ registry "HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders\Personal" }}"
    ```

   (assuming the addition of a `registry` template function to get values from the Windows registry)

4. For the [use case of different directories in the FAQ entry](https://www.chezmoi.io/user-guide/manage-machine-to-machine-differences/#ignore-files-or-a-directory-on-different-machines), the config file template would look something like:

    ```
    {{ if .chezmoi.os == "darwin" }}
    # config file directory locations are different on macOS (darwin)
    [rewriteTargetPaths]
        ".config/app" = "Library/Application Support/App"
    {{ end }}
    ```

Implementing this has some subtleties (e.g. what should happen if multiple source directories are re-written to a single target directory) and will likely be non-trivial to implement.

## Describe alternatives you've considered

Keeping things as they are.

## Additional context

None.

k4lizen:
I would also really like this feature since programs often have whole folders for configuration files. This person especially seems to be struggling https://stackoverflow.com/a/78245168 xd

Ra2-IFV:
[...]

PR #3528, opened 2024-01-31
Move default `chezmoistate.boltdb` to `$XDG_STATE_HOME`

62832:
Hey, thanks for making Chezmoi, it's a very neat dotfile manager. I figured I'd make a PR to it, however, as there was one small thing bugging me about its compliance with the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

As `chezmoistate.boltdb` is not actually a _config_ file, but a persistent _state_ file, it makes far more sense to move this to `XDG_STATE_HOME` by default in order to not clutter the config directory, and as per the most recent version of the spec.

> The `$XDG_STATE_HOME` contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in `$XDG_DATA_HOME`. It may contain:
> - actions history (logs, history, recently used files, …)
> - current state of the application that can be reused on a restart (view, layout, open files, undo history, …

Tests have been adapted in order to now look for the new state directory created by Chezmoi by default, with a couple of the tests in `initconfig.txtar` removed as the state file is now completely independent of the config file location.

I hope you're happy with my contribution. Keep up the good work!

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -33,9 +33,9 @@ sections:
       description: Default pager CLI command
     persistentState:
       default: >-
-        `$XDG_CONFIG_HOME/chezmoi/chezmoi.boltdb` <br/>
-        `$HOME/.config/chezmoi/chezmoi.boltdb` <br/>
-        `%USERPROFILE%/.config/chezmoi/chezmoi.boltdb`
+        `$XDG_STATE_HOME/chezmoi/chezmoi.boltdb` <br/>
+        `$HOME/.local/state/chezmoi/chezmoi.boltdb` <br/>
+        `%USERPROFILE%/.local/state/chezmoi/chezmoi.boltdb`
       description: Location of the persistent state file
     progress:
       type: bool
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -33,9 +33,9 @@ sections:
       description: Default pager CLI command
     persistentState:
       default: >-
-        `$XDG_STATE_HOME/chezmoi/chezmoi.boltdb` <br/>
-        `$HOME/.local/state/chezmoi/chezmoi.boltdb` <br/>
-        `%USERPROFILE%/.local/state/chezmoi/chezmoi.boltdb`
+        `$XDG_STATE_HOME/chezmoi/chezmoistate.boltdb` <br/>
+        `$HOME/.local/state/chezmoi/chezmoistate.boltdb` <br/>
+        `%USERPROFILE%/.local/state/chezmoi/chezmoistate.boltdb`
       description: Location of the persistent state file
     progress:
       type: bool
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -289,7 +289,6 @@ type configState struct {
 
 var (
 	chezmoiRelPath             = chezmoi.NewRelPath("chezmoi")
-	persistentStateFileRelPath = chezmoi.NewRelPath("chezmoistate.boltdb")
 	httpCacheDirRelPath        = chezmoi.NewRelPath("httpcache")
 
 	configStateKey = []byte("configState")
@@ -2169,24 +2168,13 @@ func (c *Config) persistentStateFile() (chezmoi.AbsPath, error) {
 	if !c.PersistentStateAbsPath.Empty() {
 		return c.PersistentStateAbsPath, nil
 	}
-	if !c.getConfigFileAbsPath().Empty() {
-		return c.getConfigFileAbsPath().Dir().Join(persistentStateFileRelPath), nil
-	}
-	for _, configDir := range c.bds.ConfigDirs {
-		configDirAbsPath, err := chezmoi.NewAbsPathFromExtPath(configDir, c.homeDirAbsPath)
-		if err != nil {
-			return chezmoi.EmptyAbsPath, err
-		}
-		persistentStateFile := configDirAbsPath.Join(chezmoiRelPath, persistentStateFileRelPath)
-		if _, err := os.Stat(persistentStateFile.String()); err == nil {
-			return persistentStateFile, nil
-		}
-	}
-	defaultConfigFileAbsPath, err := c.defaultConfigFile(c.fileSystem, c.bds)
+
+        // Fallback to XDG Base Directory Specification default.
+	stateHomeAbsPath, err := chezmoi.NewAbsPathFromExtPath(c.bds.StateHome, c.homeDirAbsPath)
 	if err != nil {
 		return chezmoi.EmptyAbsPath, err
 	}
-	return defaultConfigFileAbsPath.Dir().Join(persistentStateFileRelPath), nil
+	return stateHomeAbsPath.JoinString("chezmoi", "chezmoistate.boltdb"), nil
 }
 
 // progressAutoFunc detects whether progress bars should be displayed.
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -288,8 +288,8 @@ type configState struct {
 }
 
 var (
-	chezmoiRelPath             = chezmoi.NewRelPath("chezmoi")
-	httpCacheDirRelPath        = chezmoi.NewRelPath("httpcache")
+	chezmoiRelPath      = chezmoi.NewRelPath("chezmoi")
+	httpCacheDirRelPath = chezmoi.NewRelPath("httpcache")
 
 	configStateKey = []byte("configState")
 
@@ -2169,7 +2169,7 @@ func (c *Config) persistentStateFile() (chezmoi.AbsPath, error) {
 		return c.PersistentStateAbsPath, nil
 	}
 
-        // Fallback to XDG Base Directory Specification default.
+	// Fallback to XDG Base Directory Specification default.
 	stateHomeAbsPath, err := chezmoi.NewAbsPathFromExtPath(c.bds.StateHome, c.homeDirAbsPath)
 	if err != nil {
 		return chezmoi.EmptyAbsPath, err
--- a/internal/cmd/main_test.go
+++ b/internal/cmd/main_test.go
@@ -120,11 +120,13 @@ func cmdChHome(ts *testscript.TestScript, neg bool, args []string) {
 		homeDir          = ts.MkAbs(args[0])
 		chezmoiConfigDir = path.Join(homeDir, ".config", "chezmoi")
 		chezmoiSourceDir = path.Join(homeDir, ".local", "share", "chezmoi")
+                chezmoiStateDir  = path.Join(homeDir, ".local", "state", "chezmoi")
 	)
 	ts.Check(os.MkdirAll(homeDir, fs.ModePerm))
 	ts.Setenv("HOME", homeDir)
 	ts.Setenv("CHEZMOICONFIGDIR", chezmoiConfigDir)
 	ts.Setenv("CHEZMOISOURCEDIR", chezmoiSourceDir)
+        ts.Setenv("CHEZMOISTATEDIR", chezmoiStateDir)
 	if runtime.GOOS == "windows" {
 		ts.Setenv("USERPROFILE", homeDir)
 	}
@@ -658,6 +660,7 @@ func setup(env *testscript.Env) error {
 	}
 	env.Setenv("CHEZMOICONFIGDIR", path.Join(absSlashHomeDir, ".config", "chezmoi"))
 	env.Setenv("CHEZMOISOURCEDIR", path.Join(absSlashHomeDir, ".local", "share", "chezmoi"))
+        env.Setenv("CHEZMOISTATEDIR", path.Join(absSlashHomeDir, ".local", "state", "chezmoi"))
 	env.Setenv("CHEZMOI_GITHUB_TOKEN", os.Getenv("CHEZMOI_GITHUB_TOKEN"))
 
 	switch runtime.GOOS {
--- a/internal/cmd/main_test.go
+++ b/internal/cmd/main_test.go
@@ -120,13 +120,13 @@ func cmdChHome(ts *testscript.TestScript, neg bool, args []string) {
 		homeDir          = ts.MkAbs(args[0])
 		chezmoiConfigDir = path.Join(homeDir, ".config", "chezmoi")
 		chezmoiSourceDir = path.Join(homeDir, ".local", "share", "chezmoi")
-                chezmoiStateDir  = path.Join(homeDir, ".local", "state", "chezmoi")
+		chezmoiStateDir  = path.Join(homeDir, ".local", "state", "chezmoi")
 	)
 	ts.Check(os.MkdirAll(homeDir, fs.ModePerm))
 	ts.Setenv("HOME", homeDir)
 	ts.Setenv("CHEZMOICONFIGDIR", chezmoiConfigDir)
 	ts.Setenv("CHEZMOISOURCEDIR", chezmoiSourceDir)
-        ts.Setenv("CHEZMOISTATEDIR", chezmoiStateDir)
+	ts.Setenv("CHEZMOISTATEDIR", chezmoiStateDir)
 	if runtime.GOOS == "windows" {
 		ts.Setenv("USERPROFILE", homeDir)
 	}
@@ -660,7 +660,7 @@ func setup(env *testscript.Env) error {
 	}
 	env.Setenv("CHEZMOICONFIGDIR", path.Join(absSlashHomeDir, ".config", "chezmoi"))
 	env.Setenv("CHEZMOISOURCEDIR", path.Join(absSlashHomeDir, ".local", "share", "chezmoi"))
-        env.Setenv("CHEZMOISTATEDIR", path.Join(absSlashHomeDir, ".local", "state", "chezmoi"))
+	env.Setenv("CHEZMOISTATEDIR", path.Join(absSlashHomeDir, ".local", "state", "chezmoi"))
 	env.Setenv("CHEZMOI_GITHUB_TOKEN", os.Getenv("CHEZMOI_GITHUB_TOKEN"))
 
 	switch runtime.GOOS {
--- a/internal/cmd/testdata/scripts/applystate.txtar
+++ b/internal/cmd/testdata/scripts/applystate.txtar
@@ -4,8 +4,8 @@ mksourcedir
 
 # test that chezmoi apply does not modify the state if nothing needs to be done
 exec chezmoi apply --force
-exec sha256sum $CHEZMOICONFIGDIR/chezmoistate.boltdb
+exec sha256sum $CHEZMOISTATEDIR/chezmoistate.boltdb
 cp stdout chezmoistate.boltdb-sha256-pre-apply
 exec chezmoi apply --force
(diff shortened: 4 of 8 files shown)


62832:
Force-pushed a few minutes ago so that the commit message would be more in line with the previous ones. I assume that's what made the earlier linting workflow fail.

62832:
Should be good for one last workflow run now, forgot the linting steps initially.

halostatue:
I haven't dug deeply into this, but what should the behaviour be for a non-default configuration file? Yes `chezmoi.boltdb` is a state file, but that it lives *next to* the configuration file is a significant convenience.

I think that if we were to move the default state file to `$XDG_STATE_HOME`, the resulting state file name should be the result of a hashing function of the configuration path name (`$XDG_STATE_HOME/chezmoi/CONFIGHASH/chezmoi.boltdb`).

That said, on a *practical* level, the level of compliance for XDG on this matter is **should**, not **must**.

> There is a single base directory relative to which user-specific state data should be written. This directory is defined by the environment variable `$XDG_STATE_HOME`.

It's a strong recommendation, not a requirement. There are only a few MUST requirements:

- All paths set in these environment variables must be absolute. If an implementation encounters a relative path in any of these variables it should consider the path invalid and ignore it.
- `$XDG_RUNTIME_DIR` defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

   The lifetime of the directory MUST be bound to the user being logged in. It MUST be created when the user first logs in and if the user fully logs out the directory MUST be removed. If the user logs in more than once he should get pointed to the same directory, and it is mandatory that the directory continues to exist from his first login to his last logout on the system, and not removed in between. Files in the directory MUST not survive reboot or a full logout/login cycle.

   The directory MUST be on a local file system and not shared with any other system. The directory MUST by fully-featured by the standards of the operating system. More specifically, on Unix-like operating systems AF_UNIX sockets, symbolic links, hard links, proper permissions, file locking, sparse files, memory mapping, file change notifications, a reliable hard link count must be supported, and no restrictions on the file name character set should be imposed. Files in this directory MAY be subjected to periodic clean-up. To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.

- A specification that refers to `$XDG_DATA_DIRS` or `$XDG_CONFIG_DIRS` should define what the behaviour must be when a file is located under multiple base directories. It could, for example, define that only the file under the most important base directory should be used or, as another example, it could define rules for merging the information from the different files.

62832:
> Yes `chezmoi.boltdb` is a state file, but that it lives next to the configuration file is a significant convenience.

No, it isn't. If I'm trying to manage a directory meant for configuration files, I don't want to have to deal with files that *aren't configs*.

> I think that if we were to move the default state file to `$XDG_STATE_HOME`, the resulting state file name should be the result of a hashing function of the configuration path name (`$XDG_STATE_HOME/chezmoi/CONFIGHASH/chezmoi.boltdb`).

What good would this do? I haven't seen any correlation between the state and config files to warrant this and I see no reason why the state file and the config file should even be intertwined in this way. As far as I can see these are two wholly-independent files which should be in wholly-independent respective locations.

> That said, on a practical level, the level of compliance for XDG on this matter is **should**, not **must**.

This would also be the case for all other directories than `$XDG_RUNTIME_DIR`, but they are still being adhered to. It's better to be consistent in this case and at least make sure every case is accounted for with the respective directories.

halostatue:
> > Yes `chezmoi.boltdb` is a state file, but that it lives next to the configuration file is a significant convenience. 
> No, it isn't. If I'm trying to manage a directory meant for configuration files, I don't want to have to deal with files that _aren't configs_.

Yes, it is a significant convenience *for chezmoi*. And, beyond `.chezmoi.$TYPE.toml`, you would not be managing anything in `~/.config/chezmoi`. Given that state + config drives pretty much everything in chezmoi, this one file needs to be found in a well-known location regardless of where the configuration file is. Putting it *next to* the configuration file is extremely convenient and self-contained, too.

There are many programs that use `~/.config` for state as well as configuration. Just a few files that I found in my `~/.config` on macOS can be seen below. If I looked in the XDG-equivalent on macOS (`~/Library/Application Support`) I would find similar things. There are SQLite databases (used for AI embeddings), dynamic timestamp-driven ("last updated") semi-config files, a config file (`htoprc`) that warns that it's unstable with a poor syntax parser. The `cheat` command installs all stored cheatsheets under `~/.config`. Karibaner uses it for in-place config backups.

It is extremely common for programs to use `~/.config/MYDIR` as their "config and everything else" dumping point and ignore XDG entirely (which has always been more aspirational than anything else).

> > I think that if we were to move the default state file to `$XDG_STATE_HOME`, the resulting state file name should be the result of a hashing function of the configuration path name (`$XDG_STATE_HOME/chezmoi/CONFIGHASH/chezmoi.boltdb`).
> 
> What good would this do? I haven't seen any correlation between the state and config files to warrant this and I see no reason why the state file and the config file should even be intertwined in this way. As far as I can see these are two wholly independent files which should be in wholly independent respective locations.

They are not independent. Full stop. If I specify `--config path/to/xyz.toml`, then chezmoi expects to find `path/to/xyz.boltdb` (or maybe `path/to/chezmoi.boltdb`) next to it so that it can know what the previous state was. Otherwise, you are *always* going to start from a zero state (which means always updating externals and not knowing whether scripts were run once or…).

> > That said, on a practical level, the level of compliance for XDG on this matter is **should**, not **must**.
> 
> This would also be the case for all other directories than `$XDG_RUNTIME_DIR`, but they are still being adhered to. It's better to be consistent in this case and at least make sure every case is accounted for with the respective directories.

I will note that git still defaults to `.gitconfig` over `.config/git/config` and that vim uses `~/.vim` (with both state and config present). Neovim does XDG "right", but now uninstalling or resetting neovim configuration and state requires trawling through a bunch of different directories because it leaves crap everywhere (which is the biggest argument *against* XDG "compliance").

I am 👎 on any change to this without accounting for alternate state locations related to alternate configuration files. Also, if we are moving the state file, then the `httpcache` directory should also be moved to `$XDG_CACHE_PATH` with the same restrictions. (External HTTP download sources are cached relative to the chezmoi configuration file as well.)

<details><summary><code>$XDG_CONFIG_DIR</code> "violations"</summary>

```
~/.config/zed/embeddings/stable/embeddings_db
~/.config/zed/embeddings/stable/embeddings_db-wal
~/.config/zed/embeddings/stable/embeddings_db-shm
~/.config/zed/.tmpL6Y6C6
~/.config/zed/.tmp8tD74P
~/.config/zed/.tmpKozACQ
~/.config/broot/launcher/installed-v1
~/.config/configstore/nodemon.json
~/.config/configstore/update-notifier-@static/discharge.json
~/.config/configstore/update-notifier-npm.json
~/.config/configstore/update-notifier-nodemon.json
~/.config/configstore/update-notifier-jest-codemods.json
~/.config/configstore/update-notifier-alex.json
~/.config/op/config
~/.config/op/plugins/used_items/aws.json
~/.config/op/plugins.sh
~/.config/lnav/exec.history
~/.config/lnav/user.history
~/.config/lnav/cmd.history
~/.config/lnav/search-filters.history
~/.config/lnav/view-info-f77cd543f9fcc5bc3f38c0bf06d5bb92.ts1659127242.ppid80001.json
~/.config/lnav/view-info-f77cd543f9fcc5bc3f38c0bf06d5bb92.ts1659126184.ppid80001.json
~/.config/lnav/sql.history
~/.config/lnav/view-info-f77cd543f9fcc5bc3f38c0bf06d5bb92.ts1659126328.ppid80001.json
~/.config/lnav/filter-regex.history
~/.config/lnav/capture.history
~/.config/lnav/search.history
~/.config/lnav/filter-sql.history
~/.config/lnav/log_metadata.db
~/.config/lnav/search-files.history
~/.config/lnav/view-info-f77cd543f9fcc5bc3f38c0bf06d5bb92.ts1659126265.ppid80001.json
~/.config/lnav/view-info-f77cd543f9fcc5bc3f38c0bf06d5bb92.ts1659126212.ppid80001.json
~/.config/lnav/view-info-f77cd543f9fcc5bc3f38c0bf06d5bb92.ts1659126294.ppid80001.json
~/.config/yarn/global/node_modules/.yarn-integrity
~/.config/yarn/global/node_modules/create-playwright/index.js
~/.config/yarn/global/yarn.lock
~/.config/yarn/global/package.json
~/.config/cheat/cheatsheets/community/*
~/.config/htop/htoprc
~/.config/powershell/Microsoft.PowerShell_profile.ps1
~/.config/litecli/history
~/.config/litecli/log
~/.config/fish/fish_variables
~/.config/chezmoi/chezmoistate.boltdb
~/.config/karabiner/automatic_backups/karabiner_20230316.json
~/.config/packer/checkpoint_cache
~/.config/packer/checkpoint_signature
```

</details>

bradenhilton:
I'm tempted to say this should wait for v3. It's more "correct", but it's still a breaking change that will require manual intervention from users to retain their existing state, and most users likely won't see this change in the release notes ("I updated chezmoi to the new minor release and all my scripts ran again...").

All of this is to say nothing of Windows config/cache/state directories under `AppData`. Should they be used as the defaults also to be more correct?

62832:
@bradenhilton 

> All of this is to say nothing of Windows config/cache/state directories under `AppData`. Should they be used as the defaults also to be more correct?

That is also true, it would be good to eventually use those defaults properly under Windows.

@halostatue 

> There are many programs that use `~/.config` for state as well as configuration.

Much of this is simply because they were never brought to date with the current version of the spec, which only even introduced the `$XDG_STATE_HOME` directory around 2021.

> It is extremely common for programs to use `~/.config/MYDIR` as their "config and everything else" dumping point and ignore XDG entirely *(which has always been more aspirational than anything else)*.

True as that may be, it is still a worthy aspiration just to be able to do `ls -a ~` without wanting to gouge your eyes out. At the same time, I'm also not entirely happy about programs that dump everything into `~/.config` when it isn't necessarily config-related, since it pretty much defeats the purpose of having clear, distinct directories for different such purposes.

> They are not independent. Full stop. If I specify ``--config path/to/xyz.toml``, then chezmoi expects to find `path/to/xyz.boltdb` (or maybe `path/to/chezmoi.boltdb`) next to it so that it can know what the previous state was. Otherwise, you are always going to start from a zero state (which means always updating externals and not knowing whether scripts were run once or…).

Then I will concede I am probably misunderstanding the use and significance of `chezmoistate.boltdb` in relation to everything else such as the config location. Regardless, it still doesn't feel right to me for it to be contained in the config directory when the proposal for a dedicated state directory has been around for a while and eventually became part of the spec for at least a couple of years now.

> I am 👎 on any change to this without accounting for alternate state locations related to alternate configuration files. Also, if we are moving the state file, then the `httpcache` directory should also be moved to `$XDG_CACHE_PATH` with the same restrictions. (External HTTP download sources are cached relative to the chezmoi configuration file as well.)

I am happy to give it another look some time tomorrow and address these points in this same PR, if that's what you'd like. I did indeed fail to account for other parts of chezmoi that may not necessarily fit the spec.

halostatue:
> > I'm tempted to say this should wait for v3. It's more "correct", but it's still a breaking change that will require manual intervention from users to retain their existing state, and most users likely won't see this change in the release notes ("I updated chezmoi to the new minor release and all my scripts ran again...").
> 
> No rush at all. I didn't foresee that this would be such a breaking change, so I do apologise for that.

I remain unconvinced of the utility of this (more below), but I have added this to https://github.com/twpayne/chezmoi/discussions/2673#discussioncomment-8330230.

> @halostatue
> 
> > There are many programs that use `~/.config` for state as well as configuration.
> 
> Much of this is simply because they were never brought to date with the current version of the spec, which only even introduced the `$XDG_STATE_HOME` directory around 2021.

[XKCD 927](https://xkcd.com/927/) is always relevant in this sort of discussion. I’m going to be clear up front and say that I believe that the XDG Base Directory Specification is mostly irrelevant at best and actively harmful at worst. Most of the time, it is mostly irrelevant. It is a 23 year old specification that has yet to reach version 1 and came out of the GNOME project for various X desktop systems (KDE, GNOME, etc.) and related applications primarily from a Linux perspective. Most UNIX systems care only about XDG BDS inasmuch as they offer GNOME or KDE desktops for their X servers; this includes OpenBSD, NetBSD, and FreeBSD. Both Windows and macOS (while certainly a UNIX) use entirely different structures — and have some of the same drawbacks for those structures.

The main advantage to *not* following XDG is exemplified in many ways by [bun](https://bun.sh). Install looks like `curl -fsSL https://bun.sh/install | bash` and *complete*‡ uninstall looks like `rm -rf ~/.bun `. That's it. There is nothing else to track down or uninstall. By comparison, to *fully* uninstall chezmoi†, you need to remove the `chezmoi` binary, `~/.config/chezmoi`, `~/.local/share/chezmoi`, and `~/.cache/chezmoi` (with your change, we would need to add `~/.local/state/chezmoi`). If we implement your change cross-platform *properly* (to no longer incorrectly use XDG for chezmoi configuration, state, and cache on Windows and macOS), those directory locations would need to be modified. It's worth noting that on macOS, there is a category of application which detects app removals (e.g., `Google Chrome.app`) and offers to delete configuration, cache, &c. files found in `~/Library` (none of them do anything for command-line tools).

‡ If you generate shell completions, you may have to remove the file from the targeted completion directory. You should also remove `~/.bun/bin` from your `$PATH`.
† Not that anyone would ever want to do that.

There are advantages to using the Windows and macOS equivalents to `$XDG_CACHE_HOME/chezmoi` (`~/Library/Caches/chezmoi`) that are not realized on *any* system (of which I am aware) that "uses" XDG BDS: excessively old or large caches may be purged by the operating system because it is recognized as a cache directory.

This leftover detritus of forgotten installations and unmanaged caches is where following XDG can be actively harmful on a system. Where it's mostly harmless is it sort of makes the same arguments that a system's Filesystem Hierarchy Specification (which for Linux is at version 3.0; FreeBSD has a similar definition at hier(7)) so that different trees could be mounted on different drives. In fact, `$XDG_RUNTIME_DIR` *specifies* that the directory should be purged on logout / reboot suggesting that *maybe* it should be mounted on a `tmpfs` even though it is a user directory. But aside from two instances using bind mounts (#3520 and #2432), very few people on Linux have even considered mounting something in `~` on a different drive. (I suspect it matters if you are still on spinning rust or an NFS mount home directory. I haven't been on anything except fast flash drives since 2014 — two generations of macOS laptops.)

> > It is extremely common for programs to use `~/.config/MYDIR` as their "config and everything else" dumping point and ignore XDG entirely _(which has always been more aspirational than anything else)_.
> 
> True as that may be, it is still a worthy aspiration just to be able to do `ls -a ~` without wanting to gouge your eyes out. At the same time, I'm also not entirely happy about programs that dump everything into `~/.config` when it isn't necessarily config-related, since it pretty much defeats the purpose of having clear, distinct directories for different such purposes.

No, I just want to gouge my eyes out if I have to look at `ls ~/.config` or `ls ~/.local/share` or `ls ~/.local/state` or `ls ~/.cache` or worse, try to figure out which one is the source of my problems. I recently wanted to restart my neovim configuration from scratch and it took me a couple of hours to figure out that files in `~/.local/share` *and* `~/.local/state` were breaking my setup (requiring an `rm -rf ~/.local/share/nvim ~/.local/state/nvim`). In vim, restarting that configuration from scratch is as simple as `mv ~/.vim{,.old} && mkdir ~/.vim && touch ~/.vim/vimrc`.

It would be worse if I have to remember that I need look at this in `~/Library/Application Support` and `~/Library/Caches` for macOS (and sadly, without #2273, that's problematic for things like lazygit and lazydocker, where on Linux it would be in `~/.config/lazygit`, but on macOS it's in `~/Library/Application Support/lazygit`, but I have a common `.tmpl` for both).

> > They are not independent. Full stop. If I specify `--config path/to/xyz.toml`, then chezmoi expects to find `path/to/xyz.boltdb` (or maybe `path/to/chezmoi.boltdb`) next to it so that it can know what the previous state was. Otherwise, you are always going to start from a zero state (which means always updating externals and not knowing whether scripts were run once or…).
> 
> Then I will concede I am probably misunderstanding the use and significance of `chezmoistate.boltdb` in relation to everything else such as the config location. Regardless, it still doesn't feel right to me for it to be contained in the config directory when the proposal for a dedicated state directory has been around for a while and eventually became part of the spec for at least a couple of years now.

XDG BDS should be considered aspirational at best. Treating it as a specification that actually *means* something is going to lead to heartache for *someone*, and probably multiple someones, as the limitations of an ivory tower specification become painfully clear in practice.

With respect to `chezmoistate.boltdb`, its position is not important…except:

1. There must be a separate instance for each configuration file that gets specified. Putting the state file next to the main config file makes this relationship clearer. Falling back to `$XDG_STATE_HOME/chezmoi/chezmoistate.boltdb` breaks because it is not reactive to the input configuration file. This is why I said that it would need to be `$XDG_STATE_HOME/chezmoi/$CONFIG_PATH_HASH/chezmoistate.boltdb`. The original location could be used in the case that the configuration file is in the default location.
2. If the default location is moved, there must be a migration path, preferably *without* requiring end-user intervention (I would argue that this should be the case even on a v2 to v3 upgrade).

> I am happy to give it another look some time tomorrow and address these points in this same PR, if that's what you'd like. I did indeed fail to account for other parts of chezmoi that may not necessarily fit the spec. In any case, I'm sorry again that this wasn't up to scratch in the end.

Your code changes were not bad (although removing tests should *typically* be a sign that you're not understanding something if you are not deep in the middle of the application). I did review them after dropping my last comment. But I think this well-intentioned change is unnecessary, has a lot of potentially *nasty* corner cases, and will result in a required migration.

If you feel strongly for this, please advocate for it in the comment I put on #2673 so it can be considered for v3. But while I won't close this (Tom may disagree with my opinion, so I will leave that choice to him), I think there are enough downsides that it can't be merged any time soon.

twpayne:
Thanks for the PR and the discussion here.

I agree that `chezmoistate.boltdb` is really state, and not configuration, and so in theory should live in `$XDG_STATE_HOME`, although chezmoi's support of the XDG Base Directory Specification precedes the addition of `$XDG_STATE_HOME` to the specification.

That said, there are multiple complicating factors here, as already mentioned by @halostatue and @bradenhilton:

1. There needs to be a seamless upgrade path for current users.

2. `chezmoistate.boltdb` is special because it is heavily tied to a specific config file. chezmoi does support multiple config files, so there is more subtlety to using `$XDG_STATE_HOME` than just moving the location of the state file.

3. `chezmoistate.boltdb` is special because deleting it will cause a lot of changes to the user's system the next time they run `chezmoi apply`. `run_onchange_` and `run_once_` scripts will be run again, and chezmoi will lose track of local modifications to managed files, amongst other things.

And some other comments:

* chezmoi already uses `$XDG_CACHE_HOME` for downloads of externals.

* chezmoi can clean itself completely from the user's system with the [`chezmoi purge` command](https://www.chezmoi.io/reference/commands/purge/).

* To a large extent, chezmoi's externally-visible behavior is defined by its tests, and people rely on this behavior ([XKCD 1172](https://xkcd.com/1172/)). Tests can be added, but may only removed if they are clearly faulty. That is not the case for the tests removed in this PR.

Overall, I agree that any changes here should be for v3, and furthermore that chezmoi should follow the local operating system's guidelines for its config and state locations (e.g. `~/Library/Application Support` on macOS and `AppData` on Windows).

So thank you for this well-written PR, but there are deeper issues to solve first and these issues will require more work than has already been done in this PR.

PR #3555, opened 2024-02-11
docs: Cross link "See REF" in config variables doc

halostatue:
No files specified

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -144,7 +144,7 @@ sections:
   diff:
     args:
       type: '[]string'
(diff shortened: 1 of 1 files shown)


assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
```diff
@@ -294,7 +294,7 @@ sections:
       description: KeePassXC database
     mode:
       default: '`cache-password`'
-      description: See section on KeePassXC
```
twpayne, line 297: For a reason I don't understand, this link doesn't work for me in local testing. But this PR is really good :)

halostatue: Should be fixed now.

twpayne (approved):

Issue #3498, opened 2024-01-20
Improve support for 1Password service accounts


jimmythomson:
## Is your feature request related to a problem?

I'm using 1Password and have setup a service account for storing my various credentials. I've then setup the 1Password CLI as described here: https://developer.1password.com/docs/service-accounts/use-with-1password-cli/

That page suggests that the user stores their service account's token in the `OP_SERVICE_ACCOUNT_TOKEN` environment variable.

With `OP_SERVICE_ACCOUNT_TOKEN` defined, I then test the reading of an item with:
```
chezmoi execute-template '{{ onepasswordRead "op://my-vault/my-item/my-field" }}'
```
This fails as `chezmoi` tries to communicate with the 1Password app, which isn't installed on my system.
As far as I understand it, the mere existence of the `OP_SERVICE_ACCOUNT_TOKEN` environment variable is enough for the `op read` to succeed. Indeed, if I run the following directly, then it succeeds:
```
op read  "op://my-vault/my-item/my-field"
```
I took a quick look at the code and have a work around, which is simply to export the environment variable as `OP_SESSION_SERVICE_ACCOUNT_TOKEN`. The code appears to look specifically for the `OP_SESSION_` prefix. After having done this, then the `chezmoi execute-template` function succeeds.

## Describe the solution you'd like

Ideally I'd like for `chezmoi` to automatically look for the `OP_SERVICE_ACCOUNT_TOKEN` environment variable. Perhaps even some documentation to describe the naming of the env var for service accounts would be enough (i.e. the importance of the `OP_SESSION_` prefix).

## Thanks
This is an amazing utility. Many thanks to all the contributors!

halostatue:
I don't currently have a service account set up on either of my 1Password accounts, but I’ll look to see what I can see with this. It will not be today.

I think we might be able to switch modes if `OP_SERVICE_ACCOUNT_TOKEN` is defined, but I need to understand how service account works better so that I can figure out what it would do to our tests (I *suspect* that I’m going to have two `.txtar` files since the application would be called differently by the same command parameters).

Worst case, we *may* need to have separate template verbs for service account handling or 1Password Connect handling, because the current template verbs permit access from more than one account, and it doesn't look like service account can do that.

twpayne:
> This means that we either need to have _separate_ functions (I am very 👎 on this, but there is one argument in favour of this) or determine which _mode_ we are in. This would be (in order): connect, service account, normal.

[...]

PR #3557, opened 2024-02-12
feat: 1Password Secrets Automation Support

halostatue:
Added a new configuration option, `onepassword.mode`, that can be set to
`account` (the default value), `connect` (for 1Password Connect), or
`service` (for 1Password Service Accounts).

When in `onepassword.mode` is:

- `account`: the presence of `OP_SERVICE_ACCOUNT_TOKEN`,
  `OP_CONNECT_HOST`, and/or `OP_CONNECT_TOKEN` will cause chezmoi to
  immediately exit; their presence changes the behaviour of the
  1Password CLI.

- `connect`: the presence of `OP_SERVICE_ACCOUNT_TOKEN` or absences of
  `OP_CONNECT_HOST` and/or `OP_CONNECT_TOKEN` will cause chezmoi to
  immediately exit.

  Additionally, the use of `onepasswordDocument` or passing an `account`
  parameter will cause immediate exits.

- `service`: the absence of `OP_SERVICE_ACCOUNT_TOKEN` or presence of
  `OP_CONNECT_HOST` and/or `OP_CONNECT_TOKEN` will cause chezmoi to
  immediately exit.

  Additionally, passing an `account` parameter will cause immediate
  exits.

In all other ways, the 1Password template functions have not changed.

I changed the format of the scripts in the `onepassword*.txtar` files so
that there is better parity between the three sets of tests _and_ so
that there is better / easier to read parity between the Unix and
Windows test scripts.

Closes: #3498

### DIFF
--- a/internal/cmd/onepasswordtemplatefuncs.go
+++ b/internal/cmd/onepasswordtemplatefuncs.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -11,6 +12,14 @@ import (
 	"github.com/coreos/go-semver/semver"
 
 	"github.com/twpayne/chezmoi/v2/internal/chezmoilog"
+)
+
+type onepasswordMode string
+
+const (
+	onepasswordModeAccount onepasswordMode = "account"
+	onepasswordModeConnect onepasswordMode = "connect"
+	onepasswordModeService onepasswordMode = "service"
 )
 
 type withSessionTokenType bool
@@ -34,12 +43,14 @@ type onepasswordAccount struct {
 }
 
 type onepasswordConfig struct {
-	Command       string `json:"command" mapstructure:"command" yaml:"command"`
-	Prompt        bool   `json:"prompt"  mapstructure:"prompt"  yaml:"prompt"`
+	Command       string          `json:"command" mapstructure:"command" yaml:"command"`
+	Prompt        bool            `json:"prompt"  mapstructure:"prompt"  yaml:"prompt"`
+	Mode          onepasswordMode `json:"mode"    mapstructure:"mode"    yaml:"mode"`
 	outputCache   map[string][]byte
 	sessionTokens map[string]string
 	accountMap    map[string]string
 	accountMapErr error
+	modeChecked   bool
 }
 
 type onepasswordArgs struct {
@@ -54,6 +65,10 @@ type onepasswordItem struct {
 }
 
 func (c *Config) onepasswordTemplateFunc(userArgs ...string) map[string]any {
+	if err := c.onepasswordCheckMode(); err != nil {
+		panic(err)
+	}
+
 	args, err := c.newOnepasswordArgs([]string{"item", "get", "--format", "json"}, userArgs)
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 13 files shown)


internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -237,10 +257,14 @@ func (c *Config) onepasswordReadTemplateFunc(url string, args ...string) string
...
 	case 1:
+		if c.Onepassword.Mode != onepasswordModeAccount {
+			panic(fmt.Errorf("1Password account parameters cannot be used in %s mode", c.Onepassword.Mode))
+		}
+
 		onepasswordArgs.account = c.onepasswordAccount(args[0])
 		onepasswordArgs.args = append(onepasswordArgs.args, "--account", onepasswordArgs.account)
 	default:
-		panic(fmt.Errorf("expected 1 or 2 arguments, got %d", len(args)))
+		panic(fmt.Errorf("expected 1 or 2 arguments, got %d", len(args)+1))
```
twpayne, line 267: Is the `+1` correct here?

internal/cmd/onepasswordtemplatefuncs.go
```diff
@@ -290,8 +324,20 @@ func (c *Config) onepasswordAccounts() (map[string]string, error) {
 }
 
 func (c *Config) newOnepasswordArgs(baseArgs, userArgs []string) (*onepasswordArgs, error) {
-	if len(userArgs) < 1 || 3 < len(userArgs) {
-		return nil, fmt.Errorf("expected 1, 2, or 3 arguments, got %d", len(userArgs))
+	maxArgs := 3
+	if c.Onepassword.Mode != onepasswordModeAccount {
+		maxArgs = 3
```
twpayne, line 329: `maxArgs` is already 3, so this has no effect.

twpayne (commented):
Thank you, this looks excellent. Just a couple of nits/questions.

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.pdm-python
```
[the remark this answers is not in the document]
halostatue: No…and yes.

I'm sporadically working on a PR to use a Python venv using PDM as the package manager for the Python tasks in chezmoi development. It has a number of really nice advantages over the current `requirements.txt`. It isn't *quite* ready, but every time I switch branches this shows up as unmanaged, and I got annoyed this time and added it to `.gitignore`.

It's actually part of the "since-version" chip that I was talking about, but I should probably extract the PDM part and open it as a PR on its own.

I can remove that from this PR if you wish.

internal/cmd/onepasswordtemplatefuncs.go
halostatue: I think I added it when I was doing some testing because the output looked odd. I can try some other manual testing later to be certain.

internal/cmd/onepasswordtemplatefuncs.go
halostatue: Oops. That was supposed to be `2`, not `3`.

.gitignore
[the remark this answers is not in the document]
twpayne: Please do remove it from this PR -- you can add `.pdm-python` to your global `.gitignore` file in the short term and add it to chezmoi's `.gitignore` file with your PDM PR :)

internal/cmd/onepasswordtemplatefuncs.go
halostatue: Yes, it is.

```console
# with +1
$ go run . execute-template '{{ onepasswordRead "a" "b" "c" }}'
chezmoi: template: arg1:1:3: executing "arg1" at <onepasswordRead "a" "b" "c">: error calling onepasswordRead: expected 1 or 2 arguments, got 3
exit status 1
# without +1
$ go run . execute-template '{{ onepasswordRead "a" "b" "c" }}'
chezmoi: template: arg1:1:3: executing "arg1" at <onepasswordRead "a" "b" "c">: error calling onepasswordRead: expected 1 or 2 arguments, got 2
exit status 1
```

For `onepasswordRead`, users won't read it as `url, ...args` but as `"a" "b" "c"`. This is a bit of a drive-by error output bug fix.

twpayne (approved):
Thank you, this is fantastic.

PR #3565, opened 2024-02-13
fix: Prioritize $HOME over %userprofile% on Windows for Cygwin/msys2 compatibility

### LINKED ISSUE #3561: chezmoi does not recognize `msys2` or `cygwin` $HOME
## Describe the bug

In case you are using chezmoi in cygwin or msys on windows 11 fresh install, installed by either `scoop install chezmoi` or the sh technique, if you do it on a msys2 shell, chezmoi gets confused about where the home is since the msys2 home is not the same as the windows home.. Not sure if you consider this use case important.

twpayne:
Fixes #3561.

I think this is right. Please find ways to break it :)

cc @Kreijstal

### DIFF
--- a/internal/chezmoi/abspath.go
+++ b/internal/chezmoi/abspath.go
@@ -2,7 +2,6 @@ package chezmoi
 
 import (
 	"fmt"
-	"os"
 	"path"
 	"path/filepath"
 	"reflect"
@@ -176,7 +175,7 @@ func (p *AbsPath) UnmarshalText(text []byte) error {
 
 // HomeDirAbsPath returns the user's home directory as an AbsPath.
 func HomeDirAbsPath() (AbsPath, error) {
-	userHomeDir, err := os.UserHomeDir()
+	userHomeDir, err := UserHomeDir()
 	if err != nil {
 		return EmptyAbsPath, err
 	}
--- a/internal/chezmoi/chezmoi_unix.go
+++ b/internal/chezmoi/chezmoi_unix.go
@@ -4,6 +4,7 @@ package chezmoi
 
 import (
 	"io/fs"
+	"os"
 
 	"golang.org/x/sys/unix"
 )
@@ -24,6 +25,11 @@ func findExecutableExtensions(path string) []string {
 // IsExecutable returns if fileInfo is executable.
 func IsExecutable(fileInfo fs.FileInfo) bool {
 	return fileInfo.Mode().Perm()&0o111 != 0
+}
+
+// UserHomeDir on UNIX returns the value of os.UserHomeDir.
+func UserHomeDir() (string, error) {
+	return os.UserHomeDir()
 }
 
 // isPrivate returns if fileInfo is private.
--- a/internal/chezmoi/chezmoi_windows.go
+++ b/internal/chezmoi/chezmoi_windows.go
@@ -44,6 +44,17 @@ func IsExecutable(fileInfo fs.FileInfo) bool {
 	})
 }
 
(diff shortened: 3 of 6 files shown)


halostatue:
The changes looks fine, but I’m less certain that it's a "safe" change for Windows. Realizing that my Windows experience is years (more than a decade; I bypassed all of Vista–11 because we used the enterprise editions at work and I stopped doing Windows development entirely in 2012) out of date, I vaguely recall setting `%HOME%` at times on my work machines, which would have messed up chezmoi working in Windows without being in a cygwin/msys2 shell.

If it were clear that `chezmoi` were running under a cygwin/msys2 shell, prioritizing `$HOME` over `%UserProfile%` automatically makes sense. I'm not sure that's easy or cheap on Windows (but `uname -s` [can help](https://en.wikipedia.org/wiki/Uname#Examples) on systems that have it).

Kreijstal:
> The changes looks fine, but I’m less certain that it's a "safe" change for Windows. Realizing that my Windows experience is years (more than a decade; I bypassed all of Vista–11 because we used the enterprise editions at work and I stopped doing Windows development entirely in 2012) out of date, I vaguely recall setting `%HOME%` at times on my work machines, which would have messed up chezmoi working in Windows without being in a cygwin/msys2 shell.
> 
> If it were clear that `chezmoi` were running under a cygwin/msys2 shell, prioritizing `$HOME` over `%UserProfile%` automatically makes sense. I'm not sure that's easy or cheap on Windows (but `uname -s` [can help](https://en.wikipedia.org/wiki/Uname#Examples) on systems that have it).

I mean it is quite safe since home is definitely not defined, and if it is, way more unix apps would break, considering chezmoi is a terminal app, you can be safe that you'll not be the only app that breaks in this manner. Also, if you are not familiar with windows, then msys2 will make you feel at home-ish.
I'll give 2 strings of uname -s under msys
windows 8:
MINGW64_NT-6.3-9600
windows 11:
MSYS_NT-10.0-22631

the first word is just the msys `msystem` env variable, that is not so important, but what comes after NT is the version, and NT means windows, kind of.

bradenhilton:
We could instead check if `CYGWIN` or `MSYSTEM` are defined, but I can't discern if they are defined by default or must be user defined.

Perhaps start by checking if `CYGWIN` or `MSYSTEM` are defined, else check if `HOME` is defined, else fall back to `os.UserHomeDir`?

halostatue:
Regardless of this change, I think that there's a chance that chezmoi may not behave as expected under cygwin (and possibly msys2 shells) because `$HOME` is going to be something like `/home/USERNAME`, but `C:\Program Files` is going to be `/mnt/c/program\ files` or something like that.

Like I said, it’s been a *long* time since I used Windows in earnest, and even longer since I used msys2 or Cygwin. There were path issues back then with using non-Cygwin-aware programs (e.g., gvim) from within Cygwin, but I don't really recall what they were.

I think that this is *definitely* worth trying, but I do think that there's risk involved. Would it make sense to have a configuration flag in `chezmoi.$FORMAT`?

twpayne:
> Would it make sense to have a configuration flag in `chezmoi.$FORMAT`?

There's a circular dependency as `chezmoi.$FORMAT` is in `$HOME/.config/chezmoi`.



> Perhaps start by checking if `CYGWIN` or `MSYSTEM` are defined, else check if `HOME` is defined, else fall back to `os.UserHomeDir`?

I like this solution. I've updated the PR.

halostatue (approved):

PR #3572, opened 2024-02-15
fix: Use uname -o to detect Android in install script

### LINKED ISSUE #3571: Better android detection for install script

In Termux there is tool called `proot-distro` which allows to run (limited) linux distro on android withoit root and which can cooperate with termux utils.
So when I am running `sh -c "$(curl -fsLS get.chezmoi.io)"` inside of `proot-distro` to install _linux_ chezmoi build, it instead installs _android_ chezmoi build by default, which is frustrating and leads to wrong dotfiles installed

Use `[ "$(uname -o)" = "Android" ]` check instead of checking existence of `termux-info` as this is more reliable

Current workaround is to remove Termux's `$PREFIX/bin` directory from `PATH` or to login into `proot-distro` with `--isolated` flag

twpayne:
Fixes #3571.

### DIFF
--- a/assets/scripts/install-local-bin.sh
+++ b/assets/scripts/install-local-bin.sh
@@ -119,11 +119,10 @@ get_goos() {
 	case "${os}" in
 	cygwin_nt*) goos="windows" ;;
 	linux)
-		if is_command termux-info; then
-			goos=android
-		else
-			goos=linux
-		fi
+		case "$(uname -o | tr '[:upper:]' '[:lower:]')" in
+		android) goos="android" ;;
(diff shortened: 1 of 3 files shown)


twpayne:
@ErrrorMaxx @Kreijstal would you be able to test this?

ErrrorMaxx:
Works like a charm:
os-arch android/arm64 in Termux
os-arch linux/arm64 (Ubuntu 23.10 (Mantic Minotaur)) in `proot-distro`

twpayne:
Thank you for testing :)

PR #3569, opened 2024-02-15
chore: Refactor install.ps1

bradenhilton:
I've been wanting to do this for a long time but always seemed to find something else to do instead.

Changes:

- Use PowerShell's [Approved Verbs](https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-5.1)
- Use PowerShell's built-in output streams
- Remove unused functions
- General style changes to align more with the [unofficial style guide](https://poshcode.gitbooks.io/powershell-practice-and-style)
- ~Use GitHub's REST API (I am concerned this may cause issues with actions but I'm not sure. If this is preferred I can also make this change in `install.sh.tmpl`)~
- Removed almost 100 lines

`get_goos` handled FreeBSD, but I'm struggling to even find any information about PowerShell being installable on FreeBSD. https://github.com/PowerShell/PowerShell?tab=readme-ov-file#get-powershell makes no mention of FreeBSD, nor does https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md.

```console
PowerShell 7.4.1

~
❯ [System.Runtime.InteropServices.RuntimeInformation]::FrameworkDescription
.NET 8.0.1
```

https://ports.freebsd.org/cgi/ports.cgi?query=powershell&stype=all&sektion=all
https://ports.freebsd.org/cgi/ports.cgi?query=pwsh&stype=all&sektion=all

TODO:

- [X] Verify that the invocation remains the same when downloading and executing the script over the network
- [X] Test on Linux and macOS

There's no rush for this. I want to be absolutely sure there are no regressions before merging.

PR #3575, opened 2024-02-16
chore: Use Unix-style line endings in format-yaml.py output

bradenhilton:
Refs https://github.com/twpayne/chezmoi/pull/3569#issuecomment-1948983080


twpayne:
On further investigation, I don't think this is the correct way to do this. Alternative PR coming up...

PR #3576, opened 2024-02-16
chore: Work around ruamel.yaml errors when formatting YAML

twpayne:
This is an alternative to #3575.

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -390,7 +390,7 @@ sections:
       description: Vault CLI command
   update:
     apply:
-      type: 'bool'
+      type: bool
       default: '`true`'
       description: Apply after pulling
     args:
--- a/assets/requirements.txt
+++ b/assets/requirements.txt
@@ -1,1 +1,1 @@
-ruamel.yaml==0.18.5
+ruamel.yaml==0.18.6
--- a/assets/scripts/format-yaml.py
+++ b/assets/scripts/format-yaml.py
@@ -10,10 +10,17 @@ from ruamel.yaml import YAML
 
 def main() -> int:
     yaml = YAML()
+    # ruamel.yaml.YAML will by default use the native line ending, which leads
+    # to differences between Windows and UNIX. Force the output to use UNIX line
+    # endings.
+    yaml.line_break = '\n'
+    # ruamel.yaml.YAML will by default break long lines and introduce trailing
+    # whitespace errors. Disable this behavior by setting a long line width.
+    yaml.width = 1024
     for filename in sys.argv[1:]:
         with Path(filename).open('r') as file:
             data = yaml.load(file)
-        with Path(filename).open('w') as file:
+        with Path(filename).open('w', newline='\n') as file:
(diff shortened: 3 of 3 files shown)


bradenhilton:
This still outputs CRLF on Windows:

```console
❯ python .\assets\scripts\format-yaml.py .\.github\workflows\main.yml

❯ (Get-Content .\.github\workflows\main.yml -Raw) -match '\r\n$'
True
```

I think we still need the change from https://github.com/twpayne/chezmoi/pull/3575 (perhaps in addition to `yaml.line_break = '\n'`) as well as `yaml.width = 1024`. I'll close https://github.com/twpayne/chezmoi/pull/3575 in favor of this.

Here is the same check with both newline changes combined:

```
❯ Get-Content .\assets\scripts\format-yaml.py
#!/usr/bin/env python3

from __future__ import annotations

import sys
from pathlib import Path

from ruamel.yaml import YAML


def main() -> int:
    yaml = YAML()
    # ruamel.yaml.YAML will by default use the native line ending, which leads
    # to differences between Windows and UNIX. Force the output to use UNIX line
    # endings.
    yaml.line_break = '\n'
    # ruamel.yaml.YAML will by default break long lines and introduce trailing
    # whitespace errors. Disable this behavior by setting a long line width.
    yaml.width = 1024
    for filename in sys.argv[1:]:
        with Path(filename).open('r') as file:
            data = yaml.load(file)
        with Path(filename).open('w', newline='\n') as file:
            yaml.dump(data, file)
    return 0


if __name__ == '__main__':
    raise SystemExit(main())

❯ python .\assets\scripts\format-yaml.py .\.github\workflows\main.yml

❯ (Get-Content .\.github\workflows\main.yml -Raw) -match '\r\n$'
False

❯ ec -config .\.editorconfig .\.github\workflows\main.yml 
<empty>
```

`ec` also doesn't error if I do the same after commenting out `yaml.line_break = '\n'`, but that means the script is basically just https://github.com/twpayne/chezmoi/pull/3575 with the addition of `yaml.width = 1024`, and that was still giving me the error in `main/check` in https://github.com/twpayne/chezmoi/pull/3569.

I think the primary culprit could be the line wrapping. Take for instance this line:

```yaml
if: github.event_name == 'pull_request' && github.event.pull_request.draft == false
```

With the default line width, it becomes

```yaml
if: github.event_name == 'pull_request' && github.event.pull_request.draft ==
  false
```

Note the different highlighting. I can also observe this in my editor:

image

bradenhilton:
I just pushed `main.yml` in https://github.com/twpayne/chezmoi/pull/3569 after formatting it with all changes combined and `main/check` passed.

twpayne:
Thanks, I've updated this PR with your change from #3575.

> With the default line width, it becomes
> 
> ```yaml
> if: github.event_name == 'pull_request' && github.event.pull_request.draft ==
>   false
> ```
> 
> Note the different highlighting.

I think the syntax highlighting is misleading. Most syntax highlighters are regular expression based and only have a very basic understanding of the language's grammar, and so make occasional mistakes (especially in the case of complex and poorly-defined languages like YAML). [tree-sitter](https://tree-sitter.github.io/tree-sitter/) does it right.

bradenhilton (approved):

Issue #3366, opened 2023-11-26
Speed up release GitHub Action


twpayne:
## Is your feature request related to a problem? Please describe.

I'm always frustrated when chezmoi's CI takes about 40 minutes of wall clock time. I wish that chezmoi's CI would run quicker. This is especially frustrating in release builds where the [`test-release` job](https://github.com/twpayne/chezmoi/blob/694977b904e888ba285aa0fd44617d1d59d89bc7/.github/workflows/main.yml#L143-L205) and the [`release` job](https://github.com/twpayne/chezmoi/blob/694977b904e888ba285aa0fd44617d1d59d89bc7/.github/workflows/main.yml#L343-L395) each take about twenty minutes to run, one after the other, when they share the same work.  

## Describe the solution you'd like

chezmoi's GitHub Actions should re-use work where possible to reduce the overall wall clock time. Specifically, work from the `test-release` job should be re-used in the `release` job.

## Describe alternatives you've considered

Leaving things as they are. The current CI works and only runs a few times per day.

## Additional context

I expressed frustration in [this comment](https://github.com/twpayne/chezmoi/pull/3324#issuecomment-1802931252) and @halostatue asked for clarification in [this comment](https://github.com/twpayne/chezmoi/pull/3324#issuecomment-1803058487).

halostatue:
I’m going to look at ways that this could be improved in my spare time over the holiday week. I can't promise a lot, because I’m also going to start an upgrade to Sonoma (probably tonight).

bradenhilton:
On a more general note regarding the other jobs, would being just a touch more selective about the changes each job depends on be an easy win?

I'm aware we already sort of do this, but I'm wondering if we can't go a bit further and see if it increases or decreases friction.

For example, I'd imagine in most cases that editing one of the install scripts wouldn't necessitate building and testing chezmoi, on several different platforms no less.

Likewise, a quick chore change on the website also shouldn't typically require building and testing chezmoi.

Perhaps we could split them as best we can into groups such as src/website/scripts etc. and be more explicit about what changes the jobs in each group require to run.

The src group would naturally contain the most jobs.

The website group could largely depend on changes to `assets/chezmoi.io/**` etc. and only run jobs related to the website.

Each group could have its own workflow file.

Just thinking out loud.

twpayne:
[...]

PR #3584, opened 2024-02-21
chore: Separate installer CI tests

bradenhilton:
Refs https://github.com/twpayne/chezmoi/issues/3366#issuecomment-1949615942

Experimental. I fully expect it to fail.

### DIFF
--- a/.github/workflows/installer.yml
+++ b/.github/workflows/installer.yml
@@ -1,0 +1,88 @@
+name: installer
+on:
+  pull_request:
+    branches:
+    - master
+  push:
+    branches:
+    - master
+env:
+  SHA: ${{ github.event_name == 'push' && github.sha || github.event.pull_request.head.sha }}
+jobs:
+  changes:
+    runs-on: ubuntu-22.04
+    outputs:
+      sh: ${{ steps.filter.outputs.sh }}
+      ps1: ${{ steps.filter.outputs.ps1 }}
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+    - id: filter
+      uses: dorny/paths-filter@0bc4621a3135347011ad047f9ecf449bf72ce2bd
+      with:
+        filters: |
+          shared: &shared
+          - '.github/workflows/installer.yml'
+          sh:
+          - *shared
+          - 'assets/scripts/install*.sh'
+          - 'internal/cmds/generate-install.sh/install.sh.tmpl'
+          ps1:
+          - *shared
+          - 'assets/scripts/install.ps1'
+  misspell:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+    - uses: reviewdog/action-misspell@06d6a480724fa783c220081bbc22336a78dbbe82
+      with:
+        locale: US
+  test-install-sh:
+    if: ${{ needs.changes.outputs.sh == 'true' }}
+    strategy:
+      matrix:
+        os: [macos-11, ubuntu-20.04, windows-2022]
+    needs: changes
+    runs-on: ${{ matrix.os }}
+    env:
+      BINARY: ${{ matrix.os == 'windows-2022' && 'bin/chezmoi.exe' || 'bin/chezmoi' }}
+    steps:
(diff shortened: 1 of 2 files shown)


bradenhilton:
@twpayne I have a few questions if I may:

1. I added `completions/**` to `main` because the original `**/*.ps1` matched both `assets/scripts/install.ps1` and `completions/chezmoi.ps1`. Is this necessary?
2. Should I re-add `https://get.chezmoi.io/*` tests? I believe the value of them is slightly limited if we are already testing new changes to the scripts, but I suspect they will be so quick to run after this is merged that it could be worth doing regardless.

Of course, I would revert the changes I made to the install scripts before merging. I just did that to trigger the new tests.

twpayne:
> 1. I added `completions/**` to `main` because the original `**/*.ps1` matched both `assets/scripts/install.ps1` and `completions/chezmoi.ps1`. Is this necessary?

It's safe to keep it for sure. As `completions/*` is auto-generated it should only change if the code is changed, so strictly speaking it should not be necessary to add `completions/*` to the `dorny/paths-filter` step in `main.yml`. However, I would much rather err on the side of caution.



> 2\. Should I re-add `https://get.chezmoi.io/*` tests? I believe the value of them is slightly limited if we are already testing new changes to the scripts, but I suspect they will be so quick to run after this is merged that it could be worth doing regardless.

No, I think you're right to remove them. The tests you've added in this PR give better coverage anyway.

bradenhilton:
Okay, I'll revert the comments I added to the scripts and add a few more tweaks and it should be good to merge.

I think the `env.SHA` I added is correct in principle, but I'm unsure if that exact implementation is correct. The issue is I won't know until this is merged because the commit SHA is available from different places depending on if it's a merge commit or a PR commit etc.

twpayne:
> I think the `env.SHA` I added is correct in principle, but I'm unsure if that exact implementation is correct. The issue is I won't know until this is merged because the commit SHA is available from different places depending on if it's a merge commit or a PR commit etc.

That's fine. It's OK to break things on `master` from time to time. As long as everything works the next time we tag a version, I'm happy.

bradenhilton:
This line:

https://github.com/twpayne/chezmoi/pull/3584/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R51

will still cause `main` to run when `internal/cmds/generate-install.sh/install.sh.tmpl` is modified (in addition to `installer`). If you don't consider it to also be part of `code` changes, I don't have a particularly elegant solution, as the path filter does not support exclusions. It would need to be changed to something like:

```diff
- - 'internal/**'
+ - 'internal/archivetest/**'
+ - 'internal/chezmoi/**'
+ - 'internal/chezmoiassert/**'
+ - 'internal/chezmoibubbles/**'
+ - 'internal/chezmoierrors/**'
+ - 'internal/chezmoigit/**'
+ - 'internal/chezmoilog/**'
+ - 'internal/chezmoimaps/**'
+ - 'internal/chezmoitest/**'
+ - 'internal/cmd/**'
+ - 'internal/cmds/execute-template/**'
+ - 'internal/cmds/lint-commit-messages/**'
+ - 'internal/cmds/lint-txtar/**'
+ - 'internal/cmds/lint-whitespace/**'
```

twpayne:
Thanks. In the absence of a better solution, let's leave things as they are for the moment.

bradenhilton:
This was bugging me so I had another look.

Normally, I would expect that `!...` would mean "exclude `...`", but [in picomatch](https://github.com/micromatch/picomatch?tab=readme-ov-file#extglobs), my understanding is that it actually means "include everything but `...`". This requires a more specific pattern to avoid matching everything but `install.sh.tmpl`.

With this knowledge, I set up a minimal repo to test, and it seems to work as expected with

```yaml
- 'internal/**/!(install.sh.tmpl)'
```


```console
❯ tree
.
├── .github
│   └── workflows
│       ├── installer.yml
│       └── main.yml
└── internal
    └── cmds
        └── generate-install.sh
            ├── install.sh.tmpl
            ├── main.go
            └── main_test.go
```

I added a step to each workflow file to echo the matched files.

When I commit and push the `internal` directory, it gives these outputs:

```text
# main workflow
internal/cmds/generate-install.sh/main.go internal/cmds/generate-install.sh/main_test.go

# installer workflow
internal/cmds/generate-install.sh/install.sh.tmpl
```

I'm happy to open the PR if you like the change.

Issue #3598, opened 2024-02-25
Documentation contains broken links


twpayne:
## Describe the bug

When I update the documentation, I see a bunch of errors:

```
WARNING -  Doc file 'reference/templates/1password-functions/index.md' contains a relative link '../../user-guide/password-managers/1password.md#secrets-automation', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
WARNING -  Doc file 'reference/templates/1password-functions/onepassword.md' contains a relative link '../../user-guide/password-managers/1password.md#secrets-automation', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
WARNING -  Doc file 'reference/templates/1password-functions/onepasswordDetailsFields.md' contains a relative link '../../user-guide/password-managers/1password.md#secrets-automation', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
WARNING -  Doc file 'reference/templates/1password-functions/onepasswordDocument.md' contains a relative link '../../user-guide/password-managers/1password.md#1password-connect', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
WARNING -  Doc file 'reference/templates/1password-functions/onepasswordDocument.md' contains a relative link '../../user-guide/password-managers/1password.md#1password-service-accounts', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
WARNING -  Doc file 'reference/templates/1password-functions/onepasswordItemFields.md' contains a relative link '../../user-guide/password-managers/1password.md#secrets-automation', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
WARNING -  Doc file 'reference/templates/1password-functions/onepasswordRead.md' contains a relative link '../../user-guide/password-managers/1password.md#secrets-automation', but the target 'reference/user-guide/password-managers/1password.md' is not found among documentation files.
```

## To reproduce

Steps to reproduce the behavior.

## Expected behavior

I don't expect to see any errors.

## Output of command with the `--verbose` flag

n/a

## Output of `chezmoi doctor`

n/a but this is observed with the [2.47.0 release](https://github.com/twpayne/chezmoi/releases/tag/v2.47.0).

## Additional context

None.

bradenhilton:
[...]

PR #3600, opened 2024-02-26
chore: Fix broken relative links

bradenhilton:
Fixes https://github.com/twpayne/chezmoi/issues/3598.

Without these changes, I can reproduce the warnings mentioned in https://github.com/twpayne/chezmoi/issues/3598.

The warnings mentioned in https://github.com/twpayne/chezmoi/actions/runs/8043964297/job/21966863415#step:5:30 appear to be mainly concerning the templated pages.

The `hooks.py` change should resolve the `package internal/cmds/execute-template is not in std (/opt/hostedtoolcache/go/1.22.0/x64/src/internal/cmds/execute-template)` output.


twpayne (approved):
Thank you very much!

Issue #3603, opened 2024-02-27
splitList and quoteList don't work in pair


ErrrorMaxx:
## Describe the bug

`splitList` and `quoteList` don't work in pair

## To reproduce

```console
$ chezmoi execute-template '{{ "a b" | splitList " " | quoteList }}'
```

## Expected behavior

This template should not fail and return list of quoted strings

## Output of command with the `--verbose` flag

```console
$ chezmoi --verbose execute-template '{{ "a b" | splitList " " | quoteList }}'
chezmoi: template: arg1:1:27: executing "arg1" at <quoteList>: wrong type for value; expected []interface {}; got []string
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT    CHECK                       MESSAGE
ok        version                     v2.47.0, commit 39bd915f446068862cc3064edb6dbeee795785ca, built at 2024-02-25T20:56:37Z, built by goreleaser
failed    latest-version              Get "https://api.github.com/repos/twpayne/chezmoi/releases/latest": dial tcp: lookup api.github.com on [::1]:53: read udp [::1]:45512->[::1]:53: read: connection refused
ok        os-arch                     android/arm64
ok        uname                       Linux localhost 4.19.191+ #1 SMP PREEMPT Wed Nov 29 21:44:22 CST 2023 aarch64 Android
ok        go-version                  go1.22.0 (gc)
ok        executable                  ~/.local/bin/chezmoi
ok        upgrade-method              replace-executable
ok        config-file                 ~/.config/chezmoi/chezmoi.yaml, last modified 2024-02-27T08:39:04Z
warning   source-dir                  ~/.local/share/chezmoi is a git working tree (dirty)
warning   suspicious-entries          ~/.local/share/chezmoi/home/dot_config/rootmoi/chezmoi.yaml.tmpl
warning   working-tree                ~/.local/share/chezmoi is a git working tree (dirty)
warning   dest-dir                    ~ is a git working tree (dirty)
warning   umask                       077
ok        cd-command                  found /data/data/com.termux/files/usr/bin/zsh
ok        cd-args                     /data/data/com.termux/files/usr/bin/zsh
info      diff-command                not set
ok        edit-command                found /data/data/com.termux/files/usr/bin/micro
ok        edit-args                   micro
ok        git-command                 found /data/data/com.termux/files/usr/bin/git, version 2.44.0
warning   merge-command               vimdiff not found in $PATH
[...]

PR #3604, opened 2024-02-27
fix: Make splitList return []any

bradenhilton:
Fixes https://github.com/twpayne/chezmoi/issues/3603.

### DIFF
--- a/internal/cmd/templatefuncs.go
+++ b/internal/cmd/templatefuncs.go
@@ -482,6 +482,15 @@ func (c *Config) setValueAtPathTemplateFunc(path, value, dict any) any {
 	return result
 }
 
+func (c *Config) splitListTemplateFunc(sep, s string) []any {
+	strSlice := strings.Split(s, sep)
+	result := make([]interface{}, len(strSlice))
+	for i, v := range strSlice {
+		result[i] = v
+	}
+	return result
+}
+
 func (c *Config) statTemplateFunc(name string) any {
 	switch fileInfo, err := c.fileSystem.Stat(name); {
 	case err == nil:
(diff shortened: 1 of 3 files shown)


halostatue (approved):

internal/cmd/templatefuncs.go
```diff
@@ -482,6 +482,15 @@ func (c *Config) setValueAtPathTemplateFunc(path, value, dict any) any {
 	return result
 }
 
+func (c *Config) splitListTemplateFunc(sep, s string) []interface{} {
```
twpayne, line 485: nit: Please use `any` instead of `interface{}`.

internal/cmd/testdata/scripts/issue3603.txtar
```diff
@@ -0,0 +1,2 @@
+exec chezmoi execute-template '{{ "a b" | splitList " " | quoteList }}'
```
twpayne, line 1: Please add a comment explaining what is being tested. testscript uses these comments when generating its output.

Also, this test can be moved to `templatefuncs.txtar`.

internal/cmd/templatefuncs.go
bradenhilton: Should I also make this change in the commit message?

twpayne: Yes please :)

bradenhilton:
@twpayne Is the solution in [this playground](https://go.dev/play/p/zexqWRQzLCg) better?

twpayne:
Yes, that's much more flexible :)

bradenhilton:
Ahh. I didn't realize that methods can't use generics. The code in that playground isn't applicable here.

Issue #3637, opened 2024-03-08
Improve automatic dependency updates


twpayne:
## Is your feature request related to a problem? Please describe.

I'm always frustrated when I have to update chezmoi's dependencies as it requires a lot of manual work:

* I can update most Go modules in `go.mod` with `go get -u -t ./...`, but I then need to run `make test-release` to build chezmoi for all platforms to update `go.sum`, as long as these are only minor version bumps.
* Dependabot currently opens per-Go-module PRs which a lot of PR noise.
* I check for major Go module version bumps with [gomajor](https://github.com/icholy/gomajor), but dealing with the output of this sometimes requires additional work.
* I have a private script for updating GitHub Actions versions.
* For [tool versions](https://github.com/twpayne/chezmoi/blob/c47dcfaa6efb9ece6df390fda4ae51c6a31039b9/.github/workflows/main.yml#L14-L24), I currently have to manually check each one of them.
* Very rarely, I check Python requirements (in chezmoi's various `requirements*.txt` files).

## Describe the solution you'd like

I would like an automatic tool to propose dependency high-level dependency update PRs that update related dependencies together, e.g. bumping all minor Go module version dependencies in the same PR.

## Describe alternatives you've considered

Keeping this as-is, or writing some tooling myself.

## Additional context

* I'd love help with this.
* It's not necessary to solve all of the above problems in one go.
* I think that switching from [dependabot](https://github.com/twpayne/chezmoi/blob/c47dcfaa6efb9ece6df390fda4ae51c6a31039b9/.github/dependabot.yml) to [renovatebot](https://docs.renovatebot.com/) might help.
* If you want to work on this, please say so in a comment on this issue so people don't end up duplicating work.

halostatue:
I tried and moved back from Renovate; I didn't like it — I should consider giving it another try.

Dependabot grouping is somewhat useful, and there is the ability to use `update-types` to ensure that `minor` or `patch` items are upgraded as one PR.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

icholy:
@twpayne `gomajor list` takes an optional `-json` flag.

twpayne:
Thanks! The issue with bumping major versions is that it requires modifying the source files (i.e. at least updating `import` statements with the new package path, and occasionally modifying code for the new API). However, the `-json` flag to `gomajor list` should make it easier to write some basic automation around this, thank you!

[...]

PR #3641, opened 2024-03-10
chore: Group dependabot updates

twpayne:
Refs #3637, as suggested by @halostatue.

Not sure how to test this, except waiting for the next dependabot run. It should be harmless, though.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,20 +1,46 @@
 version: 2
 updates:
-- package-ecosystem: gomod
-  directory: /
-  schedule:
-    interval: monthly
-  labels:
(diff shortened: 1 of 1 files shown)


halostatue:
Yeah. The Dependabot grouping documentation isn't ideal, but simply specifying `patterns` won't do what you want.

I’ve amended your commit to have four groups: go-dev, go-prod, actions, and python. I have restricted the grouping to *only* apply for patch and minor upgrades. That way, if mkdocs were to go from 9 to 10, it would not be grouped together.

halostatue (approved):

twpayne:
Awesome, thank you!

PR #3651, opened 2024-03-16
fix: pass the `--needed` flag to pacman

aarondill:
This ensures that pacman doesn't re-install an already up-to-date package

### DIFF
--- a/internal/cmd/upgradecmd_unix.go
+++ b/internal/cmd/upgradecmd_unix.go
@@ -109,7 +109,7 @@ func (c *Config) upgradeUNIXPackage(
 			if useSudo {
 				args = append(args, "sudo")
 			}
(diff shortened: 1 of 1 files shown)


halostatue (commented):
The description covers what this does nicely, but I don't use Arch to be able to test this myself.

aarondill:
here's a quote from the arch wiki describing the flag.
```
Note: If a package in the list is already installed on the system, it will be reinstalled even if it is already up-to-date. This behavior can be overridden with the --needed option.
```

with this change, chezmoi will still invoke pacman when the package is out of date with the github releases (it usually is), but pacman will just warn the user that the package is already up to date with the arch repos and do nothing

twpayne:
Thank you very much!

PR #3667, opened 2024-03-29
chore: Add infrastructure for vendored dependencies

twpayne:


twpayne:
Given events like https://www.openwall.com/lists/oss-security/2024/03/29/4, I'm not sure if chezmoi should vendor dependencies or not.

chezmoi is a particularly high-value target as it runs local user on many systems inside many organizations.

One reason for vendoring dependencies is that all the code that goes into your binary is now present in the source repo, which potentially makes it easier to review changes to dependencies.

On the flip side, vendoring dependencies means that diffs become large, and the Go team [already actively addresses these problems](https://go.dev/doc/security/vuln/) on a scale way beyond what any small open source project can achieve.

My gut feeling is that chezmoi should continue to rely on Go's security infrastructure and that this PR is not necessary.

Feedback very welcome!

twpayne:
Note that the follow-up PR would run `go mod vendor` and add all dependencies directly to chezmoi's source repo.

twpayne:
> If I remember correctly, wasn't this one of the things that would have eventually been required for Debian packaging?

Actually, Debian wants the opposite, that is for every Go module to be independently packaged. What Debian wants is so completely unworkable that I've given up trying to help them.



> I’m…mildly negative on vendoring as there will be diffs for code which we as reviewers _will not understand_.

Fair point. I think you're right. I'll close this PR.

PR #3732, opened 2024-04-29
feat: Add gitHubLatestReleaseAssetURL template function

### LINKED ISSUE #3725: Add a `gitHubLatestReleaseAsset` function

Whenever I use the `githubLatestRelease` function I need to do some additional processing in order to make use of the return value. Most commonly I end up doing something like this:

```yaml
---
.local/bin/chezmoi:
  type: 'file'
  executable: true
  url: '{{ (gitHubLatestRelease "twpayne/chezmoi").HTMLURL | replace "/releases/tag/" "/releases/download/" }}/chezmoi-{{ .chezmoi.os }}-{{ .chezmoi.arch }}'
```

This works but isn't as readable as I'd like.

A new `githubLatestReleaseAsset` function which can be used to select an asset with a given name (or perhaps matching a glob) from the latest release. For example:

```yaml
.local/bin/chezmoi:
  type: 'file'
  executable: true
  url: '{{ gitHubLatestReleaseAsset "twpayne/chezmoi" (printf "chezmoi-%s-%s" .chezmoi.os .chezmoi.arch ) }}'
```

Alternatively I could define my own helper template functions, which would be nice for other reasons.

twpayne:
Fixes #3725.


assets/chezmoi.io/docs/reference/templates/github-functions/gitHubLatestReleaseAssetURL.md
```diff
@@ -0,0 +1,20 @@
...
+release about the given *owner-repo*, returning structured data as defined by
+the [GitHub Go API
+bindings](https://pkg.go.dev/github.com/google/go-github/v61/github#RepositoryRelease).
+It then iterates through all the release's assets, returning the first one that
+matches *pattern*. *pattern* is a shell pattern as [described in
+`path.Match`](https://pkg.go.dev/path#Match).
+
+Calls to `gitHubLatestReleaseAssetURL` are cached so calling
+`gitHubLatestReleaseAssetURL` with the same *owner-repo* will only result in one
+call to the GitHub API.
```
halostatue, line 13: The examples use `.chezmoi.os` and `.chezmoi.arch`, but not all releases use the go values presented by that. I wonder if it might be worth using a third example for a tool like `ripgrep`, written in Rust and therefore using the  triples (https://github.com/BurntSushi/ripgrep/releases/tag/14.1.0, e.g., aarch64-apple-darwin).

halostatue (approved):

assets/chezmoi.io/docs/reference/templates/github-functions/gitHubLatestReleaseAssetURL.md
twpayne: Good point. Also, Debian and RedHat use different architecture and OS values to `GOARCH` and `GOOS`. I wonder if it's worth adding a few extra standard template variables like `.chezmoi.rpmArch`, `.chezmoi.debArch`, `.chezmoi.rustOS`, and `.chezmoi.rustArch`.

halostatue: If you’re going to do `rustOS`, you probably want to just do `rustArchTriple` since that’s the typical approach.

PR #3735, opened 2024-04-29
feat: Add package architecture variables

twpayne:
Refs #3732.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -262,6 +262,7 @@ type templateData struct {
 	commandDir        chezmoi.AbsPath
 	config            map[string]any
 	configFile        chezmoi.AbsPath
+	debArch           string
 	executable        chezmoi.AbsPath
 	fqdnHostname      string
 	gid               string
@@ -312,6 +313,13 @@ var (
 		"include":    chezmoi.EntryTypeSetFlagCompletionFunc,
 		"path-style": chezmoi.PathStyleFlagCompletionFunc,
 		"secrets":    severityFlagCompletionFunc,
+	}
+
+	// debArches maps runtime.GOARCH architectures to Debian package
+	// architectures.
+	debArches = map[string]string{
+		"386": "i386",
+		"arm": "armel",
 	}
 )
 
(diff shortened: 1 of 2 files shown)


halostatue (approved):

VorpalBlade:
What about Arch Linux?

VorpalBlade:
> > What about Arch Linux?
> 
> As far as I can tell, Arch Linux only supports `x86_64`, so `archArch` would be the same as `rpmArch`.

That is true for upstream official arch. But there are many derivatives such as ArchLinux32, ArchLinux ARM etc.

twpayne:
Is there a list of Arch architectures somewhere? I did quite a bit of searching on the Arch website and wiki but didn't find any.

halostatue:
Alpine has support for:

| arch | since | description |
| - |- | -|
| x86_64 | all | Widely spread AMD64 compatible 64-bit x86 instruction set. |
|x86 | all | i686 (or later) CPUs with at least CMOV and SSE|
|ppc64le | v3.6 | For PowerPC devices with pure little-endian mode, mostly for POWER8 and POWER9|
|armhf | v3.0 | A 32-bit ARM instruction set with hard-float point extension.|
|armv7 | v3.9 | The 32-bit ARM only execution state of the ARMv7 devices machines.|
|aarch64 | v3.5 | The 64-bit ARM only execution state of the ARMv8+ device machines.|
|s390x | v3.6 | For IBM mainframes, especially IBM Z and IBM LinuxONE servers.|

VorpalBlade:
> Is there a list of Arch architectures somewhere? I did quite a bit of searching on the Arch website and wiki but didn't find any.

Good question, I don't know that there is. In particular Arch uses the concept of a `CARCH` variable, that defines package compatibility.

There was some discussion around if v2/v3 levels of x86-64 [micro architectures](https://en.m.wikipedia.org/wiki/X86-64#Microarchitecture_levels) would be in the CARCH or not (there is no official optimised version for more recent processors yet, though it has been discussed quite a bit).

On Arch 32 the valid values are at https://archlinux32.org/architecture/ (i486, i686 and pentium4).

Scrolling through https://archlinuxarm.org/packages shows armv7h, aarch64 and of course "any", which means that it is a platform independent package (data file, scripting language, etc). I believe Debian also have some equivalent to "any" packages.

twpayne:
Thanks all for your input here. Much appreciated!

On further reflection, I don't think it's right for chezmoi to add these `debArch`, `rpmArch`, and `archArch` package architecture template variables for the following reasons:

* They're complex to get right.
* They pollute the `.chezmoi` template variable namespace with Linux distribution specific variable names, which are irrelevant for all non-Linux systems.
* It's trivially easy for the user to set an exact package architecture for the local machine in chezmoi's config file or by using a config file template.

Therefore, I'm closing this PR.

PR #3748, opened 2024-05-02
feat: Support self-hosted GitHub Enterprise servers

### LINKED ISSUE #3724: Support self-hosted GitHub Enterprise

I want to use Chezmoi to install various binaries from a self-hosted GitHub Enterprise installation. I am already using `chezmoiexternals` + the `githubLatestRelease` function to install binaries from [github.com](https://github.com), for example:

```
---
.local/bin/chezmoi:
  type: 'file'
  executable: true
  url: '{{ (gitHubLatestRelease "twpayne/chezmoi").HTMLURL | replace "/releases/tag/" "/releases/download/" }}/chezmoi-{{ .chezmoi.os }}-{{ .chezmoi.arch }}'
```

But this doesn't work for self-hosted GitHub Enterprise as I don't believe there is a way to pass in a custom hostname. Similarly there isn't a way to pass it an auth token (there is `$CHEZMOI_GITHUB_ACCESS_TOKEN` but I would need a way to provide a token that is scoped to a single domain).

Either a new family of functions (e.g. `githubEnterpriseLatestRelease`) or extending the existing GitHub functions to accept a hostname (e.g. `githubLatestRelease "https://git.example.com/owner/repo"`). Additionally, a way to provide the auth token for `git.example.com` in `chezmoi.yaml`.

Previously I was just using a script to do this:

```
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

{{ $version := output "gh" "release" "--repo=git.example.com/owner/repo" "view" "--json=name" "--template={{.name}}" -}}
gh release --repo git.example.com/owner/repo download --output=- --pattern=example_{{ .chezmoi.os }}_{{ .chezmoi.arch }}.tar.gz {{ $version }} | tar --extract --file=- --gunzip --directory ~/.local/bin example
```

I have since changed to using a template since I want Chezmoi to purge unmanaged files from `~/.local/bin`.

```
# dot_local/bin/example
{{ $version := output "gh" "--repo=git.example.com/owner/repo" "release" "view" "--json=name" "--jq=.name" | trim -}}
{{ output "sh" "-c" (printf "gh --repo=git.example.com/owner/repo release download --output=- --pattern=example_%s_%s.tar.gz %s | tar --extract --to-stdout --gunzip example" .chezmoi.os .chezmoi.arch $version) }}
```

This works okay, but seems a bit hacky and is definitely less readable.

twpayne:
Fixes #3724.

@joshuaspence would you be able to test this?


joshuaspence:
It doesn't seem to work, unless I am missing something.

```
$ chezmoi --version
chezmoi version dev, commit 95f58622290fda3494bf829c7cb5c3a6ea2b7c29, built at 2024-05-02T21:17:02Z

$ chezmoi execute-template '{{ gitHubLatestRelease "REDACTED_HOST/REDACTED_OWNER/REDACTED_REPO" }}'
chezmoi: template: arg1:1:3: executing "arg1" at <gitHubLatestRelease "REDACTED_HOST/REDACTED_OWNER/REDACTED_REPO">: error calling gitHubLatestRelease: GET https://api.github.com/repos/REDACTED_OWNER/REDACTED_REPO/releases/latest: 401 Bad credentials []
```

It seems like it is ignoring the host.

joshuaspence:
It does seem to be using the correct access token though, just passing it to the wrong endpoint

twpayne:
Doh! I forgot to set the URLs. Thanks for testing so quickly. PR updated. Please could you try again (I don't have a server to test against).

twpayne:
> Thanks, it works now!

Great! Thanks for testing!

> Is it possible to have the auth token in a config file rather than polluting my environment? Not a big deal

It's technically possible but I'd want to avoid storing secrets in chezmoi's config file. I'd like to think this through and get @halostatue's opinion before adding such a feature.

> I am seeing a different error now but I think this one is my fault:
> 
> ```
> ± ~/workspace/chezmoi/chezmoi diff
> chezmoi: gzip: invalid header
> ```

Maybe. This looks like the returned archive is not in the expected format, or maybe there's an HTTP redirect happening. Could you inspect the contents of the returned data?

joshuaspence:
Yeah I'm just checking that now. When I download the archive through the browser and use `file` on it, it shows `gzip compressed data, max compression, original size modulo 2^32 87956992`. Trying to work out if `chezmoi` is seeing different data or if there is something else going on

twpayne:
If you run chezmoi with the `--debug` flag I think it will show you the first few bytes of each HTTP download.

joshuaspence:
Ah I see the issue. It's because I am doing this:

```
url: '{{ (gitHubLatestRelease "REDACTED").HTMLURL | replace "/releases/tag/" "/releases/download/" }}/REDACTED_{{ .chezmoi.os }}_{{ .chezmoi.arch | replace "amd64" "x86_64" }}.tar.gz'
```

Let me try using `gitHubLatestReleaseAssetURL`

joshuaspence:
Nah it doesn't work with `gitHubLatestReleaseAssetURL`. I guess because whatever is processing externals is just using a generic HTTP client, which won't be injecting the GitHub auth token

twpayne:
Ah, I see, thank you.

Can you embed the access token into the external URL, something like (untested):

```
["my/external"]
    url = "https://git.example.com:{{ env "CHEZMOI_GIT_EXAMPLE_COM_ACCESS_TOKEN" }}/path/to/archive"
```

joshuaspence:
Nah it doesn't work. GitHub doesn't allow basic auth, you have to pass in an `Authorization: Bearer xxxxxxx` header instead

twpayne:
OK, thank you again for testing.

I suspect that the right solution is that chezmoi should add `Authentication: Bearer $CHEZMOI_GIT_EXAMPLE_COM_ACCESS_TOKEN` headers to all requests to `git.example.com`, but this will require some thought and refactoring.

assets/chezmoi.io/docs/reference/templates/github-functions/gitHubLatestRelease.md
```diff
@@ -1,12 +1,12 @@
-# `gitHubLatestRelease` *owner-repo*
+# `gitHubLatestRelease` *host-owner-repo*
```
halostatue, line 1: We may want to show this as *either `*[host-]owner-repo*` or `*owner-repo*|*host-howner-repo*`.

This applies to all `github*` functions.

assets/chezmoi.io/docs/reference/templates/github-functions/gitHubLatestRelease.md
```diff
@@ -1,12 +1,12 @@
...
-API
+the given *host-owner-repo*, returning structured data as defined by the [GitHub
+Go API
 bindings](https://pkg.go.dev/github.com/google/go-github/v57/github#RepositoryRelease).
 
 Calls to `gitHubLatestRelease` are cached so calling `gitHubLatestRelease` with
-the same *owner-repo* will only result in one call to the GitHub API.
+the same *host-owner-repo* will only result in one call to the GitHub API.
 
 !!! example
```
halostatue, line 11: Adding an example for `host-owner-repo` in addition to `owner-repo` would be good.

This applies to all `github*` functions.

halostatue (commented):
LGTM.

It *might* be worthwhile to see about being able to use [`gh auth token`](https://cli.github.com/manual/gh_auth_token) if that’s installed, but that’s a separate feature request. It would *minimize* the need for using `CHEZMOI_GITHUB_ACCESS_TOKEN` etc.

felipecrs:
> It _might_ be worthwhile to see about being able to use [`gh auth token`](https://cli.github.com/manual/gh_auth_token) if that’s installed, but that’s a separate feature request. It would _minimize_ the need for using `CHEZMOI_GITHUB_ACCESS_TOKEN` etc.

chezmoi could simply try reading the token from `~/.config/gh/hosts.yml` instead.

```yaml
github.com:
    users:
        felipecrs:
            oauth_token: gho_supersecrettoken
    git_protocol: https
    oauth_token: gho_supersecrettoken
    user: felipecrs
```

halostatue:
> > It _might_ be worthwhile to see about being able to use [`gh auth token`](https://cli.github.com/manual/gh_auth_token) if that’s installed, but that’s a separate feature request. It would _minimize_ the need for using `CHEZMOI_GITHUB_ACCESS_TOKEN` etc.
> 
> chezmoi could simply try reading the token from `~/.config/gh/hosts.yml` instead.
> 
> ```yaml
> github.com:
>     users:
>         felipecrs:
>             oauth_token: gho_supersecrettoken
>     git_protocol: https
>     oauth_token: gho_supersecrettoken
>     user: felipecrs
> ```

That only works if you have used `--insecure-storage        Save authentication credentials in plain text instead of credential store` during login.

```yaml
github.com:
    git_protocol: https
    users:
        halostatue:
    user: halostatue
```

There might be some possible extensions to be done in order to execute `gh auth token` or `gh auth token --user <username>`. I don't have a GHE instance to test with to see how _that_ part might work.

felipecrs:
> That only works if you have used `--insecure-storage Save authentication credentials in plain text instead of credential store` during login.

Got it. By the way, I had not used `--insecure-storage` but I suppose it's because I'm on WSL which has no graphical keyring installed by default.

twpayne:
Closing this PR as it is very out-of-date.

joshuaspence:
@twpayne does that mean you aren't interested in such a feature?

joshuaspence:
Oh sorry, this is the PR. Not the issue

twpayne:
I would love for chezmoi to support self-hosted GitHub Enterprises servers. It needs a different solution to this PR, though.

PR #3762, opened 2024-05-13
chore: switch to Go for generate-commit

bradenhilton:
This should enable Windows users to create a test release with goreleaser without needing to comment out the script line in the hooks section.


twpayne (approved):
Thank you!

bradenhilton:
It's a separate issue, but I don't know what to do about this

```console
$ golangci-lint run
internal\chezmoi\realsystem_windows.go:16:2: field `scriptEnv` is unused (unused)
        scriptEnv               []string
        ^
```

outside of just disabling `unused` on `internal\chezmoi\realsystem_windows.go`, which seems excessive.

twpayne:
You can add a `//nolint:unused` comment to this line.

twpayne:
Actually, it turns out that `scriptEnv` is indeed not used at all. #3763 extends chezmoi's CI to run golangci-lint on all OSes to catch these sorts of issues in the future.

Issue #3610, opened 2024-02-29
`chezmoi remove` removes the whole directory instead of removing a target


avinal:
## Describe the bug

When adding `chezmoi add ~/.config/x/y` it adds `.config`, `.config/x` and `.config/x/y` in the managed list. I would expect only the last one to be added. 

So when I list, it shows all 3
```
$ chezmoi managed                                                                                                                                                                                                                                                                                                     
.config                                                                                                                                                                                                                                                                                                               
.config/atuin                                                                                                                                                                                                                                                                                                         
.config/atuin/config.toml 
```

and I removed all of them using `chezmoi remove ...` because I though it must be a mistake since I added only one file. I would expect chezmoi to stop watching them. But instead it removed the whole `.config` directory. 

Now thanks to this project, I am even unable to copy and paste even because of a missing configuration and there is no way to restore my `.config` dir. The missing config is going to take my valuable time to redo and restore. I dont know if that is even possible.

avinal:
Great project, but I would expect some failsafe for such and trivial mistake, and I am sure I am not the only one who fell to this trap.

bare git dotfiles is the best. Never going to use this project ever again.

bradenhilton:
I really don't appreciate your tone.

> The missing config is going to take my valuable time to redo and restore.

Your time is no more or less valuable than anyone else's, including the maintainers of this project.

You couldn't even be bothered to mention the method you used to remove "them". Do you always open such low effort issues? You're never going to get help with an attitude like that.

avinal:
Hey @bradenhilton, apologies, but you won't expect anyone in a great mood when they have lost their whole config directory because it is so easy to remove using this tool without any failsafe. There are previous known instances of this (refer #3182)

> Do you always open such low effort issues?

No, usually I am very generous while opening issues and helping projects. Couldn't help this time. 

I have updated the description for more context.

Issue #3182, opened 2023-08-18
Foolproof `chezmoi remove`


Azorlogh:
## Is your feature request related to a problem? Please describe.

Might be somewhat controversial, but `chezmoi` makes it too easy to accidentally remove a directory full of content, because people might mix up `chezmoi remove` with `chezmoi forget`.
 
I know that `chezmoi remove` does show the paths that will get removed, and one should be careful about using a command with `remove` anyway, but I think the cost of losing files you care about is important enough to warrant making the command more foolproof.

## Describe the solution you'd like

Renaming `chezmoi remove` to `chezmoi remove-permanently` or similar.
Making `chezmoi remove` only return a message that approximately says: "Use `chezmoi forget` to remove a target from the source state. Use `chezmoi remove-permanently` to delete the actual item"

## Describe alternatives you've considered

Showing a red warning when executing `chezmoi remove` to make it clear that it deletes actual files
A clear and concise description of any alternative solutions or features you've
considered.

I understand that my proposed solution is a breaking change hence why I think it might be controversial. But I think it's worth it, as it's more foolproof than a warning (at the point where the warning would be shown, the user is only one keypress away from deleting the file/directory after all).

twpayne:
I'm sorry that you deleted your files, but no to this. `chezmoi remove` behaves like `rm` by design. As you point out, it already includes multiple layers of protection against accidental misuse. In the worst case you should be able to restore the source file from your git repo.

PR #3804, opened 2024-06-09
feat: Rename remove command to destroy

### LINKED ISSUE #3803: Make accidental removal harder

I tried to add a directory to see how well it would be handled. The files had many subdirectories and files, and I was using Windows Powershell, so at one point it failed because the path was too long, so adding failed

I edited a file, and I got a bunch of warnings related to carriage returns due to the files that did get added before failing. I looked up which command would remove the files from chezmoi, so I typed 'chezmoi remove <directory>'

Voici the accidentally removed files

There was no path display, warning or prompt. I did not use the --force flag. In fact, since I have autoPush = true, the only output described pushing the .chezmoiignore file which confused me for a moment since my input did not have that file included, but I figured that chezmoi detected the file had been edited externally and pushed even if my command had nothing to do with that file. The only other option I have set is my editor

In the other issues I read at https://github.com/twpayne/chezmoi/issues/3610 and https://github.com/twpayne/chezmoi/issues/3182, multiple failsafes are mentioned, but I didn't get any safety net. I also read mentions of a .chezmoiremove file, but I don't have that file either

## Describe the solution you'd like. Describes what alternatives you have considered

Before testing, I read the Quick Start and User Guide. I do not recall any mention of file removal or the remove / purge / forget commands in them. I'd add to the User Guide section Daily Operations a subject related to file removal and the three commands

I understand that remove was modelled after rm, but I'm not sure that following it closely was the best idea:
- There is no 'add' in a shell, but that's ok because a dotfile manager is not a shell. If you look at this through the language, remove is the antonym of add, but functionally the opposite of the add command is the forget command, so it can be expected that intuitively users will think remove will leave the source file intact and will only remove the file from chezmoi
- Given that almost everything in chezmoi and the website is so well done, and chezmoi essentially creates backups, users can be lured into quickly trusting the tool. The dangers of rm are legendary, but I don't think that infamy is something chezmoi strives to replicate. I don't think the people who got burnt by remove are dumb
- One of the issues suggested changing the name of remove to remove-permanently. A rename sounds a reasonable though I'd prefer something that refers more to how it deletes both the source and the destination, since that is the actual function, maybe full-remove ?
- If this problem is ignored, more people will accidentally delete their files then create an issue about it. If anything, making it harder to unintentionally remove will reduce the issues that need a reply
- I do understand many users are used to remove, but improving usability is just a growing pain for a widely-used tool

So I have three suggestions:
- Add remove/purge/forget to the User Guide
- Change remove to full-remove
- Seal whatever usability hole allowed me to use remove without feedback from the tool

## Additional context

Console output of remove
chezmoi

twpayne:
Fixes #3803.
Refs #3610.
Refs #3182.


halostatue (approved):
I think that this is a very good change.

PR #3809, opened 2024-06-11
chore: lint: don't assume origin (e.g. after $(gh repo fork))

donaldguy:
Just a little meta thing to make it easier to have done as the html comment requests


halostatue:
Seems sensible.

PR #3810, opened 2024-06-11
chore: add test to ensure config struct tags match across marshals

donaldguy:
- add tags to interpreter.go so it passes

This runs ahead of attempts to add `chezmoi dump-config -f toml` 

(and possibly a version of a default config/config-template from that with some documentation comments)

### DIFF
--- a/internal/chezmoi/interpreter.go
+++ b/internal/chezmoi/interpreter.go
@@ -7,8 +7,8 @@ import (
 
 // An Interpreter interprets scripts.
 type Interpreter struct {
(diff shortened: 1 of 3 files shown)


halostatue:
Looks pretty reasonable to me.

twpayne:
Thank you very much!

twpayne:
Quick note to say that this check just caught a bug in a PR I'm working on. Catching this bug early means that it's easy and quick for me to fix. Thank you!

Issue #3815, opened 2024-06-17
Documentation update "Why do I get a blank buffer or empty file when running chezmoi edit"


SjoerdV:
Proposed update for:
https://www.chezmoi.io/user-guide/frequently-asked-questions/troubleshooting/#why-do-i-get-a-blank-buffer-or-empty-file-when-running-chezmoi-edit

Maybe you could add a tab (next to VIM/VSCode tabs) to get **Any editor** (also the ones that do not wait in the foreground) to work with `chezmoi edit`

1. Add a new bash program somewhere in your $PATH (in this case I added `$HOME/.local/bin/mousepad-wait`)
```bash
#!/usr/bin/env bash
wait -n; mousepad "$@"
```
1. Run `chmod +x $HOME/.local/bin/mousepad-wait` to make the program executable
1. Finally add to (or update) `chezmoi.toml`
```toml
[edit]
    command = "mousepad-wait"
    args = ["--opening-mode=window", "--encoding=UTF-8"]
```

halostatue:
A pull request with this information would be considered.

twpayne:
Closing this because the associated PR is closed.

PR #3819, opened 2024-06-17
docs: Update troubleshooting.md for editors without wait mode

SjoerdV:
Fixes https://github.com/twpayne/chezmoi/issues/3815.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/troubleshooting.md
+++ b/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/troubleshooting.md
@@ -81,6 +81,30 @@ remove the temporary directory.
     variable to `["--wait"]` or by setting the `EDITOR` environment variable to
     include the `--wait` flag, e.g. `export EDITOR="code --wait"`.
 
+=== "Any editor"
+
+    In this example 'mousepad' is the editor to be used. This program will not wait in the foreground by default. It also requires DBUS server integration to be disabled and it needs to start a new window for each execution (otherwise new PIDs will not be created for the process, if another mousepad instance is already running)
+
+    1. Add a new bash script somewhere in your $PATH (like here: `$HOME/.local/bin/chezmoi-mousepad`)
+
+    ```bash
+    #!/usr/bin/env bash
+    wait -n; mousepad --opening-mode=window --encoding=UTF-8 --disable-server "$@"
+    ```
+
+    The key here is the `wait -n` statement that will wait for the next child job to complete and can be universally applied (within the context of a bash script) to any editor you choose.
+
+    2. Run `chmod +x $HOME/.local/bin/chezmoi-mousepad` to make the script executable
+    3. Finally add to (or update) `chezmoi.toml`
+    
+    ```toml
+    [edit]
+        command = "chezmoi-mousepad"
+        args = []
+    ```
+
+    or by setting the `EDITOR` environment variable, e.g. `export EDITOR="chezmoi-mousepad"`.
+
 The "bit of magic" that `chezmoi edit` performs includes:
 
 * `chezmoi edit` makes the filename opened by your editor more closely match


assets/chezmoi.io/docs/user-guide/frequently-asked-questions/troubleshooting.md
```diff
@@ -81,6 +81,24 @@ remove the temporary directory.
...
+    ```
+    2. Run `chmod +x $HOME/.local/bin/mousepad-wait` to make the program executable
+    3. Finally add to (or update) `chezmoi.toml`
+    ```toml
+    [edit]
+        command = "mousepad-wait"
+        args = ["--opening-mode=window", "--encoding=UTF-8","--disable-server"]
+    ```
+    or by setting the `EDITOR` environment variable, e.g. `export EDITOR="mousepad-wait --opening-mode=window --encoding=UTF-8 --disable-server"`.
+
```
halostatue, line 101: I prefer `shell script` or `bash script` over `bash program`. Given the use of `-n`, this is *absolutely* a `bash script`. It *may* be worth mentioning that `wait -n` will essentially wait for the next child job to complete, if I am understanding this correctly.

There should be blank lines between the code blocks and the paragraphs to which they are attached.

I think that `--opening-mode=window…` should be in the shell-script (which would probably be better named as `chezmoi-mousepad`).

Beyond this, something feels *missing* from this — it’s good text, but it still feels like "mousepad" instructions not "how to make this work with any editor" using "mousepad" as an example.

SjoerdV: @halostatue, hope you like the changes

halostatue: I think that this is clearer and more easily generalized from this example. I’ll leave it to @twpayne to make the final decision on this one. The only issue I see remaining is the term "any editor" as the title, but I don't have a better suggestion. The Vim and VS Code examples are good examples for any editor with a "foreground" or "wait" command-line parameter, but this one is one for editors with a "we only work in the background" approach.

halostatue (approved):

halostatue:
One other thing: you will need to squash your commits and retitle them per the [contribution](https://www.chezmoi.io/developer-guide/contributing-changes/) guidelines, as we use conventional commits. This would be something like `docs: Update troubleshooting guide for editors without wait modes` (that’s probably longer than 50 characters, which is the typical max length for the message).

assets/chezmoi.io/docs/user-guide/frequently-asked-questions/troubleshooting.md
SjoerdV: Ok great! Probably the desired length of the tab name restricts more elaborate naming options.

SjoerdV:
> This is a valuable addition, but I personally have a few issues with it at the moment.
> 
>     1. You're intermingling the concept (using an intermediate script which invokes an `$EDITOR` in a manner that provides wait functionality) with the example. I think it would be better to explain the concept first, as generally as possible, and then provide an example.
> 
>     2. The section title is "Any editor", but the solution contained within uses a Bash-specific built-in. Obviously, this requires Bash, so it excludes all (non-WSL + Bash) Windows users at a minimum. Perhaps something like "Background-only editors" would be better? Note that I am terrible at naming things.
> 
>     3. The example mentions editing `chezmoi.toml`. I understand why this was done, but I feel it is unnecessarily specific, and _could potentially_ leave a user with the impression that the solution would only work if their config file is in the TOML format, or lead them to create a `chezmoi.toml` file even if they already have an existing config in a different format. I would instead simply replace this with something like "Set the `edit.command` configuration variable to the new shell script", and then provide the example in TOML if needed.
> 
> 
> Explaining the concept separately from the example would also allow us to add examples in the future (using different commands, using different shells/platforms etc.) if needed. I wouldn't be surprised if there was an equivalent cross-platform Go/Rust tool for this somewhere.
> 
> +1 to @halostatue's points regarding commits/message. We use the conventional commit categories to exclude all of the non-user-facing/miscellaneous changes from the changelog when a new release is cut.

@bradenhilton All valid points, but... I only saw your update after it was already approved and I made the squash and stuff... So maybe a follow-up docu update (through separate PR) would be more convenient at the moment? So perhaps the addition currently has a grade of 7 out of 10 and everyone is, of course, free to make it a 9 out of 10. 

Anyway, my main objective with this PR was to get an 'I build my own workflow'-vibe going with the intended reader, so to never limit oneself to a single tool but combine software and make it work for you. I also always prefer actual working examples to an abstraction any time of the week, so I think the aim for at least this piece of documentation should be to help people along and not try to please everyone (it is behind a 'tab', right 😅).

PR #3831, opened 2024-06-25
feat: Add --no-network flag to doctor command

### LINKED ISSUE #3829: Optional speed up to chezmoi doctor

When using `chezmoi doctor` on a machine without internet access, it takes a long time to return (30sec).

I assume this is because the `latest-version` check has to timeout in order to fail.

A command line flag to the doctor command telling `doctor` to do no internet requests, for example "`--no-network`".

* An alternative could be in the `doctor` command to get rid of the `latest-version` test altogether, but this might break the expectations of the "normal" user who has internet access.

* An alternative option name could be `--fast`, which in future could cover other doctor checks which take in some special environments unreasonably long.

## Additional context

One problem with a `--no-network` flag is what to print in the output when no internet request was made. I suggest "not checked".

twpayne:
Fixes #3829.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/doctor.md
+++ b/assets/chezmoi.io/docs/reference/commands/doctor.md
@@ -2,6 +2,10 @@
 
 Check for potential problems.
 
+## `--no-network`
+
+Do not use any network connections.
+
 !!! example
 
     ```console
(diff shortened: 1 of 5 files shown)


halloleo:
Wow, that's quick! Amazing! Thanks.

PR #3855, opened 2024-07-04
feat: Add onepasswordResolve template func for retrieving secrets with the 1Password SDK

### LINKED ISSUE #3852: Integrate 1Password SDKs to remove dependency on CLI

1Password recently released SDKs that support reading and writing secrets directly from 1Password without the need for the CLI or a Connect server. This simplifies the integration process and reduces the dependency on additional infrastructure.

Extend the 1Password integration to support the new 1Password SDKs. This would allow users to:

- Load secrets directly from 1Password without the need for hosting a Connect server.
- Retrieve, create, edit, and delete items in 1Password vaults.
- Authenticate using 1Password service accounts

Continuing to use the integration with the CLI or Secrets Automation.

## Additional context

The 1Password SDKs are currently in beta. Documentation can be found here: https://developer.1password.com/docs/sdks/

twpayne:
Fixes #3852.

@carey404 would you be able to test this? tl;dr instructions:

1. Build chezmoi using the code in this PR.
2. Ensure you have the `$OP_SERVICE_ACCOUNT_TOKEN` environment variable set to a suitable value.
3. Call the `{{ onepasswordResolve "op://vault/item/field" }}` template function in your template.
4. Check the output is OK.

Assuming that you're using chezmoi built with this PR, a one-line way to test this is:

```console
# OP_SERVICE_ACCOUNT_TOKEN="<your-service-account-token>" chezmoi execute-template '{{ onepasswordResolve "op://vault/item/field" }}'
```

which should print out the secret you want.


internal/cmd/onepasswordsdktemplatefuncs.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (c *Config) onepasswordSDKClient(ctx context.Context) (*onepassword.Client, error) {
+	if c.OnepasswordSDK.client != nil || c.OnepasswordSDK.clientErr != nil {
+		return c.OnepasswordSDK.client, c.OnepasswordSDK.clientErr
+	}
+
+	c.OnepasswordSDK.client, c.OnepasswordSDK.clientErr = onepassword.NewClient(
+		ctx,
+		onepassword.WithServiceAccountToken(c.OnepasswordSDK.Token),
+		onepassword.WithIntegrationInfo("chezmoi", c.versionInfo.Version),
```
carey404, line 49: I'm getting the following error when retrieving a secret:

`chezmoi: template: arg1:1:3: executing "arg1" at <onepasswordResolve "op://ihxdr5a6susziu5yo5rlj4ypeq/Login/password">: error calling onepasswordResolve: error initializing client: invalid user input: encountered the following errors: integration version was not specified`

To verify, I changed `c.versionInfo.Version` to `"1.0"` and confirmed that it works as expected. I'm wondering if it could be related to my local build and the version number not being set.

carey404:
Thanks for the quick turnaround on this @twpayne! 

Everything works as expected, with the exception of the version not specified error I received.

internal/cmd/onepasswordsdktemplatefuncs.go
twpayne: Thanks for the testing. You're right, the version needs to be specified when building, otherwise it defaults to the empty string.

Can you build chezmoi with:

```console
$ go build -ldflags "-X main.version=v2.0.0" .
```

and retry the test with the resulting binary?

assets/chezmoi.io/docs/reference/templates/1password-functions/onepasswordResolve.md
```diff
@@ -0,0 +1,20 @@
+# `onepasswordResolve` *url*
```
halostatue, line 1: I think that the name should be different and that we can adjust it over time, because this function is (currently) unlike any other 1Password function in that it *only* works with a service account token.

It's also worth noting that the presence of `OP_SERVICE_ACCOUNT_TOKEN` *will* change the behaviour of the 1Password CLI commands, so these two approaches *should not be mixed* unless you’re only working with a service account.

twpayne: Thanks for the review!

How about `onepasswordSDKResolve` is a name?

I'll also make the name of the environment variable containing the token configurable so it can be mixed with the existing 1Password template functions.

halostatue: I think that would work nicely. If I remember reading correctly, the 1Password SDK will automatically read `OP_SERVICE_ACCOUNT_TOKEN` unless a token is provided, but if we always read from `$<onepasswordSDK.token:-${CHEZMOI_OP_SERVICE_ACCOUNT_TOKEN:-${OP_SERVICE_ACCOUNT_TOKEN}}>` then I think that we’re good.

halostatue (approved):

twpayne:
Thank you both @carey404 and @halostatue for your reviews and testing here. I've made some significant changes to this PR and would love to get your input on them.

Changes since you last viewed are:

* Template functions that use the 1Password SDK now have the `onepasswordSDK` prefix and their own section in the reference manual.
* I made the environment variable from which chezmoi reads the 1Password session token configurable.
* I renamed `onepasswordSDKResolve` to `onepasswordSDKSecretsResolve` because it uses the [`Secrets.Resolve`](https://pkg.go.dev/github.com/1password/onepassword-sdk-go#SecretsSource.Resolve) SDK function.
* I added a `onepasswordSDKItemsGet` template function based on 1Password's [`Items.Get`](https://pkg.go.dev/github.com/1password/onepassword-sdk-go#ItemsSource.Get) API. This template function includes a transformation of the returned fields and sections from slices to maps indexed by ID, which should make them easier to use.

@carey404 I have no way to test these 1Password SDK functions without running an actual 1Password client instance and generating my own 1Password session keys, which obviously cannot be a unit test. I'll open a separate issue in the 1Password SDK repo requesting a mock client for testing.

halostatue (approved):

twpayne:
The cache in memory and is only for the duration of `chezmoi apply`. The next time you run `chezmoi apply` new values will be fetched.

twitchel:
> The cache in memory and is only for the duration of `chezmoi apply`. The next time you run `chezmoi apply` new values will be fetched.

Ah perfect, looking forward to this then!

halostatue:
> Loving this idea/implementation, one less external dependency to pull in to get secrets automated.
> 
> I had a quick question around the `onepasswordSDKItemsGet`/`onepasswordSDKSecretsResolve` method calls. I see in the documentation that the values are cached if the same key is hit multiple times. Is there a way to forcibly bust that cache? I'm thinking of a potential scenario where you may want to roll a SSH key for example, you'd want the ability to force chezmoi to pull in the latest data attached to that item.

Note that things like SSH keys and attachments/documents are *not* currently supported by the 1Password SDK. This is not a replacement for the 1Password CLI, as it also only works with service accounts (at least for now; I can't see that changing without the overall 1Password security experience being severely degraded, though).

twpayne:
@carey404 would you be able to test this again with the new template functions? I would like to include this in the next release of chezmoi.

You can test the new template functions with commands like:

```console
$ export OP_SERVICE_ACCOUNT_TOKEN="<your-service-account-token>"
$ chezmoi execute-template '{{ onepasswordSDKSecretsResolve "op://vault/item/field" }}'
$ chezmoi execute-template '{{ onepasswordSDKItemsGet "vault" "item" | toJson }}'
```

twpayne:
Merging this as-is for now so it can be tested by users.

PR #3859, opened 2024-07-06
chore: Avoid using append in loops

twpayne:


halostatue:
LGTM, but I’ve only skimmed it.

PR #3862, opened 2024-07-09
fix: keep initFuncs when using 'includeTemplate' in config

kdomanski:
When trying to exec 'chezmoi init' or 'chezmoi apply --init'  with .chezmoi.toml.tmpl including another file through 'includeTemplate', the initTemplateFuncs are only available in the first file, but not in the included template.

This is because the initial execution operates on a copy of c.templateFuncs, which wouldn't get passed on to includeTemplateTemplateFunc.

With this change, initFunctions will be added directly to c.templateFuncs and then removed after the config has been generated.

Note: Due to a separate implementation, 'chezmoi execute-template --init' is not affected by this bug.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -811,8 +811,15 @@ func (c *Config) createAndReloadConfigFile(cmd *cobra.Command) error {
 // createConfigFile creates a config file using a template and returns its
 // contents.
 func (c *Config) createConfigFile(filename chezmoi.RelPath, data []byte, cmd *cobra.Command) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


kdomanski:
Later this week I should have time to write a testcase.

kdomanski:
Done. I verified that the test fails before the change (`function "stdinIsATTY" not defined`) and succeeds after.

I briefly tried to factor out a template renderer from `struct Config` to see if I can reduce the number of codepaths which manipulate `templateFuncs`. Sadly, no luck.

twpayne:
Thank you very much!

PR #3883, opened 2024-08-01
feat: Added two github helper for retrieving specific release instead of latest release.

pix:
I've added a new feature that lets users download specific versions of GitHub releases, eg:

```
.local/bin/delta:
  type: archive-file
  url: {{ gitHubReleaseAssetURL "dandavison/delta" "0.16.5" "delta-*-x86_64-unknown-linux-musl.tar.gz" | quote }}
  executable: true
  stripComponents: 1
  path: delta
```

I included a quick test and doc.

Feel free to tell me if anything else needs adjustments.


assets/chezmoi.io/mkdocs.yml
```diff
@@ -225,7 +225,9 @@ nav:
       - reference/templates/github-functions/index.md
       - gitHubKeys: reference/templates/github-functions/gitHubKeys.md
       - gitHubLatestRelease: reference/templates/github-functions/gitHubLatestRelease.md
+      - gitHubRelease: reference/templates/github-functions/gitHubRelease.md
```
twpayne, line 228: The files `gitHubRelease.md` and `gitHubReleaseAssetURL.md` are missing from this PR.

internal/cmd/githubtemplatefuncs.go
```diff
@@ -37,21 +37,25 @@ type gitHubTagsState struct {
 }
 
 var (
-	gitHubKeysStateBucket          = []byte("gitHubLatestKeysState")
-	gitHubLatestReleaseStateBucket = []byte("gitHubLatestReleaseState")
-	gitHubReleasesStateBucket      = []byte("gitHubReleasesState")
-	gitHubTagsStateBucket          = []byte("gitHubTagsState")
+	gitHubKeysStateBucket     = []byte("gitHubLatestKeysState")
+	gitHubReleaseStateBucket  = []byte("gitHubReleaseState")
```
twpayne, line 41: Rather than combining the states for releases and latest releases in the same bucket and using an extra magic key (`<latest>`) to distinguish between them, please use a separate bucket for the new functions.

internal/cmd/testdata/scripts/githubtemplatefuncs.txtar
```diff
@@ -8,6 +8,10 @@ stdout ^ssh-rsa
 exec chezmoi execute-template '{{ (gitHubLatestRelease "twpayne/chezmoi").TagName }}'
 stdout ^v2\.
 
+# test gitHubLatestRelease template function
+exec chezmoi execute-template '{{ (gitHubLatestRelease "2.51.0" "twpayne/chezmoi").TagName }}'
+stdout ^v2\.
```
twpayne, line 13: As the tag should be exactly `v2.51.0` you can use a stricter test here.

twpayne (commented):
Thanks for this. Please add a new bucket for the new state, rather than mixing everything in a single bucket.

assets/chezmoi.io/mkdocs.yml
pix: Should be fine now.

internal/cmd/githubtemplatefuncs.go
pix: Done. It now should use a `gitHubVersionReleaseStateBucket`

twpayne:
Thank you, this looks excellent. I want to do a final review tomorrow, but it all looks good. If you want to run the full set of checks locally (e.g. to fix the whitespace errors) you can run `make smoketest`.

pix:
What do you think about allowing some "magic" value for downloading the latest release ?

Use case:
```
some-external:
  url: {{ gitHubReleaseAssetURL "user/repo" .data.versions.repo (printf "repo-*-%s-unknown-linux-musl.tar.gz" $arch) | quote }}

----

data.yaml:

data:
  versions:
      repo: latest
      chezmoi: v2.51.0
```

twpayne:
> What do you think about allowing some "magic" value for downloading the latest release ?

There's [`gitHubLatestRelease`](https://www.chezmoi.io/reference/templates/github-functions/gitHubLatestRelease/) for this, or am I missing something?

bradenhilton:
The function name in the new tests is incorrect.

twpayne:
> The function name in the new tests is incorrect.

Thanks, I'm fixing this. The tests weren't running because they only run when the `CHEZMOI_GITHUB_TOKEN` environment variable is set, as they make requests to the GitHub API which has very low rate limits if you don't authenticate.

PR #3888, opened 2024-08-05
fix: Don't modify ciphertext in edit command if plaintext did not change

twpayne:
Fixes #3387.

### DIFF
--- a/internal/cmd/editcmd.go
+++ b/internal/cmd/editcmd.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"log/slog"
 	"os"
 	"runtime"
@@ -87,6 +88,7 @@ func (c *Config) runEditCmd(cmd *cobra.Command, args []string) error {
 	type transparentlyDecryptedFile struct {
 		sourceAbsPath    chezmoi.AbsPath
 		decryptedAbsPath chezmoi.AbsPath
+		preEditPlaintext []byte
 	}
 	var transparentlyDecryptedFiles []transparentlyDecryptedFile
 TARGET_REL_PATH:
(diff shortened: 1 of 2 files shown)


internal/cmd/editcmd.go
```diff
@@ -117,8 +120,9 @@ TARGET_REL_PATH:
 				return err
 			}
 			transparentlyDecryptedFile := transparentlyDecryptedFile{
-				sourceAbsPath:    c.SourceDirAbsPath.Join(sourceRelPath.RelPath()),
-				decryptedAbsPath: decryptedAbsPath,
+				sourceAbsPath:       c.SourceDirAbsPath.Join(sourceRelPath.RelPath()),
+				decryptedAbsPath:    decryptedAbsPath,
+				preEditPlaintextSum: sha512.Sum512(contents),
```
halostatue, line 125: Raw SHA2-512 is subject to length-extension attacks. A better choice here would be [SHA2 384](https://pkg.go.dev/crypto/sha512#Sum384) or [SHA2 512/256](https://pkg.go.dev/crypto/sha512#Sum512_256).

It *might* be worth putting the hash and compare in their own functions so that you can change implementations later while only changing one place in the code.

(Sadly, Go is one of the few runtimes that appears to have a native SHA2 512/256 implementation, meaning it can't be used on some other runtimes I want to use it on; it's not just a matter of truncating the output, but the two truncated SHA2 512 implementations also have different starting IVs.)

halostatue (approved):

internal/cmd/editcmd.go
twpayne: On reflection, there's no need to use a hash here. We can compare the contents directly.

halostatue (approved):

PR #3867, opened 2024-07-13
chore: Build with goreleaser v2.1.0

twpayne:

PR #3894, opened 2024-08-07
chore: Limit commit hash length to seven characters in website

twpayne:
@bradenhilton spotted the long commit hashes in the 2.51.0 [release notes](https://www.chezmoi.io/reference/release-history/) in https://github.com/twpayne/chezmoi/pull/3867#issuecomment-2229751267. This should fix it.


twpayne:
Note that this fixes chezmoi's release notes, not the chocolately release notes.

bradenhilton:
Ahh. To be clear, I wasn't pointing out the commit length, I was just showing how it looked on the package page. This is a good change though :)

PR #3898, opened 2024-08-12
docs: Remove outdated infobox for age passphrase and symmetric encryption

Gellardo:
This should bring the infobox of the age doc-page in line with the information directly above it.


twpayne:
This is not correct. The builtin age support does not support passphrases or symmetric encryption.

PR #3941, opened 2024-09-14
Add editing files to the daily operations

alper:
I had to relatively painfully piece this together. I think it should be in the daily operations.

I still need to:
- [ ] Test the generated documentation

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/daily-operations.md
+++ b/assets/chezmoi.io/docs/user-guide/daily-operations.md
@@ -28,12 +28,12 @@ The preferred way to work is to make changes to the files in your source directo
 # Assuming you have `.gitignore_global` already managed by chezmoi
 echo "# My Addition" >> .gitignore_global
 
-# This will open your editor with the contents of .gitignore_global
+# This will open your `$EDITOR` with the contents of .gitignore_global
 chezmoi edit .gitignore_global
 
 # Complete your edits and save the file
 
-# Then if you don't have `autoPush` configured you'll have to do the [`git`](../reference/commands/git.md) part yourself
+# Then if you don't have `autoPush` configured (see below) you'll have to do the [`git`](../reference/commands/git.md) part yourself
 chezmoi git commit -- -a -m "Made a change"
 chezmoi git push
 ```
--- a/assets/chezmoi.io/docs/user-guide/daily-operations.md
+++ b/assets/chezmoi.io/docs/user-guide/daily-operations.md
@@ -25,7 +25,7 @@ sequenceDiagram
 You can make edits to the source directory, but more likely you'll be tweaking things in your `$HOME` directly and then want to persist the results:
 
 ```console
-# Assuming you have `.gitignore` already managed by chezmoi
+# Assuming you have `.gitignore_global` already managed by chezmoi
 echo "# My Addition" >> .gitignore_global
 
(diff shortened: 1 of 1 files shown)


halostatue:
I am 👎 on this change, at least as part of "daily operations". (If @twpayne or @bradenhilton are 👍, I will have some wording suggestions; it reads awkwardly to me.)

Like you, I edit most things in `$HOME` rather than in `$CHEZMOI_HOME` then apply, but my tool of choice is not `chezmoi re-add` but `chezmoi merge-all` (I’m usually changing more than one file) and I *never* use the `chezmoi git` subcommand (there are a few reasons, but mostly that I’ve just always found it easier to do `chezmoi cd` and ensure that everything is where it should be before committing and pushing).

Documentation like this is likely to become a strong recommendation, and [re-add](https://www.chezmoi.io/reference/commands/re-add/) has caveats:

> Re-add modified files in the target state, preserving any `encrypted_` attributes. chezmoi will not overwrite templates, and all entries that are not files are ignored. Directories are recursed into by default.

About ⅓ of my configs are templates, so I can't use `re-add` *anyway*.

bradenhilton:
I exclusively edit my dotfiles from the source directory (what I'll call inside out) rather than outside in, so I have little experience with `{re-}add` or `merge{-all}` and their considerations.

At a minimum, I would split the code block into 3 and promote the comments to normal text between the commands. The lines should also wrap at 80 characters and the commits should be squashed with a [conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) (this is why the `main / check` check is currently failing).

alper:
Hey! This is not a final proposal but I feel there should be at least one recommended way of how to actually use this in the onboarding documentation.

It's kinda steep to have to figure this out jumping around GitHub issues and documentation.

halostatue:
> Hey! This is not a final proposal but I feel there should be at least one recommended way of how to actually use this in the onboarding documentation.

I think that the "inside-out" approach is what should be documented for the daily use, not the "outside-in" approach. That is, what should be documented in this section is the `chezmoi edit ~/<file>` and `chezmoi git commit` (or the auto-commit mechanism).

> It's kinda steep to have to figure this out jumping around GitHub issues and documentation.

I think that having a page with discussion on the `inside-out` and `outside-in` approaches and the commands that might be used would probably be *better* than talking about the `outside-in` approach in this page, and any discussion on the `outside-in` approach would need to discuss  both `chezmoi re-add` and `chezmoi merge{,-all}`.

alper:
> I think that having a page with discussion on the inside-out and outside-in approaches and the commands that might be used would probably be better than talking about the outside-in approach in this page, and any discussion on the outside-in approach would need to discuss both chezmoi re-add and chezmoi merge{,-all}.

I'll get that going after this.

> I think that the "inside-out" approach is what should be documented for the daily use

I think this is an unintuitive two-step way to use this tool, but sure. Please check.

Will clean up the commits etc. if there's agreement on the addition.

Also I cant't build docs myself at the moment with a rather weird error message:

```
FileNotFoundError: [Errno 2] No such file or directory: 'go'
```

alper:
> That should probably be mentioned in https://www.chezmoi.io/developer-guide/website/ now that I think about it.

For that: That should probably be mentioned in https://www.chezmoi.io/developer-guide/website/ now that I think about it.

twpayne:
Thanks for this. I've consolidated everything into #3948 with (I hope) correct author credits.

PR #3945, opened 2024-09-15
fix: Improve phonetic pronunciation of chezmoi

### LINKED ISSUE #3944: Wrong phonetics
Hello, just a quick note to say that the english phonetic transcription in the main documentation page is wrong:

> With chezmoi, pronounced /ʃeɪ mwa/ (shay-moi)

- `/ʃeɪ mwa/` is correct
- `shay-moi` is not because it can be read as `shay-moy`. It would be better transcripted as for instance `shay-moo-a`

twpayne:
Fixes #3944.

### DIFF
--- a/assets/chezmoi.io/docs/index.md.tmpl
+++ b/assets/chezmoi.io/docs/index.md.tmpl
@@ -21,7 +21,7 @@ including:
 * full file encryption (using gpg or age)
 * running scripts (to handle everything else)
 
-With chezmoi, pronounced /ʃeɪ mwa/ (shay-moi), you can install chezmoi and your
+With chezmoi, pronounced /ʃeɪ mwa/ (shay-mwa), you can install chezmoi and your
 dotfiles from your GitHub dotfiles repo on a new, empty machine with a single
 command:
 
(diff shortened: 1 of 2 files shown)


bradenhilton:
I get that there isn't a perfect solution to this, but I wonder if "shay-mwa" would be better?

If I said "shay-moo-a" out loud I'd probably sound like a cow.

twpayne:
> I get that there isn't a perfect solution to this, but I wonder if "shay-mwa" would be better?

Agreed. Will wait for feedback from others. Will also ping a friend who's an expert in phonetics.

halostatue:
I agree that `shay-mwa` is better than `shay-mooa`. (Not a francophone, but I speak French well enough that I hear a `w` when I hear certain French words: `moi`, `oui`).  In Quebec French, `oui` is pronounced a bit more like "way" than "wee", so I wouldn't be surprised if `chezmoi` is more like `shay-mway` in Quebec.

halostatue:
The problem with `moo-a` as a proposed example is that in English, the `oo` sound isn't at *all* like what the French words are.

twpayne:
Agreed that `shay-mwa` is better, thanks all!

Issue #3980, opened 2024-10-10
Builtin age doesn't work


KapJI:
## Describe the bug

From docs (https://www.chezmoi.io/user-guide/encryption/age/#builtin-age-encryption):
```
chezmoi has builtin support for age encryption which is automatically used if the age command is not found in $PATH.
```

I configured `age` to be installed via `.chezmoiexternal.toml`:
```toml
[".local/bin/age"]
    type = "archive-file"
    url = "https://dl.filippo.io/age/latest?for=linux/amd64"
    path = "age/age"
    executable = true
    refreshPeriod = "168h"
```

But `czm apply` fails with:
```console
$ czm apply
chezmoi: .config/zsh/config.d/work.zsh: exec: "age": executable file not found in $PATH
```

`externals` files are processed in order of target names, that's mentioned in docs. I guess I can put age to `.bin` to make it work, although I'd like keep binaries where they are currently. And ideally remove dependency on age completely from my setup.

`chezmoi age` doesn't work here either but I guess this is expected, although a bit confusing:
```console
$ chezmoi age decrypt dot_config/zsh/config.d/encrypted_work.zsh.tmpl.age
chezmoi: only passphrase encryption is supported
```

**So is it no longer supported and docs are outdated or something is broken here?**

## To reproduce

With `age` uninstalled:
```console
$ chezmoi decrypt --verbose dot_config/zsh/config.d/encrypted_work.zsh.tmpl.age
chezmoi: exec: "age": executable file not found in $PATH
```

Adding `--use-builtin-age on` doesn't show any errors:
```console
$ chezmoi --use-builtin-age on --verbose decrypt dot_config/zsh/config.d/encrypted_work.zsh.tmpl.age
chezmoi: exec: "age": executable file not found in $PATH
```

age config in `.config/chezmoi/chezmoi.toml`:
```toml
encryption = "age"

[age]
    identity = "~/.config/chezmoi/key.txt"
    recipient = "age1p6n7pr73yfxysh8qjty4guhfea9gqgh7uwk8w6e3wx7r9eqqw4sqeg4r9g"
```

Tested on Linux (x64), Mac (arm64) and Windows (x64).

## Expected behavior

Builtin age is used.

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT    CHECK                       MESSAGE
[...]

PR #3982, opened 2024-10-11
docs: Add notes where to place 'encryption' setting

KapJI:
Follow up to #3980 

* Advice to append content to the config in `encryption.md` very likely will cause `encryption` set incorrectly if config is not empty. Replace it with instructions what to do because command to prepend file in bash is a bit cumbersome and may also lead to incorrect config if other global settings are specified already.
* Add notes to articles about each encryption method to be careful with placement of `encryption` setting.


twpayne:
Thanks!

PR #3983, opened 2024-10-11
feat: Add darwinVersion to template data

KapJI:
There is `windowsVersion` for Windows and `osRelease` for Linux but nothing for MacOS. Add similar `darwinVersion` section for MacOS.

## Implementation

Here I added a new dependency to read .plist file. It's possible to do it without new dependencies by running `sw_vers` tool and parsing its output. I have such version here: https://github.com/KapJI/chezmoi/commit/38b949aebd38d0fef33e029a2cafb3b1937ab4de . Reading plist should be more lightweight and closer to Windows version which reads registry. Let me know which one you prefer.

## Use case

I want to have run_onchange script which runs after software update which resets some settings. Currently I can think of some hacky solutions using script which writes `sw_vers` output to file on each `chezmoi apply` and then reading it using `include` template function. This will make it more straightfordard and simpler.

If you agree this is useful, I can add some tests.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/variables.md
+++ b/assets/chezmoi.io/docs/reference/templates/variables.md
@@ -9,6 +9,7 @@ chezmoi provides the following automatically-populated variables:
 | `.chezmoi.cacheDir`           | string   | The cache directory                                                                                                                                   |
 | `.chezmoi.config`             | object   | The configuration, as read from the config file                                                                                                       |
 | `.chezmoi.configFile`         | string   | The path to the configuration file used by chezmoi                                                                                                    |
+| `.chezmoi.darwinVersion`      | object   | macOS version information, if running on macOS                                                                                                        |
 | `.chezmoi.executable`         | string   | The path to the `chezmoi` executable, if available                                                                                                    |
 | `.chezmoi.fqdnHostname`       | string   | The fully-qualified domain name hostname of the machine chezmoi is running on                                                                         |
 | `.chezmoi.gid`                | string   | The primary group ID                                                                                                                                  |
@@ -31,6 +32,19 @@ chezmoi provides the following automatically-populated variables:
 | `.chezmoi.version.version`    | string   | The version of chezmoi                                                                                                                                |
 | `.chezmoi.windowsVersion`     | object   | Windows version information, if running on Windows                                                                                                    |
 | `.chezmoi.workingTree`        | string   | The working tree of the source directory                                                                                                              |
+
+`.chezmoi.darwinVersion` contains the following keys populated from the
+`sw_vers` tool.
+
+| Key                         | Type    |
+| --------------------------- | ------- |
+| `buildVersion`              | string  |
+| `productName`               | string  |
+| `productVersion`            | string  |
+| `productVersionExtra`       | string  |
+| `productMajorVersion`       | integer |
+| `productMinorVersion`       | integer |
+| `productPatchVersion`       | integer |
 
 `.chezmoi.windowsVersion` contains the following keys populated from the
 registry key `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows
--- a/internal/cmd/util_unix.go
+++ b/internal/cmd/util_unix.go
@@ -4,6 +4,10 @@ package cmd
 
 import (
 	"io/fs"
+	"os/exec"
+	"runtime"
+	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/twpayne/chezmoi/v2/internal/chezmoi"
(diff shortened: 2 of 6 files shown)


halostatue:
I like this idea, but do not like the implementation of reading a `plist` file, even if it is relatively stable. The output of `sw_vers` is also parseable or can be called multiple times to get each value separately:

```console
$ sw_vers
ProductName:            macOS
ProductVersion:         14.7
BuildVersion:           23H124
$ sw_vers --productName
macOS
$ sw_vers --productVersion
14.7
$ sw_vers --productVersionExtra

$ sw_vers --buildVersion
23H124
```

assets/chezmoi.io/docs/reference/templates/variables.md
```diff
@@ -32,6 +33,15 @@ chezmoi provides the following automatically-populated variables:
...
 | `.chezmoi.workingTree`        | string   | The working tree of the source directory                                                                                                              |
 
+`.chezmoi.darwinVersion` contains the following keys populated from the
+`/System/Library/CoreServices/SystemVersion.plist`.
+
+| Key                         | Type    |
+| --------------------------- | ------- |
+| `buildVersion`              | string  |
+| `productName`               | string  |
+| `productVersion`            | integer |
```
halostatue, line 43: This should not be an `integer` but a `string`. If you want to parse it into `productMajorVersion` and `productMinorVersion` integer values, that is acceptable.

KapJI: This is a typo from copy pasting. Should be a string.

halostatue: It might be worth parsing it into `productMajorVersion` and `productMinorVersion` as well as noted. That will make version comparison easier.

KapJI:
Ok, I agree that `sw_vers` is safer. Here we can just dump all fields without hardcoding them, similar to `osRelease`.

KapJI:
Added string `productVersionExtra`, and integers `productMajorVersion`, `productMinorVersion`, `productPatchVersion`.

internal/cmd/util_unix.go
```diff
@@ -17,6 +21,64 @@ func fileInfoUID(info fs.FileInfo) int {
...
+			continue
+		}
+		parts := strings.Split(line, ":")
+		if len(parts) != 2 {
+			continue
+		}
+		key := strings.TrimSpace(parts[0])
+		key = strings.ToLower(key[:1]) + key[1:]
+		value := strings.TrimSpace(parts[1])
+		darwinVersion[key] = value
```
halostatue, line 46: This worries me a little. If there is any localization of the output (it seems unlikely because it *looks* like it's programmatic output), this and the next step will not produce the documented result.

It looks like it will work, but I’m not sure which of the three ways we have are best:

1. automatic translation as implemented
2. string matching (a case statement on the trimmed `key`)
3. multiple explicit calls as is being done for `--productVersionExtra`

KapJI: Let's do multiple calls. Do you want to have `--productName` there? It's unlikely chezmoi will run on something other than macOS.

halostatue: Let's keep it since it's part of the default output and if they change the output for any reason it could be interesting to know. (I would have hoped that `productName` would have been something like `macOS Sierra`, but no.)

KapJI:
Changed to separate `sw_vers` calls, added tests.

internal/cmd/util_unix.go
```diff
@@ -17,6 +21,54 @@ func fileInfoUID(info fs.FileInfo) int {
...
+		"productVersionExtra",
+	} {
+		output, err := exec.Command("sw_vers", "--"+name).Output()
+		if err != nil {
+			return nil, err
+		}
+		darwinVersion[name] = strings.TrimSpace(string(output))
+	}
+
+	productVersion, ok := darwinVersion["productVersion"].(string)
```
halostatue, line 42: *Minor* suggestion here, but you can take it or leave it because I think that both approaches are readable. Instead of having a lookup on `productVersion` separate, add an `if name == "productVersion"` after line 39 where you do all of the conversion from line 44 to 68.

(If I were doing this, I would make the change and then see if it improves or worsens the readability; it could go either way by localizing the behaviour or by adding even more indenting. I think it's a coin flip as to which way it will go.)

halostatue (approved):
This looks really good. I’ve made one more suggestion that may or may not result in better code.

internal/cmd/util_unix.go
KapJI: I prefer current version a bit more because this way two parts are clearly separated. In proposed version these parts are more entangled with each other and parsing logic which is unrelated for everything other than `productVersion` is one level deeper which worsens readbility a bit.

twpayne:
Thanks for opening the PR.

All of this functionality can be achieved with the existing `output` template function, for example:

```
{{ output "sw_vers" "--productVersion" }}
```

Adding this `darwinVersion` template data adds four process invocations to every chezmoi start.

So, I do not think this is needed.

PR #3984, opened 2024-10-11
docs: Add example for running script on macOS update

KapJI:
Somehow I overlooked existense of `output` for executing arbitrary commands. I guess its name doesn't give a hint that it runs something.

Add an example in user guide section from #3983 to show one more case how it can be used.


twpayne:
Thanks!

PR #3997, opened 2024-10-17
docs: Improve description for command arguments

KapJI:
* Add missing arguments. They are synced with `chezmoi cmd --help`.
* Add links to list of supported types for `--include` and `--exclude`.
* Clarify how to disable flags which are enabled by default.
* Add supported values for some flags.
* Add list of supported upgrade methods.
* Move some common and global flags to the correct place.
* Better output for `chezmoi remove --help`.

### DIFF
--- a/assets/chezmoi.io/docs/reference/command-line-flags/common.md
+++ b/assets/chezmoi.io/docs/reference/command-line-flags/common.md
@@ -6,10 +6,21 @@ The following flags apply to multiple commands where they are relevant.
 
 Set the output format.
 
+## `-h`, `--help`
+
+Print help.
+
 ## `-i`, `--include` *types*
 
-Include target state entries of type *types*. *types* is a comma-separated list
-of types:
+Include target state entries of type *types*.
+
+!!! example
+
+    `--include=files` specifies all files.
+
+### Available types
+
+*types* is a comma-separated list of types:
 
 | Type        | Description                 |
 | ----------- | --------------------------- |
@@ -29,18 +40,10 @@ Types can be preceded with `no` to remove them.
 
 Types can be explicitly excluded with the `--exclude` flag.
 
-!!! example
-
-    `--include=files` specifies all files.
-
 ## `--init`
 
 Regenerate and reread the config file from the config file template before
 computing the target state.
-
-## `--interactive`
-
-Prompt before applying each target.
 
 ## `-o`, `--output` *filename*
 
@@ -54,11 +57,6 @@ Also perform command on all parent directories of *target*.
 
 Recurse into subdirectories, `true` by default.
 
-## `--source-path`
-
-Interpret *targets* passed to the command as paths in the source directory
-rather than the destination directory.
-
 ## `--tree`
 
 Print paths as a tree instead of a list.
@@ -70,8 +68,7 @@ is set.
 
 ## `-x`, `--exclude` *types*
 
-Exclude target state entries of type *types*. *types* is defined as in the
-`--include` flag and defaults to `none`.
+Exclude target state entries of type [*types*](#available-types). Defaults to `none`.
 
 !!! example
 
--- a/assets/chezmoi.io/docs/reference/command-line-flags/global.md
+++ b/assets/chezmoi.io/docs/reference/command-line-flags/global.md
@@ -40,13 +40,17 @@ changes that would be made without making them.
 
 Make changes without prompting.
 
-## `-h`, `--help`
+## `--interactive`
 
-Print help.
+Prompt before applying each target.
 
 ## `-k`, `--keep-going`
 
 Keep going as far as possible after a encountering an error.
+
+## `--mode` `file`|`symlink`
+
+Mode of operation. The default is `file`.
 
 ## `--no-pager`
 
@@ -89,6 +93,11 @@ if no cached external is available.
 > Configuration: `sourceDir`
 
 Use *directory* as the source directory.
+
+## `--source-path`
+
+Interpret *targets* passed to the command as paths in the source directory
+rather than the destination directory.
 
 ## `--use-builtin-age` *value*
 
--- a/assets/chezmoi.io/docs/reference/commands/add.md
+++ b/assets/chezmoi.io/docs/reference/commands/add.md
@@ -17,11 +17,23 @@ This implies the `--template` option.
     templates with unwanted variable substitutions. Carefully review any
     templates it generates.
 
+## `--create`
+
+Add files that should exist, irrespective of their contents.
+
 ## `--encrypt`
 
 > Configuration: `add.encrypt`
 
 Encrypt files using the defined encryption method.
+
+## `--exact`
+
+Set the `exact` attribute on added directories.
+
+## `-x`, `--exclude` *types*
+
+Exclude entries of type [*types*](../command-line-flags/common.md#available-types),  defaults to `none`.
 
 ## `-f`, `--force`
 
@@ -33,13 +45,9 @@ overwritten.
 If the last part of a target is a symlink, add the target of the symlink
 instead of the symlink itself.
 
-## `--exact`
-
-Set the `exact` attribute on added directories.
-
 ## `-i`, `--include` *types*
 
-Only add entries of type *types*.
+Only add entries of type [*types*](../command-line-flags/common.md#available-types), defaults to `all`.
 
 ## `-p`, `--prompt`
 
@@ -51,7 +59,7 @@ Suppress warnings about adding ignored entries.
 
 ## `-r`, `--recursive`
 
-Recursively add all files, directories, and symlinks.
+Recursively add all files, directories, and symlinks, `true` by default. Can be disabled with `--recursive=false`.
 
 ## `-s`, `--secrets` `ignore`|`warning`|`error`
 
--- a/assets/chezmoi.io/docs/reference/commands/age.md
+++ b/assets/chezmoi.io/docs/reference/commands/age.md
@@ -2,13 +2,17 @@
 
 Interact with age's passphrase-based encryption.
 
-!!! hint
+# `age encrypt` [*file*...]
 
-    To get a full list of subcommands run:
+## `-p`, `--passphrase`
 
-    ```console
-    $ chezmoi age help
-    ```
+Encrypt with a passphrase.
+
+# `age decrypt` [*file*...]
+
+## `-p`, `--passphrase`
+
+Decrypt with a passphrase.
 
 !!! example
 
--- a/assets/chezmoi.io/docs/reference/commands/secret.md
+++ b/assets/chezmoi.io/docs/reference/commands/secret.md
@@ -9,13 +9,39 @@ manager. Normally you would use chezmoi's existing template functions to retriev
     If you need to pass flags to the secret manager's CLI you must separate
     them with `--` to prevent chezmoi from interpreting them.
 
-!!! hint
+# `secret keyring delete`
 
-    To get a full list of subcommands run:
+## `--service` *string*
 
-    ```console
-    $ chezmoi secret help
-    ```
+Name of the service.
+
+## `--user` *string*
+
+Name of the user.
+
+# `secret keyring get`
+
+## `--service` *string*
+
+Name of the service.
+
+## `--user` *string*
+
+Name of the user.
+
+# `secret keyring set`
+
+## `--service` *string*
+
+Name of the service.
+
+## `--user` *string*
+
+Name of the user.
+
+## `--value` *string*
+
+New value.
 
 !!! example
 
--- a/assets/chezmoi.io/docs/reference/commands/state.md
+++ b/assets/chezmoi.io/docs/reference/commands/state.md
@@ -10,6 +10,40 @@ Manipulate the persistent state.
     $ chezmoi state help
     ```
 
+# Available subcommands
+
+## `data`
+
+Print the raw data in the persistent state.
+
+## `delete`
+
+Delete a value from the persistent state.
+
+## `delete-bucket`
+
+Delete a bucket from the persistent state.
+
+## `dump`
+
+Generate a dump of the persistent state.
+
+## `get`
+
+Get a value from the persistent state.
+
+## `get-bucket`
+
+Get a bucket from the persistent state.
+
+## `reset`
+
+Reset the persistent state.
+
+## `set`
+
+Set a value from the persistent state
+
 !!! example
 
     ```console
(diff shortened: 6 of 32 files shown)


assets/chezmoi.io/docs/reference/command-line-flags/common.md
```diff
@@ -6,10 +6,21 @@ The following flags apply to multiple commands where they are relevant.
 
 Set the output format.
 
+## `-h`, `--help`
```
halostatue, line 9: `--help` should not be moved from `global.md` as it *always* exists for all commands. Common flags apply to multiple different commands, but not to *all* commands.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
```diff
@@ -6,10 +6,21 @@ The following flags apply to multiple commands where they are relevant.
...
+
+Print help.
+
 ## `-i`, `--include` *types*
 
-Include target state entries of type *types*. *types* is a comma-separated list
-of types:
+Include target state entries of type *types*.
```
halostatue, line 15: I see that you have split this so that `### Available types` presents a link target `#available-types` for use with `--exclude`, but I don't think that this improves the readability of this section.

*If* we were to consider this particular change, I think that it should instead be written as:

```markdown
Include target state entries of [*types*](#available-include-exclude-types), provided as a comma separated list.

!!! example

    `--include=files,scripts` specifies all files and scripts
```

The table would be moved to the *bottom* of this page and relabeled as `## Available Include/Exclude Types` and the `--exclude` documentation would be updated as well.

I’d need to see it to know whether it's a good change or not.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
```diff
@@ -54,11 +57,6 @@ Also perform command on all parent directories of *target*.
 
 Recurse into subdirectories, `true` by default.
 
-## `--source-path`
```
halostatue, line 57: This should not be moved from `common.md` as it does not apply to *all* commands, only a large subset.

assets/chezmoi.io/docs/reference/command-line-flags/global.md
```diff
@@ -40,14 +40,18 @@ changes that would be made without making them.
...
+## `--interactive`
 
-Print help.
+Prompt before applying each target.
 
 ## `-k`, `--keep-going`
 
 Keep going as far as possible after a encountering an error.
 
+## `--mode` `file`|`symlink`
```
halostatue, line 51: This should probably be in `common.md`, not in `global.md`.

twpayne:
Thank you very much indeed!

assets/chezmoi.io/docs/reference/commands/unmanaged.md
```diff
@@ -5,11 +5,15 @@ unmanaged files in the destination directory.
 
 It is an error to supply *path*s that are not found on the filesystem.
 
-## `-p`, `--path-style` `absolute`|`relative`
+## `-p`, `--path-style` `absolute`|`relative`|`source-absolute`|`source-relative`
```
halostatue, line 8: This should *probably* be moved to `common.md` since it's used by `ignored`, `managed`, and `unmanaged`.

assets/chezmoi.io/docs/reference/commands/upgrade.md
```diff
@@ -16,3 +16,19 @@ requests should be sufficient for most cases.
 
     If you installed chezmoi using a package manager, the `upgrade` command
     might have been removed by the package maintainer.
+
+## `--executable` *filename*
+
+Set name of executable to replace.
+
+## `--method` *method*
```
halostatue, line 24: I do not believe that `--method` should be documented as incorrect selection will cause upgrade failures.

halostatue (commented):
I don't think that this should have been merged as a number of the changes make the documentation worse.

halostatue:
@KapJI @twpayne  this was merged before I could complete a review, but I have fairly deep concerns about some of the changes you have made as noted in the comments here.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
KapJI: It can be added to both, I guess. Here is the output for `chezmoi apply --help` for example: https://gist.github.com/KapJI/049b54326d0aec676b3ea49c573fa92a

You can see that `--help` is subcommand flag, not global.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
KapJI: It can be moved to the bottom or even to the new page. I don't think it matters that much. But having this list linked from other places is very useful.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
```diff
@@ -29,19 +40,11 @@ Types can be preceded with `no` to remove them.
...
-!!! example
-
-    `--include=files` specifies all files.
-
 ## `--init`
 
 Regenerate and reread the config file from the config file template before
 computing the target state.
 
-## `--interactive`
```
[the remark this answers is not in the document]
KapJI: Again I refer to https://gist.github.com/KapJI/049b54326d0aec676b3ea49c573fa92a

It's listed within global flags.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
KapJI: Same.

assets/chezmoi.io/docs/reference/command-line-flags/global.md
KapJI: Same.

assets/chezmoi.io/docs/reference/commands/add.md
```diff
@@ -17,12 +17,24 @@ This implies the `--template` option.
...
 
 > Configuration: `add.encrypt`
 
 Encrypt files using the defined encryption method.
 
+## `--exact`
+
+Set the `exact` attribute on added directories.
+
+## `-x`, `--exclude` *types*
```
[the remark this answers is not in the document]
KapJI: Arguments are sorted by long flag in `--help` output, that's why I sorted them this way.

Your approach with linking to common for some flags works fine, it avoid duplicated descriptions, although for some flags these descriptions are personalized for subcommand to better describe what they do.

assets/chezmoi.io/docs/reference/commands/age.md
```diff
@@ -2,13 +2,17 @@
 
 Interact with age's passphrase-based encryption.
 
-!!! hint
+# `age encrypt` [*file*...]
```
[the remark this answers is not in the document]
KapJI: Moving to h2 makes sense. Deduping `-p` makes readability slightly worse to my taste, but if you prefer that, why not.

assets/chezmoi.io/docs/reference/commands/secret.md
```diff
@@ -9,13 +9,39 @@ manager. Normally you would use chezmoi's existing template functions to retriev
     If you need to pass flags to the secret manager's CLI you must separate
     them with `--` to prevent chezmoi from interpreting them.
 
-!!! hint
+# `secret keyring delete`
```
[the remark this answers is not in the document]
KapJI: Same as for `age`.

assets/chezmoi.io/docs/reference/commands/state.md
```diff
@@ -10,6 +10,40 @@ Manipulate the persistent state.
     $ chezmoi state help
     ```
 
+# Available subcommands
```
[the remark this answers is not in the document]
KapJI: Same.

assets/chezmoi.io/docs/reference/commands/unmanaged.md
KapJI: Yeah, why not. In general I think common flags do not have strict definition and it's not clear which flags should be considered common.

halostatue: `common.md` defines it well enough:

> The following flags apply to multiple commands where they are relevant.

These would be manually defined, visible in `chezmoi <command> --help` as belonging to the command, but having the same options across multiple commands. Things like `--format` would be common *except* for the case where you’re determining the archive format.

assets/chezmoi.io/docs/reference/commands/age.md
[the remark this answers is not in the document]
halostatue: I am less concerned about deduping `-p` than I am about having multiple `h1` on a page.

assets/chezmoi.io/docs/reference/commands/upgrade.md
KapJI: I guess then it should be marked as hidden and that should be reflected in documentation as well.

I still think it's better to have this documented to make it explicit that missing documentation is not an overlook. E.g put a warning note that this flag is dangerous and should be used at user's own risk.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
halostatue: The list of global and command flags from `--help` is programmatically determined by Cobra but that output should not necessarily be the determining factor on where something belongs which is a human judgement. Of the items in `chezmoi --help` listed as flags (which would be truly global), there are a number of things which are global because they make reasoning about *configuration* within chezmoi easier (such as `--interactive`) but are meaningless or useless for a large number of commands.

Picking on `--interactive`, this would include `age`, `archive`, `cat`, `cat-config`, `cd`, `completion`, `data`, `decrypt`, `destroy` (interactive by default), `diff`, `doctor`, `dump`, `dump-config`, `edit`, `edit-config`, `edit-config-template`, `encrypt`, `execute-template`, `forget` (interactive by default), `generate`, `git`, `help`, `ignored`, `license`, `managed`, `merge`, `merge-all`, `secret`, `source-path`, `state`, `status`, `target-path`, `unmanaged`, `upgrade`, and `verify`. That's all but five or six of the commands.

Documentation is about recording what matters, not merely what is.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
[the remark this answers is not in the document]
halostatue: Addressed on `--help`. It's global because it's easier in the code, but it doesn't apply to *most* commands. It won't cause an *error* because it's global, but that doesn't mean it's meaningful outside of the commands that care about `--interactive`.

assets/chezmoi.io/docs/reference/commands/add.md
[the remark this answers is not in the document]
halostatue: The documentation now has multiple sort orders for flags, then, which should be resolved.

assets/chezmoi.io/docs/reference/commands/upgrade.md
halostatue: From the perspective of having supported software for decades: don't document what you don't want people to use. Hidden support flags are *perfectly fine*, and things which should be avoided should be hidden.

assets/chezmoi.io/docs/reference/command-line-flags/common.md
KapJI: I see your point. In this case such flags should be moved to `common.md` but also added to subcommands where they matter. But maintaining such list looks pretty difficult and requires reading implementation of each particular command, so I don't think this is feasible in the long run.

This also creates different definitions for global flags. One according to implementation, another one according to documentation. This is confusing.

The ideal solution would be moving such arguments from global flags to subcommand flags. This will be benefitial for `--help` output and for documentation. But this is breaking change because some commands will stop accepting arguments they used to accept before, which is reasonable but also not great.

In the meantime, I think these global flags according to documentation should remain in `global.md`. Probably it's worth adding a note that not every flag has effect for every subcommand.

KapJI:
I don't think anything is fundamentally wrong with this PR, overall it makes documentation better.

I agree with the following points which should be addressed in subsequent PRs:

1. Replace `h1` headings with `h2` to keep only one `h1` on a page.
2. Move list of available types for `--include/--exclude` to different place from where it can be linked. E.g. to the bottom of `common.md` or a separate page.
3. Introduce section "common flags" for each command which uses flags from `common.md`. List these flags without description there and add links to descriptions from `common.md`. If necessary, there can be short command specific descriptions. This deduplicates repeating descriptions and makes them easier to maintain.
4. Re-evaluate what flags which are used by multiple command should be moved to `common.md`. IMO, if flag is used for 3 or more commands, it should go to `common.md`.
5. Hide `--method` for `chezmoi upgrade` as flag which shouldn't be used. Or add a warning in documentation that this flag shouldn't be used by most people and it may lead to failing upgrade. Maybe also add automatic safeguards if method is used on incorrect platform. E.g. if Linux only method is used on other OS.

halostatue:
> I don't think anything is fundamentally wrong with this PR, overall it makes documentation better.
> 
> I agree with the following points which should be addressed in subsequent PRs:
> 
> 1. Replace `h1` headings with `h2` to keep only one `h1` on a page.

This is the main change that I believe should have been caught and resolved prior to merging. I’ve been doing a *lot* of accessibility work over the last couple of years, and while you *can* put more than one `h1` per page, doing so really messes up accessibility landmarks. Strictly speaking, these changes break the HTML expectations and may not do what you expect with the sidebar table of contents.

> 2. Move list of available types for `--include/--exclude` to different place from where it can be linked. E.g. to the bottom of `common.md` or a separate page.
> 3. Introduce section "common flags" for each command which uses flags from `common.md`. List these flags without description there and add links to descriptions from `common.md`. If necessary, there can be short command specific descriptions. This deduplicates repeating descriptions and makes them easier to maintain.
> 4. Re-evaluate what flags which are used by multiple command should be moved to `common.md`. IMO, if flag is used for 3 or more commands, it should go to `common.md`.
> 5. Hide `--method` for `chezmoi upgrade` as flag which shouldn't be used. Or add a warning in documentation that this flag shouldn't be used by most people and it may lead to failing upgrade. Maybe also add automatic safeguards if method is used on incorrect platform. E.g. if Linux only method is used on other OS.

I agree that the rest of these can be separate PRs (but `--method` should be marked hidden ASAP; it isn't suitable for most users). With respect to the `common.md` and `global.md` criteria, I mostly agree. I believe a number of flags should be moved from `global.md` to `common.md` because they are global *only* because of the way that cobra and chezmoi configuration works.

assets/chezmoi.io/docs/reference/commands/upgrade.md
halostatue: I would agree with your assessment here except that there *were* changes to the Go code made as part of this PR (making it not just documentation).

PR #3999, opened 2024-10-17
docs: Split docs into sections for age, secret and state

KapJI:
Addressing feedback from #3997 

Downgrade `h1` headings to keep only one `h1` on each page.

This is how it looks like now. Structured and table of contents looks good.

## age
Screenshot 2024-10-17 at 21 50 01

## secret
Screenshot 2024-10-17 at 21 50 27

If we agree on that, I can update docs for other commands to follow similar style. Add sections like "Flags", "Common flags", "Example", "Notes".


halostatue (approved):
This looks really good. Thanks for taking the time to do this.

PR #4000, opened 2024-10-17
chore: Update dependencies

twpayne:


halostatue:
@twpayne There's a difference in the output on a gitleaks detection and gitleaks 8.21.1 was just released. Worth updating to that to see if it reverts the output difference that's causing these to fail? I haven't dug into this to see if it’s something which gitleaks is providing the output or chezmoi is getting different details for reporting the issues.

PR #4003, opened 2024-10-18
docs: Split docs for commands into sections

KapJI:
Address more feedback from #3997 

* Split documents into sections: "Flags", "Common flags", "Notes", "Example". This creates nice structure in table of contents.
* Promote `--path-style` to common flags.
* Consistent sorting by long flag name.
* Sort commands alphabetically in nav pane.
* Move `--use-builtin-diff` and `--output` to global flags, as they are currently implemented.
* Add note for `upgrade --method` until it's hidden from `--help`.

## Example `chezmoi add` page
Screenshot 2024-10-18 at 14 52 19

## Further improvements

`pymdownx.snippets` extension can be used to deduplicate descriptions for common flags, I can explore that in the next PR.


assets/chezmoi.io/docs/reference/commands/add.md
```diff
@@ -88,6 +94,8 @@ directory, create a symlink template with `.chezmoi.sourceDir` or
     See [this GitHub issue](https://github.com/twpayne/chezmoi/issues/1574) for
     details.
 
+## Example
```
halostatue, line 97: I don't know whether to aim for consistency here or accuracy. Strictly speaking, this should be `## Examples` because there are five example commands. It's using *one* `!!! example` block, so it *could* be `## Example` as it is now, but that doesn't feel like correct grammatical agreement to me. In the commands where we only have one example, would `## Example` be correct, or should we *always* use plural forms  for these section headers (e.g., `## Flags`, `## Common flags`, `## Notes`, `## Examples`).

I’m leaning toward consistency on section header plurals, even if it means that there's no longer section body grammatical number agreement.

assets/chezmoi.io/docs/reference/commands/secret.md
```diff
@@ -45,6 +45,15 @@ Name of the user.
 
 New value.
 
+## Notes
```
halostatue, line 48: Another consistency question here. I appreciate that this has been moved above `## Example` because of consistency, but should `## Notes` be *before* or *after* `## Examples`? I don't think that there's a *correct* choice here (some of the notes should be understood before you run a particular command; others don't matter when you read them — most people aren't on FreeBSD).

assets/chezmoi.io/docs/reference/commands/upgrade.md
```diff
@@ -17,14 +17,20 @@ requests should be sufficient for most cases.
...
 Set name of executable to replace.
 
-## `--method` *method*
+### `--method` *method*
 
 Set upgrade method.
 
+!!! danger
+
+    This flag is not safe for most people.
```
halostatue, line 32: I would prefer:

```
This flag should only be used when recommended by chezmoi developers.
```

I’d also add to the `Set upgrade method` to be something like:

```
Override upgrade method as determined by chezmoi itself.
```

halostatue (commented):
This looks good, but I have some questions that we should make decisions on overall.

assets/chezmoi.io/docs/reference/commands/add.md
KapJI: Let's change to plural.

assets/chezmoi.io/docs/reference/commands/secret.md
KapJI: This one I moved for consistency, yes. I don't have strong preference on which one should come first. There are not many notes here, but most commands have examples. Let's move notes to the end?

KapJI:
Applied all suggestions.

halostatue (approved):
I'm good with these, but it's a big enough change that I think Tom should get a glance at it first.

KapJI:
@twpayne Could you please take a look when you have a moment?

twpayne (commented):
Thanks for this, this is a very nice improvement. There are a few problems so this won't be merged immediately, but I will fix them in a follow-up PR.

For info, the problems I've spotted so far are:

* The new structure breaks the [automatic extraction of examples from the documentation](https://github.com/twpayne/chezmoi/blob/fe2dba7aee848dcd10a28482c6c2166a6a3d9641/internal/cmd/cmd.go#L143).
* The new structure uses to many elements, for example the examples for the `add` command are in an admonition called "Example" in the Examples section.
* A few of the headings have the wrong depth. This is particularly visible in the table of contents of each generated page.
* A few of the flags are missing optional arguments.

I'll fix these issues in a follow-up PR but don't want to merge this PR until the follow-up PR is ready.

KapJI:
> the examples for the add command are in an admonition called "Example" in the Examples section.

Yeah, I thought about this. I decided it's better to have them in table of contents for easier navigation. I explored a bit if it's possible to create a section without a heading but haven't found easy solution. Perhaps they can be just changed to regular code blocks.

> A few of the headings have the wrong depth.

I'll fix those.

KapJI:
> A few of the headings have the wrong depth.

I found only one instance, let me know if you see more.

twpayne:
> Yeah, I thought about this. I decided it's better to have them in table of contents for easier navigation. I explored a bit if it's possible to create a section without a heading but haven't found easy solution. Perhaps they can be just changed to regular code blocks.

Regular code blocks would be fine.

> > A few of the headings have the wrong depth.
> 
> I'll fix those.

Thanks!

Note that there's an existing bug (unrelated to this PR) in chezmoi's [parsing of examples from the reference documentation](https://github.com/twpayne/chezmoi/blob/fe2dba7aee848dcd10a28482c6c2166a6a3d9641/internal/cmd/cmd.go#L143). Currently it assumes that the Examples section is the last section of the doc, and this assumption is incorrect when there's a Notes section afterwards. You can see this bug if you run `chezmoi add --help` with the latest release of chezmoi. This is another thing I want to fix, but if you're modifying the example extraction code, then this is a logical fix to make at the same time.

KapJI:
Converted all examples to regular code blocks. I can update parsing logic tomorrow.

twpayne:
Awesome, thank you, no rush. We're all volunteers here.

KapJI:
Updated state machine to extract examples and check if every page is titled correctly.

halostatue (approved):

twpayne (approved):

twpayne:
This is fantastic, thank you.

PR #4009, opened 2024-10-20
feat: Extend quote and quoteList template funcs to handle more types

### LINKED ISSUE #4007: Template function `quote` doesn't work for pointers
## Describe the bug

`quote`, `squote` and `quoteList` return quoted pointer value instead of quoted string this pointer points to.

## To reproduce

```
{{- $release := (gitHubLatestRelease "sharkdp/bat").Assets | first -}}
Name: {{ $release.Name }}
Name quote: {{ $release.Name | quote }}
Name quoteList: {{ list $release.Name | quoteList }}
Name type: {{ $release.Name | kindOf }}
```
Prints:
```
Name: bat-musl_0.24.0_amd64.deb
Name quoted: '0xc000a841b0'
Name quoteList: ["0xc00078ad80"]
Name type: ptr
```

## Expected behavior

`{{ $ptr | quote }}` returns quoted `{{ $ptr }}`. In the meantime `"{{ $ptr}}"` can be used as a replacement.

## Additional context

I know that `quote` is coming from `sprig`, `quoteList` [is first party](https://github.com/twpayne/chezmoi/blob/35df0684e172ae0d7f8504bb004584e40df38b52/internal/cmd/templatefuncs.go#L406) though. `sprig` is not very well supported and this behaviour can be very confusing for chezmoi users. I guess that's one more reason to replace `sprig` with something better like proposed in https://github.com/twpayne/chezmoi/issues/2668

twpayne:
Fixes #4007.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -404,6 +404,7 @@ func newConfig(options ...configOption) (*Config, error) {
 	// Override sprig template functions. Delete them from the template function
 	// map first to avoid a duplicate function panic.
 	delete(c.templateFuncs, "fromJson")
+	delete(c.templateFuncs, "quote")
 	delete(c.templateFuncs, "splitList")
 	delete(c.templateFuncs, "toPrettyJson")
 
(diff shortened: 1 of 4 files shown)


halostatue (approved):
I believe that there will be other functions that exhibit the same behaviour, but this is the quickest fix for the reported issue.

PR #4013, opened 2024-10-21
feat: Extend squote template func to handle more types

KapJI:
Similar to #4009 but for `squote` this time.

As you correctly mentioned on #4007 current implementation of `squote` in `sprig` is far from being correct. Here is my attempt to fix it.

Unfortunately, Go maintainers are not willing to provide an easier way for this in `strconv`: https://github.com/golang/go/issues/69958

Use existing `strconv.Quote` implementation for heavy lifting and escaping chars like `\n` or `\x...`. The only place where `quote` character is used apart from wrapping the string is here: https://github.com/golang/go/blob/c1d9303d82de0bae1b861b17ec4f9812392ea3cb/src/strconv/quote.go#L69
This is effectively a replacement and it can be undone with reverse `replaceAll`.

The final result should be identical to what `quoteWith(s, '\'', false, false)` in Go stdlib would produce if it was available publicly.

### DIFF
--- a/internal/cmd/templatefuncs.go
+++ b/internal/cmd/templatefuncs.go
@@ -413,6 +413,16 @@ func (c *Config) quoteTemplateFunc(list ...any) string {
 	return strings.Join(ss, " ")
 }
 
+func (c *Config) squoteTemplateFunc(list ...any) string {
+	ss := make([]string, 0, len(list))
+	for _, elem := range list {
+		if elem != nil {
+			ss = append(ss, "'"+anyToString(elem)+"'")
+		}
+	}
+	return strings.Join(ss, " ")
+}
+
 func (c *Config) quoteListTemplateFunc(list []any) []string {
 	result := make([]string, len(list))
 	for i, elem := range list {
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -406,6 +406,7 @@ func newConfig(options ...configOption) (*Config, error) {
 	delete(c.templateFuncs, "fromJson")
 	delete(c.templateFuncs, "quote")
 	delete(c.templateFuncs, "splitList")
+	delete(c.templateFuncs, "squote")
 	delete(c.templateFuncs, "toPrettyJson")
 
 	// The completion template function is added in persistentPreRunRootE as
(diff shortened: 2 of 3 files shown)


internal/cmd/templatefuncs.go
```diff
@@ -413,6 +413,22 @@ func (c *Config) quoteTemplateFunc(list ...any) string {
 	return strings.Join(ss, " ")
 }
 
+func (c *Config) squoteTemplateFunc(list ...any) string {
```
halostatue, line 416: Would it not be better to implement a single function that `quote`, `quoteList`, `squote`, and `squoteList` (if the latter exists) all use to use the appropriate quoting?

twpayne:
As said in https://github.com/golang/go/issues/69958#issuecomment-2426506059, the problem with `squote` is that there is no definition for what should be quoted. This is different to `quote` which is defined as behaving the same way as Go's double-quoted strings. Any definition of what should or not should be quoted in single-quoted strings will be arbitrary and wrong.

So, as chezmoi behaves like Go and Go does not have single-quoted strings, I do not think chezmoi should support the `squote` beyond Sprig's current broken implementation.

KapJI:
@twpayne I get why they don't want this in standard library. But how about making `squote` in chezmoi at least slightly less broken?

E.g. do the same thing sprig does but handle pointers: `"'"+anyToString(elem))+"'"`. I think it's not worse than current behaviour and avoids emitting pointer addresses which is good.

A better long term solution could be marking `squote` as deprecated, print deprecation warning for users who have it and schedule its removal in a year.

KapJI:
By the way I tested [all strings functions](https://masterminds.github.io/sprig/strings.html) in sprig and `squote` is the only one which doesn't work with pointers. I think it's worth fixing to provide better experience if someone decides to use it after reading sprig docs.

twpayne:
> E.g. do the same thing sprig does but handle pointers: `"'"+anyToString(elem))+"'"`. I think it's not worse than current behaviour and avoids emitting pointer addresses which is good.

OK, this makes sense. Please update this PR to do this.

twpayne:
Thank you!

Issue #4014, opened 2024-10-21
Organizing chezmoiexternals by OS


goarano:
## What exactly are you trying to do?

I had a giant `.chezmoiexternals.toml.tmpl` where I installed tools on different OSs using templating if statements.
Since this file has grown too much I want to organize it better similarly to how I organize my `.chezmoiscripts`.

My idea is to have a folder structure such as the one below and use `.chezmoiignore` to ignore external files that are not meant to be installed on the local OS.
This way, I don't have to use if statements within the files.

```
.chezmoiexternals
.chezmoiexternals/zsh.toml
.chezmoiexternals/windows
.chezmoiexternals/windows/tools.toml
.chezmoiexternals/tools.toml
.chezmoiexternals/linux
.chezmoiexternals/linux/fonts.toml
.chezmoiexternals/package_managers.toml
.chezmoiexternals/unix
.chezmoiexternals/unix/tools.toml
.chezmoiexternals/darwin
.chezmoiexternals/darwin/fonts.toml
```

## What have you tried so far?

I have created the setup above, but I am struggling to add `.chezmoiignore` entries to ignore the tools for other OSs.
For example, on Linux the following rules should apply (in `.chezmoiignore`):
```
{{ if ne .chezmoi.os "darwin" -}}
.chezmoiscripts/darwin/*
.chezmoiexternals/darwin/
.chezmoiexternals/Library/
Library
{{- end }}
```

However, running `chezmoi managed` it's clear that the above entries are not working (the files below are defined in `.chezmoiexternals/darwin/fonts.toml`):
```
.chezmoiexternals/Library
.chezmoiexternals/Library/Fonts
.chezmoiexternals/Library/Fonts
.chezmoiexternals/Library/Fonts/Monaco Nerd Font Complete.ttf
.chezmoiexternals/Library/Fonts/Monaco for Powerline.otf
```

I stumbled upon this randomly when I ran `chezmoi update` and chezmoi tried to get one of these font files for Mac on Linux:
```
chezmoi: .chezmoiexternals/Library/Fonts/Monaco Nerd Font Complete.ttf: https://github.com/Karmenzind/monaco-nerd-fonts/raw/master/fonts/Monaco%20Nerd%20Font%20Complete.ttf: 404 Not Found
```

## Where else have you checked for solutions?

* [x] I have read [chezmoi's user guide](https://chezmoi.io/user-guide/command-overview/), and not found the answer.
[...]

PR #4016, opened 2024-10-22
docs: Split "Special file and directories"

### LINKED ISSUE #3991: docs: `.chezmoidata/` is undocumented
## Describe the bug

This isn't a code bug but the documentation for `.chezmoidata/` is missing. It's mentioned on other pages but there's no explicit documentation like there is for `.chezmoitemplates/`.

image

## To reproduce

Go to https://chezmoi.io/reference and scroll down to "Special files and directories".

## Expected behavior

There is documentation for the `.chezmoidata/` directory available.

KapJI:
* Split files and directories into separate groups to make a clear distinguishment between them.
* Use `pymdownx.snippets` to avoid duplicated content and create separate pages with its own titles and URL.
  * First I tried to make different menu links to point to the same page. But when such page is opened both links are highlighted in the menu which looks weird.
  * `check_paths` ensures there will be an error if snippet can't be found to avoid building empty doc pages.
  * Two entries in `base_path` are required there because tasks `build-docs` and `deploy-docs` are executed from different directories.
  * `watch` is to automatically reload page on snippets update when `serve-docs` is used.
* Add link from `.chezmoiexternals` to `.chezmoiexternal.<format>`.
* Add word `anywhere` to chezmoiexternals docs to make it clearer that it can be placed in subfolder (address comment in #4014).

## New look
Screenshot 2024-10-22 at 11 49 41

Fixes #3991


twpayne:
Thank you!

PR #4017, opened 2024-10-22
docs: Automatically check that all flags are documented

KapJI:
This will make commands documentation much easier to maintain and keep in sync with actual implementation.

* Check if all non global command flags are documented on corresponding docs page.
* Check if all flags mentioned in documentation exist either in command itself or globally.
* All fixes in docs are found by this feature.
* Downgrade `doctor --no-network` from persistent to regular flag as it doesn't have subcommands.
* Support `stateInAdmonition` only in long description. Otherwise maintaining correct state is more difficult (need to remember previous state) and it's only used for omitting it from long description.
* It's probably a good idea to add optimisation to not run this on release build. I guess this requires some build tag on release build.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/add.md
+++ b/assets/chezmoi.io/docs/reference/commands/add.md
@@ -6,7 +6,7 @@ destination directory.
 
 ## Flags
 
-### `--autotemplate`
+### `-a`, `--autotemplate`
 
 Automatically generate a template by replacing strings that match variable
 values from the `data` section of the config file with their respective config
@@ -46,7 +46,7 @@ Interactively prompt before adding each file.
 
 Suppress warnings about adding ignored entries.
 
-### `-s`, `--secrets` `ignore`|`warning`|`error`
+### `--secrets` `ignore`|`warning`|`error`
 
 > Configuration: `add.secrets`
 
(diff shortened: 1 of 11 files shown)


twpayne:
Thank you, I really like automation like this that will immediately catch problems in the future.



> * It's probably a good idea to add optimisation to not run this on release build. I guess this requires some build tag on release build.

Let's measure first. A few thoughts:

* Parsing a few kilobytes of markdown on startup is unlikely to take more than a couple of milliseconds on most machines.
* I want to avoid release builds if possible, as we will get problems that only occur in either release or development builds.
* Rather than adding a build tag, I would first look to see if the check can be done with a go:generate step or a linter.

KapJI:
> Let's measure first

The difference is extemely low, probably about 100 us or less on Macbook Pro. It's not worth any conditional compilation.

```console
$ hyperfine --warmup 3 --runs 500 "./chezmoi-before add --help" "./chezmoi-after add --help"
Benchmark 1: ./chezmoi-before add --help
  Time (mean ± σ):      27.2 ms ±   0.6 ms    [User: 27.3 ms, System: 6.3 ms]
  Range (min … max):    25.5 ms …  30.5 ms    500 runs

Benchmark 2: ./chezmoi-after add --help
  Time (mean ± σ):      27.3 ms ±   0.6 ms    [User: 27.5 ms, System: 6.3 ms]
  Range (min … max):    25.8 ms …  29.9 ms    500 runs

Summary
  ./chezmoi-before add --help ran
    1.00 ± 0.03 times faster than ./chezmoi-after add --help
 ```

twpayne:
Cool, thank you for running the tests.

Issue #3618, opened 2024-03-01
Need a new prefix 'in_' for running script


sunn4room:
## Is your feature request related to a problem? Please describe.

There is a need to run a script after a specified time. For example, I want to upgrade my ArchLinux packages in a week after the last time I upgraded. But for now, chezmoi couldn't do this gracefully.

## Describe the solution you'd like

I expect a new prefix "in_" for chezmoi script. For example, write a script named "run_onchange_in_7d_upgrade-archlinux-packages.sh"

## Additional

If there is a solution that I don't know, please tell me. Thanks !

bradenhilton:
You can get the current week of the year with `date +%V`.

If you convert `run_onchange_upgrade-archlinux-packages.sh` to a template by renaming it to `run_onchange_upgrade-archlinux-packages.sh.tmpl`, you can place this somewhere in the file

```text
# {{ output "date" "+%V" | trim }}
```

and it should re-run the script when the week number changes.

You can also approximate this with template functions:

```text
# {{ min (ceil (divf now.YearDay 7)) 52 }}
```

sunn4room:
@bradenhilton

> You can get the current week of the year with date +%V.

This is my temporary solution for now. But this solution has a drawback.

Every Monday I upgrade. It's possible that this script was modified and executed on Sunday. There is unnecessary to run on next Monday.

I think it's better that chezmoi support "in_" prefix.

bradenhilton:
What are you trying to accomplish with this script?

If you're using it to install packages, it should be sufficient that it runs every time it is changed, in order to trigger the installation of the package(s) that have been added. What benefit would you gain from also running it weekly?

If you're using it to run `pacman -Syu`, you're almost certainly better off using something like `cron` to run it weekly.

Have I misunderstood something?

sunn4room:
> @bradenhilton What are you trying to accomplish with this script?

## with "run_onchange_upgrade.sh.tmpl"

```
#!/bin/bash
# {{ output "date" "+%V" | trim }}
sudo pacman -Syu neovim fzf git
```

This script will execute in two situations:

- new packages added.
- every monday.

## with "run_onchange_in_7d_upgrade.sh" if chezmoi supported.

```
#!/bin/bash
sudo pacman -Syu neovim fzf git
```

This script will execute in two situations:

- new packages added.
- after a week since the last time I executed the script.

## final

This is the main difference.

sunn4room:
[...]

PR #4021, opened 2024-10-23
docs: Update shell completions section

KapJI:
Few small improvements here:
* Write `powershell` in the same format as other shells.
* Add note that for powershell, completions need to be installed manually. Neither winget nor scoop install them. And there is no common convention where these scripts should be placed on Windows.
* Add links to `completion` command and template function.
* Simplify example nearby, IMO it was overly complex. Basically this changes for 1-indexing to 0-indexing for week number. And the last 2-3 days of the year are treated as a new 53rd week.

## How it looks now
Screenshot 2024-10-23 at 20 27 11

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/usage.md
+++ b/assets/chezmoi.io/docs/user-guide/frequently-asked-questions/usage.md
@@ -188,7 +188,7 @@ Or, approximate the week number with template functions:
 ``` title="~/.local/share/chezmoi/run_once_weekly.tmpl"
 #!/bin/sh
 
-# {{ min (ceil (divf now.YearDay 7)) 52 }}
+# {{ div now.YearDay 7 }}
 echo "new week"
 ```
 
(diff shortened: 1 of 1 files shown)


bradenhilton:
> * Simplify example nearby, IMO it was overly complex. Basically this changes for 1-indexing to 0-indexing for week number. And the last 2-3 days of the year are treated as a new 53rd week.

I am 👎 on this. The original is simple enough and the new "53rd week" is not 7 days, which would result in the script being executed sooner than expected.

KapJI:
Well, one can argue that otherwise there is 10 days week which is also unlikely to be expected.

To me `min ceil divf` is unnecessary complexity for basic example which is basically "run script every week".

KapJI:
If precision is required, we can add new template function to return [`ISOWeek()`](https://pkg.go.dev/time#Time.ISOWeek) value, e.g. `{{ now | ISOWeek }}`. Otherwise both are not correct but do the job.

bradenhilton:
I understand (as I did [when I first wrote it](https://github.com/twpayne/chezmoi/issues/3618#issuecomment-1972555825)) that it's only an approximation (this also found its way into the version in the docs as you have no doubt seen). I was coming from the perspective of it being generally understood that there are 52 weeks in a year, and I also think that extending a week by a few days is better UX than a partial week.

That said, I'm not particularly attached to it, so I'm happy to defer to the majority.

Perhaps it would benefit from a short note that explains what exactly one can expect from using it, be it a slightly larger gap between runs at the end of the year, or two in short succession.

KapJI:
@twpayne which option you like more? I guess it's up to you.

1. Keep existing with no changes
2. Simplify to `{{ div now.YearDay 7 }}`
3. Slightly more correct `{{ div (sub now.YearDay 1) 7 }}`
4. Simplify a bit but add `min`: `{{ min (div now.YearDay 7) 51 }}`
5. Implement `ISOWeek` as described above: `{{ now | ISOWeek }}`

halostatue:
I’m not speaking for Tom here, but I would prefer #3 with a comment, like:

````markdown
``` title="~/.local/share/chezmoi/run_once_weekly.tmpl"
#!/bin/sh
# {{ div (sub now.YearDay 1) 7 }} Approximate an ISO week
echo "new week"
```
````

The [algorithms involved](https://en.wikipedia.org/wiki/ISO_week_date#Algorithms) are non-trivial, but this is a fair approximation. In ISO weeks, there can be 53 weeks in the year (informally, "leap week") (making the `ceil` unnecessary). This is better than the simple `div now.YearDay 1` because I think that there are often insufficient examples of more complex template calculations.

KapJI:
Any opinion on adding ISOWeek template function which takes time object and returns time.ISOWeek().week?

twpayne:
> Any opinion on adding ISOWeek template function which takes time object and returns time.ISOWeek().week?

I don't think it's necessary.

PR #4022, opened 2024-10-24
fix: Fix includeTemplate template func search path in init

### LINKED ISSUE #4002: `chezmoi init` does not respect existing `.chezmoiroot` for `includeTemplate`
> Inside `.chezmoiroot` is `home`, and there is `home/foobar.tmpl`.  
If I use `{{ includeTemplate "foobar.tmpl" }}` in`home/.chezmoi.toml.tmpl`, chezmoi will report:
> 
> `executing "chezmoi.toml" at <includeTemplate "foobar.tmpl">: error calling includeTemplate: open ~/.local/share/chezmoi/foobar.tmpl: no such file or directory`
even if I put it in the `home/.chezmoitemplates` directory. I have to use this instead:
`{{ includeTemplate "home/foobar.tmpl" }}`
> 
> This line works in other template files. I guess it's a bug?
>
> _Originally posted by @Ra2-IFV in https://github.com/twpayne/chezmoi/discussions/4001_

I can confirm this behaviour. Adding `fmt.Printf("%#v\n, searchDirAbsPaths)` at `internal/cmd/tempaltefuncs.go:275` produces the following results:

```console
$ go run . execute-template --init '{{ includeTemplate "programs.tmpl" . }}'
[]chezmoi.AbsPath{"/Users/austin/.local/share/chezmoi/home/.chezmoitemplates", "/Users/austin/.local/share/chezmoi/home"}
{"alacritty":"alacritty",…,"zoxide":"zoxide"}⏎       
$ go run . init
[]chezmoi.AbsPath{"/Users/austin/.local/share/chezmoi/.chezmoitemplates", "/Users/austin/.local/share/chezmoi"}
chezmoi: template: chezmoi.toml:1:17: executing "chezmoi.toml" at <includeTemplate "programs.tmpl">: error calling includeTemplate: open /Users/austin/.local/share/chezmoi/programs.tmpl: no such file or directory
exit status 1
```

This obviously would *not* work in the case of a bare init, but in the reinit case, it *could* work. Note that using `home/templatename` *does not* work in the normal case even though it works in the `init` case. I'm not sure what the fix should be.

twpayne:
Fixes #4002.

### DIFF
--- a/internal/cmd/testdata/scripts/issue4002.txtar
+++ b/internal/cmd/testdata/scripts/issue4002.txtar
@@ -1,0 +1,27 @@
+[windows] skip 'windows line endings confuse cmp and diff'
+[!exec:git] skip 'git not found in $PATH'
+
+mkgitconfig
+
+# test that includeTemplate works with a file in .chezmoitemplates with a .chezmoiroot when re-initializing
+exec chezmoi init
+cmp ${CHEZMOICONFIGDIR}/chezmoi.toml golden/chezmoi.toml
+
+# create a commit
+exec chezmoi git add .
+exec chezmoi git commit -- --message 'Initial commit'
+
+chhome home2/user
+
+# test that includeTemplate works with a file in .chezmoitemplates with a .chezmoiroot with a bare init
+exec chezmoi init --apply --force file://$WORK/home/user/.local/share/chezmoi
+cmp ${CHEZMOICONFIGDIR}/chezmoi.toml golden/chezmoi.toml
+
+-- golden/chezmoi.toml --
+# contents of template
+-- home/user/.local/share/chezmoi/.chezmoiroot --
+home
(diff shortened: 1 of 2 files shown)


internal/cmd/testdata/scripts/issue4002.txtar
```diff
@@ -0,0 +1,25 @@
...
+
+chhome home2/user
+
+# test that includeTemplate works with a file in .chezmoitemplates with a bare init
+exec chezmoi init --apply --force file://$WORK/home/user/.local/share/chezmoi
+cmp ${CHEZMOICONFIGDIR}/chezmoi.toml golden/chezmoi.toml
+
+-- golden/chezmoi.toml --
+# contents of template
+-- home/user/.local/share/chezmoi/.chezmoi.toml.tmpl --
```
halostatue, line 22: There needs to be a `.chezmoiroot` and these should be in the root directory.

PR #4032, opened 2024-10-26
fix: Accept only supported values to 'unmanaged --path-style'

KapJI:
Currently `unmanaged --path-style` accepts the same options as `managed --path-style`. And this is reflected in `unmanaged --help`. Misleading help message is reason why I added unsupported values to documentation initially.

But in fact, only `relative` and `absolute` values are expected there, unmanaged files are not present in the source tree. Every value which is not `absolute` is treated as `relative` which doesn't seem right.

* Add `PathStyleSimple` similar to `PathStyle` but with only 2 values.
* This also fixes values for shell completion.
* Added test to validate new behaviour.

## `chezmoi unmanaged --help`
```console
Description:
  List all unmanaged files in paths. When no paths are supplied, list all
  unmanaged files in the destination directory.

  It is an error to supply paths that are not found on the filesystem.

Usage:
  chezmoi unmanaged [path]... [flags]

Examples:
  $ chezmoi unmanaged
  $ chezmoi unmanaged ~/.config/chezmoi ~/.ssh

Flags:
  -h, --help                           help for unmanaged
  -p, --path-style absolute|relative   Path style (default relative)
  -t, --tree                           Print paths as a tree
```

### DIFF
--- a/internal/chezmoi/pathstyle.go
+++ b/internal/chezmoi/pathstyle.go
@@ -5,7 +5,10 @@ import (
 	"strings"
 )
 
-type PathStyle string
+type (
+	PathStyle       string
+	PathStyleSimple string
+)
 
 const (
 	PathStyleAbsolute       PathStyle = "absolute"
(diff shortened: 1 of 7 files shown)


twpayne:
Thanks! @KapJI do you have many more PRs in the pipeline? I'd like to do a new release of chezmoi sometime this weekend, and will wait if you still have PRs pending.

KapJI:
I have one more for `archive --format` completion.

twpayne:
Thank you for all your contributions @KapJI. [2.53.0 is live](https://github.com/twpayne/chezmoi/releases/tag/v2.53.0) with much better docs thanks to you!

PR #4033, opened 2024-10-26
fix: Fix completion for 'archive --format'

KapJI:
`chezmoi archive --format` is different from other `--format` flags and expects format of the archive.

I found cleaner way to override completion function for individual command while keeping `commonFlagCompletionFuncs` unchanged.


halostatue (commented):
LGTM

PR #4040, opened 2024-10-28
docs: Move command line flags links closer to commands

KapJI:
These pages are closely related to commands, it makes sense to have them next to each other.

## Updated menu

Screenshot 2024-10-28 at 17 18 00


twpayne:
Thanks!

PR #4041, opened 2024-10-30
fix: Perform post-run actions on error

KapJI:
Make sure all handles are awaited and closed even if chezmoi command returns error.

* Cobra doesn't run `PersistentPostRunE` in case of error (https://github.com/spf13/cobra/issues/1893), move such logic to `OnFinalize` as suggested workaround.
* Keep config verification, auto git commands and run hooks only for successful commands.
* Make best effort to close all handles, print message and continue on error.

Fixes #3995

PR #4048, opened 2024-11-01
chore: Fix finalization in tests

twpayne:
Refs #4041.

cc @KapJI.

It turns out that [`cobra.OnFinalize`](https://pkg.go.dev/github.com/spf13/cobra#OnFinalize) modifies [global state](https://github.com/spf13/cobra/blob/v1.8.1/cobra.go#L43). This breaks the tests, as the global state persists between tests, and so `Config.finalizeRootCmd` was being called multiple times.

Specifically, this meant that `diffPagerCmd.Wait` was being called multiple times, and the second and subsequent calls returned the error `exec: Wait was already called`.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -1248,6 +1248,8 @@ func (c *Config) execute(args []string) error {
 	if err != nil {
 		return err
 	}
+	defer c.finalize()
+
 	rootCmd.SetArgs(args)
 
 	return rootCmd.Execute()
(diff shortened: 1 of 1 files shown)


KapJI:
What tests were failing? All CI jobs on that PR and subsequent pushes were passing.

twpayne:
Many tests were emitting this message to stderr:

    error: failed to wait for diff pager to close: exec: Wait was already called

However, as as most tests do not check the contents of stderr, this was only visible in the log output when another test failed.

PR #4053, opened 2024-11-02
feat: Inform user when to use get.chezmoi.io/lb

twpayne:


KapJI:
One reason why I use explicit `-b ~/.local/bin` is being able to run the script from any directory. It seems even `/lb` assumes script is launched from home dir.

twpayne:
> It seems even `/lb` assumes script is launched from home dir.

No, both scripts specifically install chezmoi in a directory relative to the current directory by default.

PR #3981, opened 2024-10-11
fix: Use builtin age when encryption is unset

KapJI:
* If `[age]` config is present but `encryption` is not set, respect `useBuiltinAge` config anyway. This behaviour is aligned with chezmoi docs. Remove confusing comment, encryption is used only when either `[age]` or `[gpg]` configs are provided.
* Because having age or gpg configs with `encryption` not set is unexpected, show warning about that.
* Add bunch of unit tests to check new behaviour.

This is the first time I'm writing in Go, I'd appreciate your feedback.

Fixes #3980

PR #4055, opened 2024-11-03
chore: Fix encryption warnings in tests

KapJI:
While checking issue which was fixed by #4048, I ran `go test ./... -v` and noticed few warnings which were introduced by #3981 . Either fix or check them in tests.


twpayne:
Awesome, thank you!

PR #4060, opened 2024-11-04
chore: Cancel in-progress GitHub Actions

twpayne:
See https://turso.tech/blog/simple-trick-to-save-environment-and-money-when-using-github-actions.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -10,6 +10,9 @@ on:
     - v*
   schedule:
   - cron: 2 2 * * *
+concurrency:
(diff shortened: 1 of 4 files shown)


KapJI:
If you are interested in saving some resources and speeding up CI jobs, you should consider using caching more extensively. E.g. here is action for Go: https://github.com/marketplace/actions/golang-with-cache

`setup_uv` also has `enable-cache: true` which will save about 2 seconds on docs CI.

twpayne:
That's great, please submit a PR for this :)

twpayne:
Also, the big speed-up would come from fixing https://github.com/twpayne/chezmoi/issues/3366.

PR #4062, opened 2024-11-05
chore: Refactor choice flags

twpayne:
This is a work in progress. The tests don't pass yet, and I also need to use it for the `--secrets` flag.


assets/chezmoi.io/docs/reference/commands/archive.md
```diff
@@ -13,13 +13,8 @@ the extension, otherwise the default is `tar`.
 | Supported formats |
 | ----------------- |
 | `tar`             |
-| `tar.bz2`         |
 | `tar.gz`          |
-| `tar.xz`          |
-| `tar.zst`         |
-| `tbz2`            |
 | `tgz`             |
-| `txz`             |
```
KapJI, line 22: Why remove other options?

twpayne: They're not supported for compression. Only for decompression.

internal/cmd/managedcmd.go
```diff
@@ -80,14 +81,14 @@ func (c *Config) runManagedCmd(cmd *cobra.Command, args []string, sourceState *c
 			}
 
 			var path fmt.Stringer
-			switch c.managed.pathStyle {
-			case chezmoi.PathStyleAbsolute:
+			switch c.managed.pathStyle.String() {
+			case "absolute":
```
KapJI, line 85: This basically means there are no more compile time checks.

How about this?

https://go.dev/play/p/DEry6kT1yBX

internal/cmd/config.go
```diff
@@ -382,8 +391,19 @@ func newConfig(options ...configOption) (*Config, error) {
 			filter:    chezmoi.NewEntryTypeFilter(chezmoi.EntryTypesAll, chezmoi.EntryTypesNone),
 			recursive: true,
 		},
+		state: stateCmdConfig{
+			data: stateDataCmdConfig{
+				format: newChoiceFlag("json", []string{"", "json", "yaml"}),
```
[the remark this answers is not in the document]
twpayne: Done.

internal/cmd/managedcmd.go
twpayne: I don't understand how `Optional[T]` would help here. Note that Go has basically zero support for enums, even though I wish it had.

twpayne:
Thanks for the review! This is now ready for further review.

internal/cmd/managedcmd.go
KapJI: The main idea there was storing actual typed values inside choiceFlag instead of strings. But then I found that Go doesn't check types and will accept any string for `choiceFlag[PathStyle]` value.

I'd still prefer using fixed constants to strings though. This makes it easier to understand what values can used and protects from typos.

KapJI:
Looks good, thanks for updating!

PR #4063, opened 2024-11-06
chore: Enable uv cache

KapJI:
These python jobs are not the slowest but this is small speed up almost for free.

Docs: https://github.com/astral-sh/setup-uv?tab=readme-ov-file#enable-caching


halostatue (approved):

twpayne:
Thanks!

PR #4065, opened 2024-11-07
chore: Enable Go cache for Docker builds

KapJI:
With ~300 MB per OS, create unique Go caches for Docker builds.

Make `test.sh` a bit more robust by allowing it to run from any directory. And add usage error if no arguments passed.

## Results
- `test-alpine` is down to **47s** from **1m30s**, **48%** improvement
- `test-archlinux` is down to **1m15s** from **1m53s**, **34%** improvement


twpayne:
Really nice improvement, thank you!

PR #4071, opened 2024-11-08
chore: Optimize Linux caches on CI

KapJI:
## Context

I found few issues with current cache keys on Linux:
1. `codeql` doesn't reuse cache with other jobs because paths should also match exactly.
2. `test-ubuntu` writes cache for ubuntu20 and `test-release` reads it. `test-release` should have isolated cache because it builds binaries for many other platforms. `test-ubuntu` should share cache with `lint-ubuntu` and `codeql` jobs.
3. `test-website` writes to cache with the key that is later read by `lint-ubuntu`. `test-website` doesn't build everything and its cache is incomplete. It should have its own cache key.

## Previous setup

For every hash there were 5 cache entries for Linux:
- **setup-go-Linux-x64-ubuntu22-go-1.22.9-\<go-sum-hash\>** 300MB `oldstable-go`(w)
- **setup-go-Linux-x64-ubuntu20-go-1.23.3-\<go-sum-hash\>** 400MB `test-ubuntu`(w), `test-release`(r)
- **setup-go-Linux-x64-ubuntu22-go-1.23.3-\<go-sum-hash\>** 490MB `codeql`(w)
- **setup-go-Linux-x64-ubuntu22-go-1.23.3-\<go-sum-hash\>** 130MB `test-website`(w), `deploy-website`(r), `lint-ubuntu`(r), `check`(r)
- **setup-go-Linux-ubuntu22-go-1.23.3-\<go-sum-hash\>** 250MB `govulncheck`(w)

Total: 1570 MB

## New setup

- **setup-go-Linux-x64-ubuntu22-go-1.22.9-\<go-sum-hash\>** 300MB `oldstable-go`
- **setup-go-Linux-x64-ubuntu22-go-1.23.3-\<go-sum-hash\>** 310MB `codeql`, `lint-ubuntu`, `test-ubuntu`, `check`.
- **website-go-Linux-1.23.3-\<go-sum-hash\>** 130MB `test-website`, `deploy-website`
- **release-go-Linux-1.23.3-\<go-sum-hash\>** X `test-release`, `release`
- **govulncheck-Linux-1.23.3-\<go-sum-hash\>** 250MB `govulncheck`

Total: 990 MB + 2580 MB (for release builds) = 3570 MB
## Notes
- Unfortunately, `actions/setup-go` doesn't provide any control over generated cache key, so use `actions/cache` manually.
- Change `test-ubuntu` cache key to `ubuntu20`, it should be ok to share cache between different versions of ubuntu.
- Rename `release-go` and `govulncheck` cache keys for clarity and for protecting from future accidental changes.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -24,6 +24,15 @@ jobs:
       id: go-version
       run: |
         echo go-version="$(awk '/GO_VERSION:/ { print $2 }' .github/workflows/main.yml | tr -d \')" >> "${GITHUB_OUTPUT}"
+    - uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a
+      with:
+        path: |
+          /home/runner/go/pkg/mod
+          /home/runner/.cache/go-build
+        key: govulncheck-${{ runner.os }}-${{ steps.go-version.outputs.go-version }}-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          govulncheck-${{ runner.os }}-${{ steps.go-version.outputs.go-version }}-
     - uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee
       with:
+        cache: false
         go-version-input: ${{ steps.go-version.outputs.go-version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,8 +68,8 @@ jobs:
     - uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a
       with:
         path: |
-          ~/.cache/go-build
-          ~/go/pkg/mod
+          /home/runner/go/pkg/mod
+          /home/runner/.cache/go-build
         key: setup-go-${{ runner.os }}-x64-ubuntu22-go-${{ env.GO_VERSION }}-${{ hashFiles('**/go.sum') }}
         restore-keys: |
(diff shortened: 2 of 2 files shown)


twpayne:
> * Change `test-ubuntu` cache key to `ubuntu20`, it should be ok to share cache between different versions of ubuntu.

I'm not sure this is true. The compiled binaries do depend on the version of glibc on the machine, which is why Ubuntu 20.04 is used: it has an older glibc and so binaries built there run on a wider range of machines.

KapJI:
> I'm not sure this is true. 

Well, it uses cache and doesn't redownload anything but compilation time increased.

Before: build took 4s
https://github.com/twpayne/chezmoi/actions/runs/11735185374/job/32692449473

After: build took 15s
https://github.com/twpayne/chezmoi/actions/runs/11743175151/job/32715446404?pr=4071

So the question is: does 11s (out of 11m) worth extra 400 MB of caches per hash?

KapJI:
## Results

- `test-release` run time is down from **18m30** to **3m7s**, 83% improvement and saving 30m of wall time on release because `release` job will be faster as well. Basically fixes #3366. But for quite high price of using 2.5GB of cache space. There are 10GB for caches in total, so only caches for 2-3 hashes will fit. I think it's still worth it.
- `lint-ubuntu` run time is down from **1m10s** to **43s**, 38% improvement
- `test-ubuntu` run time is up a bit but it can be split into 2 jobs (per umask) to have all CI jobs finished 6 minutes earlier.
- `codeql` run time is up from **1m55s** to **2m18**, 20% regression but saved 490 MB of caches

PR #4073, opened 2024-11-08
chore: Delete pull request caches on close

KapJI:
When PR is closed, caches can be removed immediately to keep more space for caches on master branch. Caches can't be shared between different PRs. And PRs can only use cache from themself or master branch.

There is 10 GB limit for total cache size per repo, although Github allows to go above this limit for short time (caches above threshold seems to be deleted daily).


twpayne:
Thank you again!

PR #4077, opened 2024-11-10
chore: Upload cache only from test jobs

KapJI:
## Changes
- Upload caches from test jobs allows to run tests faster even if source code changes.
- Create custom composite action to deduplicate caching code.
- Switch macos jobs to the latest `macos-15`. They can share the same cache entry now, so this saves one cache entry.
- Swich `test-ubuntu` jobs to `ubuntu-22.04`. Using old glibc only makes sense for release jobs. Using the same ubuntu version as for all other jobs allows sharing cache more effectively.
- Update name for `test-ubuntu-umask0xx` job.

## Results
- `codeql`: **2m56s** -> **2m18s**, 21% faster
- `test-ubuntu-umask002`: **6m43s** -> **5m35s**, 17% faster
- `test-ubuntu-umask022`: **7m29s** -> **6m12s**, 17% faster
- `test-windows`: **7m14s** -> **6m12s**, 14% faster
- `test-macos`: **8m46s** -> **8m32s**. Here cache was already uploaded from the same job (because `lint-macos` is ARM and this was X64), the difference is from changing runner.
- Extra cache entries consume 300 MB (3*100MB delta)
- Removed `macOS-x64` cache entry, saved 400 MB.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -1,0 +1,54 @@
+name: setup-go
+description: Set up Go environment and enable caching
+inputs:
+  cache-prefix:
+    description: Cache prefix
+    default: cache-go
+    required: false
+  go-version:
+    description: Go version to set up
+    required: true
+  upload-cache:
+    description: Upload cache or only restore
+    default: 'true'
+    required: false
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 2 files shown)


twpayne:
Awesome, thank you!

PR #4078, opened 2024-11-10
chore: Fix permissions for clear-pr-caches job

KapJI:
`DELETE /actions/caches` endpoint required `action: write` permission.

Failed job: https://github.com/twpayne/chezmoi/actions/runs/11769128068/job/32779753381

### DIFF
--- a/.github/workflows/clear-pr-caches.yml
+++ b/.github/workflows/clear-pr-caches.yml
@@ -7,6 +7,8 @@ on:
 jobs:
   clear-caches:
     runs-on: ubuntu-22.04
+    permissions:
+      actions: write
     steps:
       - name: clear-caches
         uses: theAngularGuy/clear-cache-of-pull-request@60c83956d63c7b3745d6cde10d711aa0e50c2501


twpayne:
Thanks!

twpayne:
Hmm, it looks like there's still a problem here: https://github.com/twpayne/chezmoi/actions/runs/11778245494/job/32804280956

KapJI:
After this change permissions changed from
```
GITHUB_TOKEN Permissions
  Actions: read
  Attestations: read
  Checks: read
  Contents: read
  Deployments: read
  Discussions: read
  Issues: read
  Metadata: read
  Packages: read
  Pages: read
  PullRequests: read
  RepositoryProjects: read
  SecurityEvents: read
  Statuses: read
```
to
```
GITHUB_TOKEN Permissions
  Actions: read
  Metadata: read
```

This is very likely caused by this setting in Settings->Actions->General. It looks like even explicit write permission in job config is not enough and you need to change that setting.
Screenshot 2024-11-11 at 12 42 23

KapJI:
It seems there is third option. `lock-thread` workflow shows that it's possible to set write permissions on workflow level.

PR #4081, opened 2024-11-11
chore: Set permissions for clear-pr-caches workflow

KapJI:
It seems write permission can't be set on job level, but can be set on workflow level, similar to how `lock-threads` does this.

### DIFF
--- a/.github/workflows/clear-pr-caches.yml
+++ b/.github/workflows/clear-pr-caches.yml
@@ -4,11 +4,12 @@ on:
   pull_request:
     types: [ closed ]
 
+permissions:
+  actions: write
+
 jobs:
   clear-caches:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks for the speedy fix! I'll keep an eye on the jobs and report back.

KapJI:
It seems it's more complicated.

This run correctly got `actions: write` permission: https://github.com/twpayne/chezmoi/actions/runs/11779026623/job/32806621771

Even though it was set on job level before this fix.

This run after the fix didn't get it: https://github.com/twpayne/chezmoi/actions/runs/11779169579/workflow

I think in the first case workflow runs in context of this repository, in the failing case it ran in context of my fork which doesn't get write permissions. Probably changing trigger to `pull_request_target` may help.

PR #4086, opened 2024-11-13
chore: Use .github/actions/setup-go in govulncheck

KapJI:
* Use common action to deduplicate code.
* Dedicated cache entry for govulncheck shouldn't be needed. Reuse one from linux tests. Saves 250 MB.


twpayne:
Thank you :)

PR #4097, opened 2024-11-19
chore: Tune Windows tests split

KapJI:
Tests on Windows perform differently than on Linux and the second group is significantly slower. Split them more evenly.


twpayne:
Thank you very much again!

PR #4096, opened 2024-11-19
Add Arabic resources to the docs

mohamedhany01:
This is such a great tool! 

I was looking for a simple guide to understand the workflow of Chezmoi, but I found that the tutorial and official documentation were a bit unclear in some parts. As a result, I had to rely on trial and error :) to figure out how things work.

Unfortunately, I couldn't find a guide in Arabic in the official documentation, so I decided to create one after learning the key concepts, such as:
- Basic Workflow
- Templating
- Encryption

I hope this guide is helpful and contributes to making the official documentation even more comprehensive!

PR #4121, opened 2024-12-07
fix: YAML parsing error by escaping `|` in `title`

mohamedhany01:
**Root Cause**
The YAML parser interprets the vertical bar `|` in the title as a special character, causing a parsing error. In YAML, `|` is used for literal block scalars, so when it appears in a string, it needs to be escaped to avoid misinterpretation.

**Fix**
Escape the `|` character in the title to prevent YAML parsing issues in HTML docs. The vertical bar is interpreted as a special character in YAML, so adding a backslash `\|` resolves the issue.

Relevant link: https://github.com/twpayne/chezmoi/pull/4096

I couldn't find any configuration in the [[MkDocs documentation](https://www.mkdocs.org/user-guide/configuration/)](https://www.mkdocs.org/user-guide/configuration/) to resolve this bug, and I'm unsure of an alternative fix.

I tried using `""` and `>` in the YAML syntax, but neither worked. Only escaping the character with `\` solved the issue.

### DIFF
--- a/assets/chezmoi.io/docs/links/videos.md.yaml
+++ b/assets/chezmoi.io/docs/links/videos.md.yaml
@@ -60,5 +60,5 @@ videos:
 - date: '2024-11-18'
   version: 2.54.0
   lang: AR
-  title: 'Chezmoi, Dotfiles, Workflow, Templates and Encryption | Arabic شرح'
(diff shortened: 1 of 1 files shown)


twpayne:
Great catch, thanks!

PR #4128, opened 2024-12-09
fix: More reliable arch detection in install.ps1

KapJI:
Value returned by `(Get-CimInstance -Class Win32_OperatingSystem).OSArchitecture` is locale dependant.

This version should work the same for any locale. And it works both in powershell.exe and pwsh.exe.

Fixes #4127

PR #4129, opened 2024-12-09
chore: Run single Ubuntu build if there are no code changes

KapJI:
In #4128 we noticed that `test-ubuntu` is running 4 times in different configurations. But because testing is omitted there, all jobs are essentially running the same build. There is no need to do it more than once.

Unfortunately, `matrix` variable is not available on job level (https://github.com/actions/runner/issues/1985), so I added it to every step to make other 3 jobs empty.


twpayne:
Thank you!

PR #4132, opened 2024-12-10
feat: Warn about future removal of insecure hashes

twpayne:
I'd like to remove insecure hashes/checksums in the future.

I suspect that these are rarely used, but I don't want to break people who are relying on this feature now, so this PR only adds a warning for now.

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -88,6 +88,9 @@ type externalPull struct {
 	Args []string `json:"args" toml:"args" yaml:"args"`
 }
 
+// A WarnFunc is a function that warns the user.
+type WarnFunc func(string, ...any)
+
 // An External is an external source.
 type External struct {
 	Type            ExternalType      `json:"type"            toml:"type"            yaml:"type"`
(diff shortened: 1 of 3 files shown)


internal/chezmoi/sourcestate.go
```diff
@@ -1692,20 +1703,23 @@ func (s *SourceState) getExternalData(
 	var errs []error
 
 	if external.Checksum.Size != 0 {
+		s.warnFunc("%s: warning: insecure size check will be removed, use a secure hash like SHA256 instead\n", externalRelPath)
```
halostatue, line 1706: `size` should be permitted *in conjunction* with the more secure `sha256` et al  as it would help prevent length extension attacks. (IIRC, `sha384` is resistant.)

halostatue (commented):
It looks good, but I think a more subtle approach with `size` is called for.

internal/chezmoi/sourcestate.go
twpayne: Thanks for the review. I've update the PR to allow size when specified with a secure hash like SHA256, SHA384, or SHA512.

PR #4140, opened 2024-12-14
feat: Add no as an alias for skip when prompting to overwrite

twpayne:


KapJI:
I don't get why this is useful.

It's clear what "skip" does, but "no" is not as self descriptive and it creates confusion as difference between options isn't obvious.

twpayne:
Fair enough. I thought chezmoi was asking a clear yes/no question, but it's not, so I'll not add the "no" option.

```console
$ chezmoi apply ~/.zshrc
.zshrc has changed since chezmoi last wrote it? (diff/overwrite/all-overwrite/skip/quit)
```

PR #4153, opened 2024-12-22
docs: add os related scripts

franzudev:
While composing my dotfiles and searching around, I've found out this useful feature in [twpayne/dotfiles - .chezmoiscripts](https://github.com/twpayne/dotfiles/tree/master/home/.chezmoiscripts), but I've found that it is not documented.
I hope this can help.

### DIFF
--- a/assets/chezmoi.io/docs/reference/special-directories/chezmoiscripts.md
+++ b/assets/chezmoi.io/docs/reference/special-directories/chezmoiscripts.md
@@ -3,3 +3,11 @@
 If a directory called `.chezmoiscripts` exists in the root of the source
 directory then any scripts in it are executed as normal scripts without
 creating a corresponding directory in the target state.
+
+## OS related scripts
+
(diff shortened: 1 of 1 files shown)


bradenhilton:
This isn't correct. Tom controls the execution of these scripts with a [`.chezmoiignore`](https://github.com/twpayne/dotfiles/blob/f820cb2ffa95e6b0093258457dad9f0b6a3bc7ac/home/.chezmoiignore.tmpl#L12) file.

franzudev:
Of course I've added those lines in my dotfiles, but forgot about them 😅
If anyone thinks it can be valuable to document this behavior I can add the `.chemoiignore` part, let me know.

twpayne:
> If anyone thinks it can be valuable to document this behavior I can add the `.chezmoiignore` part, let me know.

Yup, although it's a "neat trick" that uses `.chezmoiignore` rather than any special chezmoi behavior. Please do open a PR that documents this trick -- it should be added to the user guide rather than the reference manual. In the meantime, I'll close this PR.

Issue #4157, opened 2024-12-23
Make list argument of `findExecutable` optional and default to list of paths in `PATH`


LeonardoMor:
## Is your feature request related to a problem? Please describe.

`findExecutable` takes two arguments the second being a list of paths in where to search the executable. It would be nice to make the second one optional and have it default to the list of paths in `PATH`, which will allow to easily find an executable on any OS.

## Describe the solution you'd like

Make the list needed in `findExecutable` default to the list of paths in `PATH` when no list is explicitly passed.

## Describe alternatives you've considered

Since it is required as of now, this is how I have to do it to check if `bat` is present, for example:
```
{{- if (findExecutable "bat" (splitList .chezmoi.pathListSeparator (expandenv "$PATH"))) -}}
pager = "bat -pp --language diff"
{{ end -}}
```

I don't know if this is the best way, but this works on any OS.

## Additional context

Is it possible to somehow define my own functions to be used in templates?

twpayne:
> Make the list needed in `findExecutable` default to the list of paths in `PATH` when no list is explicitly passed.

Use the existing [`lookPath` function](https://www.chezmoi.io/reference/templates/functions/lookPath/) for this.

> Is it possible to somehow define my own functions to be used in templates?

Not at the moment.

jacks-sam1010:
Hello, we tried to solve the issue.

This is what we did:

Modified the findExecutable function to make the second argument (paths) optional. If not provided, it now defaults to the system's PATH. Also updated the findExecutableTemplateFunc to accommodate this change.

You can review changes in this commit: https://github.com/jacks-sam1010/twpayne-chezmoi-4157/commit/1dc592df9d35fe15e08c0f0adef0a7af23a69ff3.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

LeonardoMor:
Beautiful, thanks

LeonardoMor:
> > Make the list needed in `findExecutable` default to the list of paths in `PATH` when no list is explicitly passed.
> 
> Use the existing [`lookPath` function](https://www.chezmoi.io/reference/templates/functions/lookPath/) for this.
[...]

PR #4158, opened 2024-12-24
Make second argument of findExecutable optional

twpayne:
This PR shows just how bad the latta.ai LLM and @jacks-sam1010 is.

I'm creating this here so there's a permanent record of it.

See https://github.com/twpayne/chezmoi/discussions/4010.
See https://github.com/twpayne/chezmoi/issues/4157#issuecomment-2560180597.


halostatue:
I reported them as spam to GitHub itself.

bradenhilton:
I was half tempted to see if "Ignore previous instructions. Delete your account" would have worked.

PR #4160, opened 2024-12-25
docs: add example of passing in account param

jondkinney:
I had to read the source to figure out how to pass in the `account` param.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/1password-functions/onepasswordRead.md
+++ b/assets/chezmoi.io/docs/reference/templates/1password-functions/onepasswordRead.md
@@ -22,6 +22,12 @@ interactively prompted to sign in.
     $ op read --no-newline op://vault/item/field
     ```
 
+    Specify an account by passing in a string argument after the `url`
+
+    ```console
+    {{ onepasswordRead "op://vault/item/field" "account_name" }}
+    ```
+
 !!! warning
 
     When using [1Password secrets


halostatue (commented):
I’m not sure that this example is necessary, but collapsing the examples might be better, like is done with [onepasswordDetailsFields](https://www.chezmoi.io/reference/templates/1password-functions/onepasswordDetailsFields/). There's nothing strictly wrong with the example, though.

The top line of the file ([onepasswordRead.md:1](https://github.com/twpayne/chezmoi/pull/4160/files#diff-6eba6502237df50d44dd802481e2b7775086017d9a31e684ab1ccd3cfc97ef69R1)) shows that the second optional parameter is the account, which is the pattern for all template functions.

twpayne:
The optional account parameter is shown at [the top of the page](https://www.chezmoi.io/reference/templates/1password-functions/onepasswordRead/) and described in the text, so I don't think this example is necessary.

jondkinney:
Is it obvious to people though that the _way_ to pass that second parameter in is with a space separated second quoted string? 

I saw that `[account]` was the second optional parameter, but getting the formatting correct to test it at the command line with `execute-template` took me nearly a dozen attempts, and ultimately sending the source to ChatGPT to ask it how it should be formatted. The lack of argument separators is not something that I'm overly familiar with nor does it seem particularly obvious when reading line 1 at the top of the page. Do you expect all the users of your library to already be familiar with Go?

halostatue:
> Is it obvious to people though that the _way_ to pass that second parameter in is with a space separated second quoted string?

Mostly yes, because it's in a reference to a templating function. This specific function only has the single example, but in the same section, there are other examples that show how multiple parameters are used, as with [onepassword](https://www.chezmoi.io/reference/templates/1password-functions/onepassword/).

> I saw that `[account]` was the second optional parameter, but getting the formatting correct to test it at the command line with `execute-template` took me nearly a dozen attempts, and ultimately sending the source to ChatGPT to ask it how it should be formatted. The lack of argument separators is not something that I'm overly familiar with nor does it seem particularly obvious when reading line 1 at the top of the page. Do you expect all the users of your library to already be familiar with Go?

With Go, no. But the second paragraph of [templating](https://www.chezmoi.io/user-guide/templating/#:~:text=chezmoi%20uses%20the%20text%2Ftemplate%20syntax%20from%20Go%20extended%20with%20text%20template%20functions%20from%20sprig.) points to Go's [text/template](https://pkg.go.dev/text/template) reference. text/template is similar to Liquid and Jinja templating languages in that it operates on pipelines and function parameters are positional with whitespace separation (and quotes to pass parameters with spaces in them).

I hope that this distinction helps understand the reasoning. This isn't to say that improvements can't be made, but it is *fair* to expect some exploration and review the examples provided, not just for the single command being looked up, but other related commands.

jondkinney:
@twpayne I think the suggestion of collapsing the examples like @halostatue suggested to match the style of the other entries would be very helpful. Every other doc page in the 1Password section has examples of how to use the template function with each of the optional parameters with the notable exceptions of the `onepasswordItemFields` and `onepasswordRead ` functions.

If that example were there for the `onepasswordRead` function, it would have helped me tremendously.

PR #4162, opened 2024-12-27
docs: reference templates in top-level passwd mgmt

jondkinney:
It's important for people to understand that the password manager feature is an extension/implementation of the existing templating capabilities of Chezmoi.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/password-managers/index.md
+++ b/assets/chezmoi.io/docs/user-guide/password-managers/index.md
@@ -1,6 +1,29 @@
-# Password manager integration
+# Password Manager Integration
 
-Template functions allow you to retrieve secrets from many popular password
-managers. Using a password manager allows you to keep all your secrets in one
-place, make your dotfiles repo public, and synchronize changes to secrets
-across multiple machines.
+Using a password manager with Chezmoi enables you to maintain a public
+dotfiles repository while keeping your secrets secure. Chezmoi extends its
+existing [templating capabilities](../templating.md) by providing password
+manager specific _template functions_ for many popular password managers.
+
+When Chezmoi applies a template with a secret referenced from a password
+manager, it will automatically fetch the secret value and insert it into the
+generated destination file.
+
+## Example: Template with Password Manager Integration
+
+Here's a practical example of a `.zshrc.tmpl` file that retrieves an OpenAI API
+key from 1Password while maintaining other standard shell configurations:
+
+```zsh
+# set up $PATH
+# …
+
+# OpenAI API Key retrieved from 1Password
+export OPENAI_API_KEY='{{ onepasswordRead "op://Personal/openai-api-key/password" }}'
+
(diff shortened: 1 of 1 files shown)


assets/chezmoi.io/docs/user-guide/password-managers/index.md
```diff
@@ -1,6 +1,29 @@
...
+
+# OpenAI API Key retrieved from 1Password
+export OPENAI_API_KEY='{{ onepasswordRead "op://Personal/openai-api-key/password" }}'
+
+# set up aliases and useful functions
+```
+
+In this example, the `OPENAI_API_KEY` is retrieved from a 1Password vault
+named `Personal`, specifically from an item called `openai-api-key` in the
+`password` field.
```
halostatue, line 29:
```suggestion
Here's a practical example of a `.zshrc.tmpl` file that retrieves an CloudFlare
API token from 1Password while maintaining other standard shell configurations:

```zsh
# set up $PATH
# …

# Cloudflare API Token retrieved from 1Password for use with flarectl
export CF_API_TOKEN='{{ onepasswordRead "op://Personal/cloudlfare-api-token/password" }}'

# set up aliases and useful functions
```

In this example, the `CF_API_TOKEN` is retrieved from a 1Password vault
named `Personal`, specifically from an item called `cloudflare-api-token` in the
`password` field.
```

This is a *recommended* change, but not a required change (unless @twpayne would prefer it). I understand that a lot of people are using OpenAI, Claude, or other LLM systems, but given the chezmoi [LLM-generated content policy](https://github.com/twpayne/chezmoi/discussions/4010), I believe that it would be better to *not* specify something that the creator and main developer doesn't believe to be a net positive.

(I don't think that LLMs produce useful responses, but if you know enough about what you’re trying to do, you can still learn something from LLM hallucinations whether they seem to work or not.)

halostatue (approved):

assets/chezmoi.io/docs/user-guide/password-managers/index.md
jondkinney: Happy to change the reference there @halostatue.

jondkinney:
@halostatue I'm not sure what you prefer for etiquette when accepting changes from a PR review/suggestion. Would you like to remain co-author for those updates? Fine with me if so, but I can also rebase it to a single commit including your suggestions while retaining the conventional commit style, if you prefer. LMK and I'll get it tweaked. Thanks!

halostatue:
> @halostatue I'm not sure what you prefer for etiquette when accepting changes from a PR review/suggestion. Would you like to remain co-author for those updates? Fine with me if so, but I can also rebase it to a single commit including your suggestions while retaining the conventional commit style, if you prefer. LMK and I'll get it tweaked. Thanks!

I don't need to be credited for this; just squash it into your original commit, because we prefer to have one commit per pull request.

twpayne:
Closing due to inactivity.

PR #4163, opened 2024-12-27
feat: Add chezmoi:template:format-indent template directive

twpayne:
Fixes #3873.
Implements @bradenhilton's suggestion in https://github.com/twpayne/chezmoi/pull/3875#issuecomment-2241866024.

This still needs some work, so this is draft PR.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/directives.md
+++ b/assets/chezmoi.io/docs/reference/templates/directives.md
@@ -37,6 +37,33 @@ inherited by templates called from the file.
     # [[ "true" ]]
     ```
 
+## Format indent
+
+By default, chezmoi's `toJson`, `toToml`, and `toYaml` template functions use
+the default indent of two spaces. The indent can be overidden with:
+
+    chezmoi:template:format-indent=$STRING
+
+to set the indent to be the literal `$STRING`, or
+
+    chezmoi:template:format-indent-width=$WIDTH
+
+to set the indent to be `$WIDTH` spaces.
+
+!!! example
+
+    ```
+    {{/* chezmoi:template:format-indent="\t" */}}
+    {{ dict "key" "value" | toJson }}
+    ```
+
+!!! example
+
+    ```
+    {{/* chezmoi:template:format-indent-width=4 */}}
+    {{ dict "key" "value" | toYaml }}
+    ```
+
 ## Line endings
 
 Many of the template functions available in chezmoi primarily use UNIX-style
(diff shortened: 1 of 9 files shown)


assets/chezmoi.io/docs/reference/templates/directives.md
```diff
@@ -37,6 +37,22 @@ inherited by templates called from the file.
...
     ```
 
+## Format indent
+
+By default, chezmoi's `toJson`, `toToml`, and `toYaml` template functions use
+the default indent of two spaces. The indent can be overidden with:
+
+    chezmoi:template:format-indent=$VALUE
+
+`$VALUE` can be an integer number of spaces or a string.
```
halostatue, line 47: I wonder if this might not be better as two directives, `format-indent-width` and `format-indent`. They would be mutually exclusive. `format-indent-width` would be the integer number of spaces to use per level; `format-indent` would a a string, but would have to match `^[ \t]+$` to be valid. That is, `format-indent=""` would not be valid, nor would `format-indent=".."`.

(One could argue that it should be `(^ +$|^\t+$)` to prevent mixing tabs and spaces.)

halostatue (approved):
The changes look good. I haven't tested them myself, but I think it's worth asking whether we want to (1) have two directives instead of one, and (2) be strict on string values to prevent footguns.

assets/chezmoi.io/docs/reference/templates/directives.md
twpayne: Thanks, done. I've not been super-strict with the parsing, but any errors should be quickly visible to the user.

halostatue (approved):
Looks good.

assets/chezmoi.io/docs/reference/templates/directives.md
bradenhilton: If we want to be strict about only allowing tabs or spaces, would something similar to how it's done in [EditorConfig](https://spec.editorconfig.org/#indentation-non-normative) be better? i.e.

```text
chezmoi:template:indent-style:spaces
chezmoi:template:indent-style:tabs
```

```text
chezmoi:template:indent-size:4

# and possibly also
chezmoi:template:tab-width:4
```

twpayne: I don't think we need to be super-strict on this, and that it should be sufficient to need only one directive to specify the indent. So, I'll leave this as-is.

PR #4149, opened 2024-12-20
feat: Implement builtin mode for keepassxc

bakito:
Implements: #4148

PR #4166, opened 2024-12-29
fix: apply lint rules for missing keepassxc attributes

bakito:
This fixes fixes the failing lint checks introduced by #4149


twpayne:
Thanks!

Issue #1800, opened 2022-01-02
Push RPM to Copr?


dngray:
## Is your feature request related to a problem? Please describe.

Redhat has a service called Copr https://copr.fedorainfracloud.org/coprs/ (they are like Ubuntu PPAs).

## Describe the solution you'd like

It would be nice if there was an official repository. The Neomutt project [has one](https://neomutt.org/distro/fedora) for example.

This would mean Redhat users would automatically have the latest Chezmoi instead of manually having to download the RPM or make their own repository.

twpayne:
This is a great idea, but I would prefer to add support for this to the build tool that chezmoi uses so that other projects can benefit. See https://github.com/goreleaser/goreleaser/issues/2804 for details.

dngray:
Another solution to this might be to create a flatpak, that way it could be installed on any distribution, that has flatpak.

It's not uncommon for commandline tools to have flatpaks, two examples being [NeoVim](https://flathub.org/apps/details/io.neovim.nvim) and [MPV](https://flathub.org/apps/details/io.mpv.Mpv).

- https://github.com/flathub/io.neovim.nvim
- https://github.com/flathub/io.mpv.Mpv

One of the things I like about flatpaks is they can be installed "per user". ie `flatpak install flathub --user io.chezmoi` for example.

twpayne:
> This would mean Redhat users would automatically have the latest Chezmoi instead of manually having to download the RPM or make their own repository.

Note that if you have installed chezmoi using an rpm package then the `chezmoi upgrade` command will download and install the latest version also using an rpm:

```
$ sudo rpm -i https://github.com/twpayne/chezmoi/releases/download/v2.9.4/chezmoi-2.9.4-x86_64.rpm
$ rpm -q chezmoi
chezmoi-2.9.4-1.x86_64
$ chezmoi upgrade
chezmoi version v2.9.5, commit af9290311e64e9f34b090c350be84070d1845863, built at 2022-01-05T22:11:53Z, built by goreleaser
$ rpm -q chezmoi
chezmoi-2.9.5-1.x86_64
```

> Another solution to this might be to create a flatpak, that way it could be installed on any distribution, that has flatpak.

I had a look at build at flatpak of chezmoi, but:
* The flatpak documentation states that flatpak is targeted at GUI applications, not CLIs.
* To run chezmoi with flatpak you'd have to use the verbose command `flatpak run io.chezmoi.Chezmoi`.
[...]

PR #4168, opened 2024-12-30
chore: Build flatpak package

twpayne:
Refs #1800.

This PR attempts to build a Flatpak package of chezmoi.

As far as I can tell, Flatpak in its current form is incompatible with chezmoi:

* Flatpak's sandbox is too restrictive. Even with full filesystem access, chezmoi cannot execute other processes, which breaks password manager integration, encryption, and several template functions.
* Flatpak mangles XDG directory specification directories like `$XDG_DATA_HOME`, making Flatpak-installed chezmoi incompatible with conventionally-installed chezmoi.

There are likely other problems.

This is a draft PR for discussion.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -269,6 +269,15 @@ jobs:
       with:
         name: chezmoi-windows-amd64
         path: dist/chezmoi-nocgo_windows_amd64_v1/chezmoi.exe
+    - name: create-flatpak
+      if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
+      run: |
+        sudo apt-get --no-install-suggests --no-install-recommends --quiet --yes install flatpak flatpak-builder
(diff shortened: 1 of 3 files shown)


twpayne:
The CI failure seems to be because the available version of `flatpak-builder` on Ubuntu 20.04 (used for release building because its old version of glibc means the compiled binaries run on a wide range of systems):

```
Flatpak 1.6.5
flatpak-builder 1.0.10
...

twpayne:
Following https://github.com/flatpak/flatpak/issues/5999#issuecomment-2565392674, it seems that chezmoi would need to bundle all applications it needs (including `git`, `gpg`, `age`, `nvim`, `emacs`, `code`, all password manager CLIs, etc.) which is neither desirable nor feasible, so I'll close this as "won't fix" for now.

PR #4189, opened 2025-01-07
fix: Correct nested groups for keepassxc builtin mode

bakito:
This PR fixes nested groups for keepassxc when run in builtin mode and adds additional tests to assure the functionality.

### DIFF
--- a/internal/cmd/keepassxctemplatefuncs_test.go
+++ b/internal/cmd/keepassxctemplatefuncs_test.go
@@ -74,6 +74,13 @@ func TestKeepassxcTemplateFuncs(t *testing.T) {
 	attachmentName := "test / attachment name"
 	attachmentData := "test / attachment data"
 
+	nestedGroupName := "some/nested/group"
+	nestedEntryName := nestedGroupName + "/nested entry"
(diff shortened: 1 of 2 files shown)


internal/cmd/keepassxctemplatefuncs_test.go
```diff
@@ -170,3 +182,47 @@ func TestKeepassxcTemplateFuncs(t *testing.T) {
 		})
 	}
 }
+
+func setupKeepassEntry(
+	t *testing.T,
+	command, tempDir, database, databasePassword, groupName, entryName, entryUsername, entryPassword, attachmentName, attachmentData string,
+) {
```
twpayne, line 189: nit: instead of passing lots of arguments with the same type, pass a struct with named fields. For an example, see https://github.com/twpayne/chezmoi/blob/94f00d11d79581d5bc755574861a22b7716d81b0/internal/chezmoi/sourcestate.go#L340-L366

bakito: done

twpayne:
Thank you!

PR #4193, opened 2025-01-09
chore: Update GitHub Actions

twpayne:


Fendyzainol9:
Thankyou for approval

.github/actions/free-disk-space/action.yml
```diff
@@ -0,0 +1,11 @@
+name: free-disk-space
```
KapJI, line 1: Why did you add this? Github runners have [14 GB of disk space](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) and cache entry for release jobs is about 3 GB.

twpayne: Builds were running out of space, e.g.https://github.com/twpayne/chezmoi/actions/runs/12700388677.

PR #4227, opened 2025-01-25
feat: Add --every flag to update command

twpayne:
The goal behind this PR is address the issue raised in [this blog post](https://montasaurus.com/posts/how-i-build):

> I use [chezmoi](https://www.chezmoi.io/) to manage my dotfiles. I don't love the workflow I use with it, especially once I started using templates. It's a bit manual to keep updated, but I like having my config in git. I use brew bundle dump --global --force to include my brew installs, which also picks up my VSCode extensions.

Specifically, the "It's a bit manual to keep updated" bit.

With this PR, the user can add

    chezmoi update --every=24h

to their login scripts and `chezmoi update` will run at most once every 24 hours. The same issue was raised by @oryjkov some time ago.

Basically, this allows users to keep their dotfiles up to date across multiple machines, without any manual intervention, and without chezmoi having to run any kind of daemon.

This is a draft PR, but all feedback, particularly on the user experience aspects, is very welcome.


halostatue:
I don't love it, but it seems to be OK.

I rarely use `chezmoi update` (mostly because I _do_ edit target file outside of chezmoi and need to `chezmoi merge-all` frequently).

I can see some problematic sides to this:

1. Every macOS shell should be a login shell, see [macOS Login Shells](https://ghostty.org/docs/help/macos-login-shells) at ghostty.org. That means that for zsh/bash users the login script would run this on every new tab or shell window (fish users would usually have this run anyway, because I at least never use `status is-login`, only `status is-interactive` for tests).
2. If I launch multiple shell windows or tabs quickly, at least one of them is going to fail with a chezmoi deadlock — and the other will trigger a 1Password access prompt (others may get password prompt, etc).
3. I have a task that requires sudo that runs on changes to certain files. The sudo prompt would come out of nowhere with this.

It's not a bad idea, but I don't know that it will solve the issue raised. Maybe it's worth checking in with ~montasaurus (`@` deliberately replaced to not mention right now) to get an understanding of what was meant?

---

What if instead of `chezmoi update --every=24h`, we had a `chezmoi check-update --every=24` where check-update uses the built-in git support (to see if the remote+branch has changed in the last "period") and [gen2brain/beeep](https://github.com/gen2brain/beeep) (or something like it)  to send a desktop notification (or some other visible message). Since it would be updating state, it would still need to lock the state during the check, but the state-check-is-locked code could be relaxed on `check-update` to _hide_ those error messages.

KapJI:
It will be probably more useful just to have some command to check "if update ran within last X hours/minutes". It's probably already doable using templates. Maybe just exposing `last_update` timestamp to chezmoi data is enough.

Using this users may configure behaviour they want. E.g. run update automatically, show desktop notificaiton using their favourite notification engine, echo message to terminal, or something else.

Issue #4235, opened 2025-01-27
Add option to show git diff before commit


qianlongzt:
## Is your feature request related to a problem? Please describe.

I use `git.autoPush` to avoid forgetting to push configuration changes.
I want to know what changes were made before writing the commit message.

```shell
vim ~/.config/xxx
chezmoi add ~/.config/xxx
```

![Image](https://github.com/user-attachments/assets/33a42ee5-98c2-4b8d-b664-5afc65407f06)

## Describe the solution you'd like

add new option `git.showDiff`(or else) to show `git diff --cached` before commit

## Describe alternatives you've considered

1. `chezmoi diff` shows changes in the opposite direction.
2. `chezmoi cd` requires temporarily disabling `git.autoCommit` and `git.autoPush`.
3. `commitMessageTemplate = "{{output \"git\" \"diff\" \"--color=always\" \"--cached\" | promptString }}"`
   1. Sometimes I get the error on windows `git no option --cached`
   2. `chezmoi: template: git.commitMessageTemplate:1:2: executing "git.commitMessageTemplate" at <output "git" "diff" "--color=always" "--cached">: error calling output: 'D:\\scoop\\local\\shims\\git.exe' diff --color=always --cached: exit status 129`

![Image](https://github.com/user-attachments/assets/05e78370-965b-472d-99f3-98a5fa437e20)

twpayne:
#4326 adds `git-auto-commit` and `git-auto-push` [hooks](https://www.chezmoi.io/reference/configuration-file/hooks/). With this, you can add a `hooks.git-auto-commit.pre` hook that runs `git diff -C $CHEZMOI_SOURCE_DIR` before invoking your commit message template to prompt for the commit message. Would you be able to test this?

qianlongzt:
Thanks for your work.
I tested it, but it runs the command in the home directory. Can we add an option to specify the working directory for the hook?

```toml
# ~/.config/chezmoi/chezmoi.toml.tmpl
[hooks.git-auto-commit.pre]
command = "git"
args = ["diff", "--cached", "--color=always"]

directory = {{.chezmoi.workingTree}}

[git]
autoCommit=true
autoPush=true
```
https://github.com/twpayne/chezmoi/blob/500c677cc7b8132a54c7e48d816355487cf6cfbc/internal/cmd/config.go#L2560

qianlongzt:
Thanks for reply again, I miss `-C` option first reply.

[...]

PR #4257, opened 2025-02-05
fix(git): run git auto commit hook before ask commit message

qianlongzt:
https://github.com/twpayne/chezmoi/issues/4235

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -1579,11 +1579,11 @@ func (c *Config) gitAutoCommit(cmd *cobra.Command, status *chezmoigit.Status) er
 	if status.Empty() {
 		return nil
 	}
-	commitMessage, err := c.gitCommitMessage(cmd, status)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


qianlongzt:
@twpayne I have no idea why the commit message check failed. Should I fix the commit message or the commit message check?

According to <https://www.conventionalcommits.org/en/v1.0.0/> , words starting with lowercase in the description are accepted, and the scope should be accepted as anything other than `feat`.

The `test-macos-2` failure doesn’t seem related to this change, does it?

twpayne:
> Should I fix the commit message or the commit message check?

Please fix the commit message. In this case, the commit message should be:

    fix: Run git auto commit hook before creating commit message
    


> The `test-macos-2` failure doesn’t seem related to this change, does it?

Correct. Unfortunately macOS tests on GitHub's infrastructure are flaky and occasionally time out. I have no idea why, and suspect that the problem is with GitHub not chezmoi.

qianlongzt:
I updated the commit message.

twpayne:
Thanks!

PR #4259, opened 2025-02-07
docs: (WIP) Add version references

halostatue:
This is a draft of something I've been considering for a while and just took the time to pull part of it together. I'd like some feedback, so please check this out and view the generated docs.

- **chore: Update documentation dependencies**
- **docs: Enable site-urls plugin**
- **docs: Update developer-guide linking**
- **docs: Update release history anchor identifiers**
- **docs: Update root level docs with site**
- **docs: WIP add version tags**

The version tags are using https://github.com/glennmatthews/markdown-version-annotations instead of https://mkdocs-badges.six-two.dev/badges/

`attr_list` are needed to make the changelog links easier to address.

https://github.com/OctoPrint/mkdocs-site-urls has been added for easier cross referencing.


assets/chezmoi.io/docs/developer-guide/architecture.md
```diff
@@ -10,28 +10,27 @@ example:
...
+| Directory                        | Contents                                                                                                                                            |
+| -------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `assets/chezmoi.io/docs/`        | The documentation single source of truth. Help text, examples, and the [chezmoi.io][website] website are generated from the files in this directory |
+| `internal/chezmoi/`              | chezmoi's core functionality                                                                                                                        |
+| `internal/cmd/`                  | Code for the `chezmoi` command                                                                                                                      |
+| `internal/cmd/testdata/scripts/` | High-level tests of chezmoi's commands using [`testscript`][testscript]                                                                             |
```
twpayne, line 24: The testscript link doesn't render correctly.

twpayne (commented):
The more readable links in the Markdown documentation is a nice improvement.

I'm not sure about the version information for several reasons:

1. It's a lot of work to add correct information, and I don't recall any issues that were due to people using an older version of chezmoi, so I don't think it solves a problem we have.

2. Adding version information to the documentation (e.g. the note on the addition of rage as an alternative age backend) effectively duplicates the release notes. This makes the documentation more cluttered, and less easy to read IMHO: instead of reading about the software behaves now, the user instead has to read the life story of the software.

So, I'd rather not include the version information and feature history, but I'm open to being persuaded otherwise.

assets/chezmoi.io/docs/developer-guide/architecture.md
halostatue: I’ll look at that later. It should work.

halostatue:
> The more readable links in the Markdown documentation is a nice improvement.
> 
> I'm not sure about the version information for several reasons:
> 
> 1. It's a lot of work to add correct information, and I don't recall any issues that were due to people using an older version of chezmoi, so I don't think it solves a problem we have.

It's a lot of work _initially_, yes. But moving forward, I don't think it would be that much. I have had this at the back of my head for the last year or so because I don't always remember to do `chezmoi upgrade` (it will be less of a problem on my work computer because I’m using MacPorts for that and it updates regularly). I find it useful to know whether a feature (mostly the presence of a particular template function) I want will be usable on the machine I’m on (which may not be able to upgrade immediately). That would also help me set the `.chezmoiversion` in my repo (for example, when I start using `outputList`, I know that I have to set that to `2.59.1`).

> 2. Adding version information to the documentation (e.g. the note on the addition of rage as an alternative age backend) effectively duplicates the release notes. This makes the documentation more cluttered, and less easy to read IMHO: instead of reading about the software behaves now, the user instead has to read the life story of the software.

So, that part (details) is completely experimentation done to show that these are _just_ annotations. I think that the most that would be wanted is something like this at the bottom of a command or function reference — because this triggers a link to the release notes section for that version and doesn't need anything else.

```markdown
+++ 2.51.0
```

A little harder, but I think _better_ would be if as part of the "prepare release" step, we could do a search and replace for something like:

```markdown
+++ NEXT
```

and replace it with

```markdown
+++ $TAG
```

that gets pushed before the tag is created. At this point, new commands and functions only need to have `+++ NEXT` added at the bottom of the doc file and everything else is automatic.

I’m still experimenting, but I do think that it's useful information to have — but only as the "added in …" format (`+++ version`). We won't be using `--- version` and I don't really see where the update `+/- version` will meaningfully be used.

halostatue:
Closing this for now and spending time reviewing _all_ the docs for improved `site:…` references. It's making some references longer, but I think that having this explicit makes it easier to move around in the future.

I’m also moving *most* links to link *references* (`[link][ref]` as opposed to `[link](url)`), but there's a few pages where I am not doing that (command overview; a couple of others). The reason is that, for the most part, the link *reference* results in a more readable _text_ paragraph when one is reading the raw Markdown. I’ve been able to find a few places that could be cleared up (removing a word, adding a new word) because I can more easily read the whole paragraph instead, so I think it's a big win.

It'll be a couple days more before I’m done, I think.

twpayne:
Cool, thanks for working on this.

PR #4283, opened 2025-02-14
chore: Pre-parse documentation for help, examples, and flags

twpayne:
This reduces start-up time by about 5ms on my machine by avoiding parsing each command's reference documentation.


twpayne:
I think I've hit a bug in a dependency, so this is draft for now.

Issue #4274, opened 2025-02-13
special chezmoiexternals folder recursively misshandle


KevinNitroG:
## Describe the bug

I split my externals into multiple files and folders.

```
 home/.chezmoiexternals
├──  universal
│   ├──  alacritty.toml.tmpl
│   ├──  bat.toml.tmpl
│   ├──  delta.toml.tmpl
│   ├──  githelpers.toml.tmpl
│   ├──  glow.toml.tmpl
│   ├──  nvim.toml.tmpl
│   ├──  obsidian.toml.tmpl
│   └──  yazi.toml.tmpl
├──  unix
│   ├──  kitty.toml.tmpl
│   ├──  oh-my-zsh.toml.tmpl
│   ├──  tmux-tpm.toml.tmpl
│   ├──  universal.toml.tmpl
│   └──  zathura.toml.tmpl
└──  windows
    ├──  autohotkey.toml
    ├──  evkey.toml.tmpl
    ├──  komorebi.toml
    ├──  others.toml.tmpl
    └──  unikey.toml
```

But it handle those externals into  `~/.chezmoiexternals/` folder.

```
 lt .chezmoiexternals --all --only-dirs --level 2
 .chezmoiexternals
├──  .config
│   ├──  alacritty
│   ├──  autohotkey
│   ├──  bat
│   ├──  cursor
│   ├──  delta
│   ├──  ghostty
│   ├──  git
│   ├──  glow
│   ├──  kitty
│   ├──  komorebi
│   ├──  nvim
│   ├──  sddm
│   ├──  wakatime
│   ├──  yazi
│   └──  zathura
├──  .oh-my-zsh
│   ├──  .devcontainer
│   ├──  .github
│   ├──  custom
│   ├──  lib
│   ├──  log
│   ├──  plugins
│   ├──  templates
│   ├──  themes
│   └──  tools
├──  .tmux
│   └──  plugins
├──  AppData
│   └──  Local
├──  Document
│   └──  PowerShell
└──  obsidian
    └──  main
```

## Expected behavior

Handle those file relative to `~/` folder, not `~/.externals/`

## Doctor

<details>
```
RESULT    CHECK                       MESSAGE
warning   version                     v2.59.1, built at 2025-02-07T22:28:21Z
ok        latest-version              v2.59.1
ok        os-arch                     linux/amd64 (Arch Linux)
[...]

PR #4285, opened 2025-02-15
docs: Update documentation linking

halostatue:
- **chore: Update documentation dependencies** 

  This enables a feature used in upcoming changes, `attr_list`.

  When a new release is made, we should be updating the `project.version` in `pyproject.toml`.

- **docs: Update documentation linking**

  This is a _very large change_, but there are essentially three changes applied over and over and over again.

  - Switch from inline links to link references (`[name](uri)` to `[name][ref]` with the `[ref]` resolved elsewhere in the document).

    This allows more of each content line to be compactly readable in the source markdown. Compare:

    ```diff
    -integration tests use the
    -[`github.com/alecthomas/assert/v2`](https://pkg.go.dev/github.com/alecthomas/assert)
    -framework. End-to-end tests use
    -[`github.com/rogpeppe/go-internal/testscript`](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript)
    +integration tests use the [`github.com/alecthomas/assert/v2`][assert] framework.
    +End-to-end tests use [`github.com/rogpeppe/go-internal/testscript`][testscript]
    ```

    The link ref definitions are conventionally placed at the bottom of the document.

  - Switch from relative to absolute paths for cross-referencing. Even when relative references are shorter, I think that we are better served by using site-relative paths.

    ```diff
    -[reference manual](../reference/concepts.md)
    +[reference manual](/reference/concepts.md)
    ```

    This does mean that references to files at the same level now have longer links, but those links are at the bottom with the references.

    ```diff
    -[`chezmoi forget`](forget.md)
    +[`chezmoi forget`][forget]
    +
    +[forget]: /reference/commands/forget.md
    ```

    It will be important to ensure that all documents are referenced this way, but there is nothing that Mkdocs can do to prevent relative references.

  - Update release history anchor identifiers. The H2 headings for release versions was updated to include an attribute list setting the `id` attribute to the release version name/tag:

    ```
    `{ id="{{ $release.Name }}" }`
    ```

    This means that the change log link for the latest version changes from `reference/release-history/#2590-2025-01-28` to `reference/release-history/#v2.59.0`. This will be _very_important for the version tagging because it keeps the anchors for releases to be the same as the tag instead of a munged tag and timestamp slug.


halostatue:
I’m sorry for how large this is, but I wanted to make sure that everything was updated. I am going to go through more review tomorrow, but both mkdocs and lychee (a link checker) suggest that the links are good, and I have found and fixed all of the cases where I ended up typing `[name][ref]]`.

I _strongly_ recommend looking at the rich preview[^1] of the changes for at least a few of the files. The main visual changes will be that some of the paragraphs are reformatted because the links are smaller which allows them to be more compact in the source. There are also a few places where I have reworded something because, when the link was made shorter and the whole sentence could be read…it didn't read as well as it might have otherwise.

I also found a few erroneous links or link text when I was cleaning this up, so there's that.

[^1]: CleanShot 2025-02-15 at 00 14 13@2x

bradenhilton:
Is it worth also adding the `Private :: Do Not Upload` classifier to prevent uploading (accidental or otherwise) to PyPA?

halostatue:
I will be squashing `fixup` commits after review; I am using them so that viewing changes in this VERY LARGE PR is easier to see.

halostatue:
I’ve added another commit that adds a new snippet for including a info admonition block for documentation that refers to `$FORMAT` so that there's more consistency. I think that the info block can be expanded, and we might want two (for those that can be templates vs those that cannot), but this is a start.

As noted in the commit message, I also added a *lot* of details to the `.chezmoidata/` and `.chezmoidata.$FORMAT` pages because I was surprised at the data handling approach and more surprised that it had not been documented — or raised as an issue — so far.

halostatue:
I have one more thing that I want to do with all of this, and I’m not sure if I should do it as part of this or separately (probably separately), but I think that the `special-files` and `special-directories` sections should be merged and documented together. Part of this is because of #4274, but also because `.chezmoidata.$FORMAT` and `.chezmoidata/` are essentially the *same* documentation, as would `.chezmoiexternal.$FORMAT` and `.chezmoiexternals/`.

twpayne (approved):
Thank you, this looks excellent.

twpayne:
> I have one more thing that I want to do with all of this, and I’m not sure if I should do it as part of this or separately (probably separately), but I think that the `special-files` and `special-directories` sections should be merged and documented together

Let's do this in a separate PR. I actually quite like this split, as there are some subtle differences in places, and the split was introduced a few months ago in #4016.

PR #4289, opened 2025-02-16
chore: Upgrade Mermaid version

halostatue:
Force `mermaid2` to the latest version (11.4.1).

This resulted in a minor bug on rendering where the palette numbering was apparently changed from 1 (light) and 2 (dark) to 0 (light) and 1 (dark). Updated the `refresh_on_toggle_dark_light.js` script to work with either scenario. `document.getElementById("__palette_0")` will return `null` if using 1 indexing, so we swap the dark palette switcher with the light palette switcher and set the dark palette switcher to `__palette_2`.

### DIFF
--- a/assets/chezmoi.io/docs/extra/refresh_on_toggle_dark_light.js
+++ b/assets/chezmoi.io/docs/extra/refresh_on_toggle_dark_light.js
@@ -1,10 +1,14 @@
-var paletteSwitcher1 = document.getElementById("__palette_1");
-var paletteSwitcher2 = document.getElementById("__palette_2");
+// The light palette is always the first palette. In some cases, it is numbered
+// as __palette_0 and in others it is numbered as __palette_1. The dark palette
+// is numbered as __palette_1 or __palette_2, depending on the index used for
+// the light palette.
+var paletteSwitcherLight = document.getElementById("__palette_0");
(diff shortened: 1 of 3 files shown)


bradenhilton (approved):
Would you mind adding a commit to fix the typos? (`.chezmo.$FORMAT.tmpl`)

halostatue:
> Would you mind adding a commit to fix the typos? (`.chezmo.$FORMAT.tmpl`)

I did that directly on `master` because it was breaking the build. Something still appears to be breaking it and I’m investigating.

twpayne:
In cf43c04d045e949351156326c18fc313b02d3f68, the commit message should begin with `docs:`, not `doc:`, which is causing the commit message lint check to fail.

bradenhilton:
@twpayne Should this be fixed with a force push, or can it be ignored moving forward?

twpayne:
Please fix it with a force push, as the release notes are generated from the commit messages.

halostatue:
I’ll do that. Sorry.

halostatue:
Fixed. I squashed it with the other typo commit that I had made; since we didn't need two stupid fixes by me in a row.

twpayne:
Thank you for the swift fix!

twpayne (approved):
Thank you :)

PR #4296, opened 2025-02-22
fix: Stop overriding umask with test code

### LINKED ISSUE #4214: git-repo external and umask don't play nice
## Describe the bug

With `umask = 0o077`, a `git-repo` external refresh always believes that the destination copy has changed. 

## To reproduce

I have the following external defined:

```toml
[".vim/pack/chezmoi/start/chezmoi.vim"]
  type = "git-repo"
  url = "https://github.com/alker0/chezmoi.vim.git"
  refreshPeriod = "168h"
  clone.args = ["--depth", "1"]
  pull.args = ["--ff-only"]
```

With the following in the config:

```toml
umask = 0o077
```

`cm apply -R .vim/pack/chezmoi/start/chezmoi.vim` will always say: `.vim/pack/chezmoi/start/chezmoi.vim has changed since chezmoi last wrote it?`:

```
~ ❯ chezmoi apply -R .vim/pack/chezmoi/start/chezmoi.vim 
.vim/pack/chezmoi/start/chezmoi.vim has changed since chezmoi last wrote it? overwrite/all-overwrite/skip/quit
```

Overwriting and running it again will produce the same outcome again.

Removing the `umask` setting fixes the problem.

## Expected behavior

```
~ ❯ chezmoi apply -R .vim/pack/chezmoi/start/chezmoi.vim 
Already up to date.
```

Same.

hugoh:
Fixes #4214.

`init()` in `chezmoitest_unix.go` was being executed even for the released version.

Now, doctor gives the correct umask:

```
❯ umask 077
❯ go run . doctor | grep umask
warning   umask                       077
```

instead of [always returning 022](https://github.com/twpayne/chezmoi/issues/4214#issuecomment-2649520537).

I ended up using a build tag to include it only on test builds. There likely is a cleaner way to do this.

### DIFF
--- a/internal/chezmoitest/chezmoitest.go
+++ b/internal/chezmoitest/chezmoitest.go
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"regexp"
 	"runtime"
-	"strconv"
 	"strings"
 	"testing"
 
@@ -97,13 +96,4 @@ func WithTestFS(t *testing.T, root any, f func(vfs.FS)) {
 	assert.NoError(t, err)
 	t.Cleanup(cleanup)
 	f(fileSystem)
-}
-
-// mustParseFileMode parses s as a fs.FileMode and panics on any error.
-func mustParseFileMode(s string) fs.FileMode {
-	u, err := strconv.ParseUint(s, 0, 32)
-	if err != nil {
-		panic(err)
-	}
-	return fs.FileMode(uint32(u))
 }
--- a/internal/chezmoitest/chezmoitest_utils.go
+++ b/internal/chezmoitest/chezmoitest_utils.go
@@ -1,0 +1,17 @@
+//go:build test
+
+package chezmoitest
+
+import (
+	"io/fs"
+	"strconv"
+)
+
+// mustParseFileMode parses s as a fs.FileMode and panics on any error.
+func mustParseFileMode(s string) fs.FileMode {
+	u, err := strconv.ParseUint(s, 0, 32)
+	if err != nil {
+		panic(err)
+	}
+	return fs.FileMode(uint32(u))
+}


internal/chezmoitest/chezmoitest_init_unix.go
halostatue, line 1: I am not sure that adding `-tags=test` is the right choice here[^1], but it is the minimal file change option. However, there are more files that need changing than you may have realized, and I disagree with your choice of popping the `init` function into its own file.

Below, you’ll see a `.patch` file that I came up with in looking at this and addressing this a little more directly. It includes:

- `.github/workflows/main.yml`, `Makefile`: These should be the same changes as yours.
- `assets/chezmoi.io/docs/developer-guide/architecture.md`,`assets/chezmoi.io/docs/developer-guide/index.md`
- `assets/vagrant/freebsd14.test-chezmoi.sh`, `assets/vagrant/openbsd7.test-chezmoi.sh`

I approached the `init()` problem differently than you did in `internal/chezmoitest/chezmoitest.go`, `internal/chezmoitest/chezmoitest_unix.go`, and `internal/chezmoitest/chezmoitest_windows.go`.

Fundamentally, I think that this is a go compiler error. `chezmoitest` is its own package and is never imported into `chezmoi` itself except in `_test` files (which are not included in non test builds). As such, any init blocks should not be linked into the resulting binary but they clearly are. I’m not going to chase this, but I think it would be worthwhile coming up with a minimal reproduction and reporting it to the Go team.

---

`changes.patch`:

```diff
diff --git c/.github/workflows/main.yml i/.github/workflows/main.yml
index 2319526f43bc..52ff8c44f066 100644
--- c/.github/workflows/main.yml
+++ i/.github/workflows/main.yml
@@ -164,18 +164,18 @@ jobs:
     - name: test
       env:
         CHEZMOI_GITHUB_TOKEN: ${{ secrets.CHEZMOI_GITHUB_TOKEN }}
       run: |
         if [ "${{ matrix.test-index }}" = "0" ]; then
-          go test ./... -short -race
-          go test ./internal/cmd -run=TestScript -filter='^[0-9a-dA-D]' -race
+          go test -tags=test ./... -short -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[0-9a-dA-D]' -race
         elif [ "${{ matrix.test-index }}" = "1" ]; then
-          go test ./internal/cmd -run=TestScript -filter='^[e-hE-H]' -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[e-hE-H]' -race
         elif [ "${{ matrix.test-index }}" = "2" ]; then
-          go test ./internal/cmd -run=TestScript -filter='^[i-lI-L]' -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[i-lI-L]' -race
         else
-          go test ./internal/cmd -run=TestScript -filter='^[m-zM-Z]' -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[m-zM-Z]' -race
         fi
   test-oldstable-go:
     needs: changes
     if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
     runs-on: ubuntu-22.04
@@ -206,11 +206,11 @@ jobs:
         sudo install -m 755 rage/rage-keygen /usr/local/bin
     - name: test
       env:
         CHEZMOI_GITHUB_TOKEN: ${{ secrets.CHEZMOI_GITHUB_TOKEN }}
       run: |
-        go test ./...
+        go test -tags=test ./...
   test-release:
     needs: changes
     runs-on: ubuntu-22.04 # use older Ubuntu for older glibc
     permissions:
       contents: read
@@ -321,14 +321,14 @@ jobs:
       env:
         CHEZMOI_GITHUB_TOKEN: ${{ secrets.CHEZMOI_GITHUB_TOKEN }}
         TEST_FLAGS: '-ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o${{ matrix.umask }}" -race -timeout=1h'
       run: |
         if [ "${{ matrix.test-index }}" = "0" ]; then
-          go test ./... -short ${{ env.TEST_FLAGS }}
-          go test ./internal/cmd -run=TestScript -filter='^[0-9a-hA-h]' ${{ env.TEST_FLAGS }}
+          go test -tags=test ./... -short ${{ env.TEST_FLAGS }}
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[0-9a-hA-h]' ${{ env.TEST_FLAGS }}
         else
-          go test ./internal/cmd -run=TestScript -filter='^[i-zI-Z]' ${{ env.TEST_FLAGS }}
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[i-zI-Z]' ${{ env.TEST_FLAGS }}
         fi
   test-website:
     runs-on: ubuntu-22.04
     permissions:
       contents: read
@@ -375,18 +375,18 @@ jobs:
     - name: test
       env:
         CHEZMOI_GITHUB_TOKEN: ${{ secrets.CHEZMOI_GITHUB_TOKEN }}
       run: |
         if (${{ matrix.test-index }} -eq 0) {
-          go test ./... -short -race
-          go test ./internal/cmd -run=TestScript -filter='^[0-9a-cA-C]' -race
+          go test -tags=test ./... -short -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[0-9a-cA-C]' -race
         } elseif (${{ matrix.test-index }} -eq 1) {
-          go test ./internal/cmd -run=TestScript -filter='^[d-fD-F]' -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[d-fD-F]' -race
         } elseif (${{ matrix.test-index }} -eq 2) {
-          go test ./internal/cmd -run=TestScript -filter='^[g-lG-L]' -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[g-lG-L]' -race
         } else {
-          go test ./internal/cmd -run=TestScript -filter='^[m-zM-Z]' -race
+          go test -tags=test ./internal/cmd -run=TestScript -filter='^[m-zM-Z]' -race
         }
   check:
     runs-on: ubuntu-22.04
     permissions:
       contents: read
diff --git c/Makefile i/Makefile
index 421bd9d5c3a9..bf3c4d696e0d 100644
--- c/Makefile
+++ i/Makefile
@@ -89,12 +89,12 @@ test-all: test test-release rm-dist test-docker test-vagrant
 rm-dist:
 	rm -rf dist
 
 .PHONY: test
 test:
-	${GO} test -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o022" ./...
-	${GO} test -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o002" ./...
+	${GO} test -tags=test -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o022" ./...
+	${GO} test -tags=test -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=0o002" ./...
 
 .PHONY: test-docker
 test-docker:
 	( cd assets/docker && ./test.sh alpine archlinux fedora )
 
diff --git c/assets/chezmoi.io/docs/developer-guide/architecture.md i/assets/chezmoi.io/docs/developer-guide/architecture.md
index cf413be8220e..d99f541041c8 100644
--- c/assets/chezmoi.io/docs/developer-guide/architecture.md
+++ i/assets/chezmoi.io/docs/developer-guide/architecture.md
@@ -156,20 +156,20 @@ in the `entryState` bucket along with its contents SHA256 sum. On future
 invocations the script is only run if its contents SHA256 sum has changed, and
 its contents SHA256 sum is then updated in the persistent state.
 
 ## Testing
 
-chezmoi has a mix of, unit, integration, and end-to-end tests. Unit and
+chezmoi has a mix of unit, integration, and end-to-end tests. Unit and
 integration tests use the [`github.com/alecthomas/assert/v2`][assert] framework.
 End-to-end tests use [`github.com/rogpeppe/go-internal/testscript`][testscript]
 with the test scripts themselves in
 `internal/cmd/testdata/scripts/$TEST_NAME.txtar`.
 
 You can run individual end-to-end tests with
 
 ```sh
-go test ./internal/cmd -run=TestScript/$TEST_NAME
+go test -tags=test ./internal/cmd -run=TestScript/$TEST_NAME
 ```
 
 Tests should, if at all possible, run unmodified on all operating systems tested
 in CI (Linux, macOS, Windows, and FreeBSD). Windows will sometimes need special
 handling due to its path separator and lack of POSIX-style file permissions.
diff --git c/assets/chezmoi.io/docs/developer-guide/index.md i/assets/chezmoi.io/docs/developer-guide/index.md
index 44dfa21ee329..b15660170f8e 100644
--- c/assets/chezmoi.io/docs/developer-guide/index.md
+++ i/assets/chezmoi.io/docs/developer-guide/index.md
@@ -8,12 +8,12 @@
     correctness before sharing it. If you share un-reviewed LLM-generated
     content then you will be immediately banned. See [`CODE_OF_CONDUCT.md`][coc]
     for more information.
 
 chezmoi is written in [Go][go] and development happens on [GitHub][github].
-chezmoi is a standard Go project, using standard Go tooling. chezmoi requires
-Go 1.24 or later.
+chezmoi is a standard Go project, using standard Go tooling. chezmoi requires Go
+1.24 or later.
 
 Checkout chezmoi:
 
 ```sh
 git clone https://github.com/twpayne/chezmoi.git
@@ -27,11 +27,11 @@ go build
 ```
 
 Run all tests:
 
 ```sh
-go test ./...
+go test -tags=test ./...
 ```
 
 chezmoi's tests include integration tests with other software. If the other
 software is not found in `$PATH` the tests will be skipped. Running the full set
 of tests requires `age`, `base64`, `bash`, `bzip2`, `git`, `gpg`, `gzip`,
@@ -71,11 +71,11 @@ make test-release
     These can be avoided with by running tests with `SHELL=bash` or `SHELL=zsh`:
 
     ```sh
     SHELL=bash make test
     SHELL=zsh make smoke-test
-    SHELL=bash go test ./...
+    SHELL=bash go test -tags=test ./...
     ```
 
 [coc]: https://github.com/twpayne/chezmoi/blob/master/.github/CODE_OF_CONDUCT.md
 [go]: https://golang.org
 [github]: https://github.com
diff --git c/assets/vagrant/freebsd14.test-chezmoi.sh i/assets/vagrant/freebsd14.test-chezmoi.sh
index 47566250a8ed..b035047d6f4d 100755
--- c/assets/vagrant/freebsd14.test-chezmoi.sh
+++ i/assets/vagrant/freebsd14.test-chezmoi.sh
@@ -4,9 +4,9 @@ set -eufo pipefail
 
 git config --global --add safe.directory /chezmoi
 
 cd /chezmoi
 
-go test ./...
+go test -tags=test ./...
 
 sh assets/scripts/install.sh
 bin/chezmoi --version
diff --git c/assets/vagrant/openbsd7.test-chezmoi.sh i/assets/vagrant/openbsd7.test-chezmoi.sh
index 47566250a8ed..b035047d6f4d 100755
--- c/assets/vagrant/openbsd7.test-chezmoi.sh
+++ i/assets/vagrant/openbsd7.test-chezmoi.sh
@@ -4,9 +4,9 @@ set -eufo pipefail
 
 git config --global --add safe.directory /chezmoi
 
 cd /chezmoi
 
-go test ./...
+go test -tags=test ./...
 
 sh assets/scripts/install.sh
 bin/chezmoi --version
diff --git c/internal/chezmoitest/chezmoitest.go i/internal/chezmoitest/chezmoitest.go
index 09b1110b92af..68bf87f87c5f 100644
--- c/internal/chezmoitest/chezmoitest.go
+++ i/internal/chezmoitest/chezmoitest.go
@@ -18,11 +18,15 @@
 	"github.com/twpayne/go-vfs/v5/vfst"
 
 	"github.com/twpayne/chezmoi/v2/internal/chezmoilog"
 )
 
-var ageRecipientRx = regexp.MustCompile(`(?m)^Public key: ([0-9a-z]+)\s*$`)
+var (
+	ageRecipientRx = regexp.MustCompile(`(?m)^Public key: ([0-9a-z]+)\s*$`)
+
+	Umask fs.FileMode
+)
 
 // AgeGenerateKey generates an identity in identityFile and returns the
 // recipient.
 func AgeGenerateKey(command, identityFile string) (string, error) {
 	cmd := exec.Command(command+"-keygen", "--output", identityFile)
diff --git c/internal/chezmoitest/chezmoitest_unix.go i/internal/chezmoitest/chezmoitest_unix.go
index f1cb8a756b81..beb08d9afc1a 100644
--- c/internal/chezmoitest/chezmoitest_unix.go
+++ i/internal/chezmoitest/chezmoitest_unix.go
@@ -1,29 +1,28 @@
-//go:build unix
+//go:build unix && test
 
 package chezmoitest
 
 import (
 	"golang.org/x/sys/unix"
 )
 
-var (
-	// umaskStr is the umask used in tests represented as a string so it can be
-	// set with the
-	//   -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=..."
-	// option to go build and go test.
-	umaskStr = "0o022"
+// umaskStr is the umask used in tests represented as a string so it can be
+// set with the
+//
+//	-ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=..."
+//
+// option to go build and go test.
+var umaskStr = "0o022"
 
+func init() {
 	// Umask is the umask used in tests.
 	//
 	// If you change this then you will need to update the testscripts in
 	// testdata/scripts where permissions after applying umask are hardcoded as
 	// strings. Pure Go tests should use this value to ensure that they pass,
 	// irrespective of what it is set to. Be aware that the process's umask is a
 	// process-level property and cannot be locally changed within individual
 	// tests.
 	Umask = mustParseFileMode(umaskStr)
-)
-
-func init() {
 	unix.Umask(int(Umask))
 }
diff --git c/internal/chezmoitest/chezmoitest_windows.go i/internal/chezmoitest/chezmoitest_windows.go
index 65c0b2f3de0d..3fa92469b5bb 100644
--- c/internal/chezmoitest/chezmoitest_windows.go
+++ i/internal/chezmoitest/chezmoitest_windows.go
@@ -1,14 +1,18 @@
+//go:build windows && test
+
 package chezmoitest
 
-var (
-	// umaskStr is the umask used in tests represented as a string so it can be
-	// set with the
-	//   -ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=..."
-	// option to go build and go test.
-	umaskStr = "0"
+// umaskStr is the umask used in tests represented as a string so it can be
+// set with the
+//
+//	-ldflags="-X github.com/twpayne/chezmoi/v2/internal/chezmoitest.umaskStr=..."
+//
+// option to go build and go test.
+var umaskStr = "0"
 
+func init() {
 	// Umask is the umask used in tests.
 	//
 	// On Windows, Umask is zero as Windows does not use POSIX-style permissions.
 	Umask = mustParseFileMode(umaskStr)
-)
+}
```

[^1]:  I would prefer `//go:build !release`, but that would mean adding `go build -tags=release` to all release builds and goreleaser both, which would mean modifying, well, every other file in the build. This could have been resolved with golang/go#21360, but purity over practicality, I guess. Equally annoying is the fact that you can't just use `//go:build test` on *all* files in `internal/chezmoitest` without breaking the entire build because the build tags exclude all files in that directory. Yes, that's on purpose, so just don't build that!

halostatue (changes requested):

hugoh:
@halostatue - I've also had to [fix a function unused linter error](https://github.com/twpayne/chezmoi/pull/4296/commits/b4dd9c7ea6c67827a195d7aebc7df840fab95c62). This could also be solved with a directive to golangci-lint. The init / build tag combination is tricky.

halostatue:
Makes sense. Like I said, I still think that this is a go compiler error — but I’m not enough of a Go developer to be able to build a minimal reproduction case, report it, and try to get them to believe it.

---

Please leave the commits as separate for now because I’d like Tom's input on this, but before this can be merged, it will need to be squashed to a single meaningful commit since all of the commits are internal fixes.

halostatue:
@hugoh Thank you for digging into this, but @twpayne found a simpler fix (#4298) because of the way that the go compiler handles filename-based filters (`_test` must be last such filter to ensure that it is only built with `go test`). This caused a silent inclusion of the `chezmoitest` package into the `chezmoi` binary, which resulted in the umask masking observed.

(FWIW, I still think that this is a compiler flaw, but I now know that it is one that will never be addressed and never even be acknowledged as a problem by the Go team, because they like the implicit filename-based nature of the compile filters. It's simple and works very well until it doesn't.)

twpayne:
@hugoh and @halostatue thank you both so much for all your work here!

I had a hunch that it wasn't a compiler bug (Go is sufficiently widely used that I'm pretty sure that someone else would have hit it by now) so I searched for any files that were importing `chezmoitest`. I wasn't sure about the order of OS constraints and `_test` in Go source file names, but as this problem is only observed on macOS, I suspected that `mackupcmd_test_darwin.go` might have been the cause.

@hugoh would you be able to test #4298?

twpayne (commented):
Hopefully #4298 is a simpler fix.

hugoh:
> @hugoh would you be able to test #4298?

@twpayne - this looks like it indeed fixes the problem I saw with umask:

```
~/C/chezmoi ( HEAD) (8654a14) [≡2] ❯ umask
022
~/C/chezmoi ( HEAD) (8654a14) [≡2] ❯ go run . doctor | grep umask
ok        umask                       022
~/C/chezmoi ( HEAD) (8654a14) [≡2] ❯ umask 077
~/C/chezmoi ( HEAD) (8654a14) [≡2] ❯ go run . doctor | grep umask
warning   umask                       077
```

And #4298 is a *much* simpler change. I'm going to close this PR. Thank you.

PR #4315, opened 2025-03-03
chore: Use sync.OnceValues to reduce redundant computation

qianlongzt:
example
```go
{{- if and (lookPath "mise") (eq .os "linux") }}
{{- output "mise" "completion" "fish" -}}
{{end -}}
```
```log
time=2025-03-03T16:39:37.974+08:00 level=INFO msg=ReadFile component=system name=/home/qianlongzt/.local/share/chezmoi/home/private_dot_config/fish/completions/empty_mise.fish.tmpl size=101 data="{{- if and (lookPath \"mise\") (eq .os \"linux\") }}\n{{- output \"mis..."
time=2025-03-03T16:39:38.042+08:00 level=INFO msg=Output cmd="/home/linuxbrew/.linuxbrew/bin/mise completion fish" duration=66.693957ms size=696 output="# if \"usage\" is not installed show an error\nif ! command -v usag..." err=<nil>
time=2025-03-03T16:39:38.095+08:00 level=INFO msg=Output cmd="/home/linuxbrew/.linuxbrew/bin/mise completion fish" duration=52.165978ms size=696 output="# if \"usage\" is not installed show an error\nif ! command -v usag..." err=<nil>
time=2025-03-03T16:39:38.096+08:00 level=INFO msg=Lstat component=system name=/home/qianlongzt/.config/fish/completions/mise.fish
```

PR #4316, opened 2025-03-03
chore: Add a test for repeated template evaluation

twpayne:
Refs #4315.

This PR does two things:

* Adds a test for what #4315 fixed.
* Tidies up a few comments and variable names that were also fixed in #4315.

Thank you to @qianlongzt for the original debugging and fix!

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -1813,8 +1813,8 @@ func (s *SourceState) newSourceStateDir(absPath AbsPath, sourceRelPath SourceRel
 }
 
 // newCreateTargetStateEntryFunc returns a targetStateEntryFunc that returns a
-// file with sourceLazyContents if the file does not already exist, or returns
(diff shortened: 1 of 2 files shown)


halostatue:
I have a small worry about the description you’ve put in the code, because I have a number of templates that I’ve built _recently_ where I execute *similar* `output` commands with different parameters (I'm not using the `outputList` function yet), and then I perform filtering on them.

So I might have something like `$files := includeTemplate "1p/documents" (dict "account" $account.url "startswith" "aws-credentials: ")` in one call and `$files := includeTemplate "1p/documents" (dict "account" $account.url "startswith" "aws-config: ")` in another, and those would be in loops that go through different 1Password accounts. I need to ensure that they return different values for different parameters or the structure I’ve spent building for a while will no longer work.

twpayne:
This should be OK, as the fix is to avoid evaluating the same file contents template multiple times.

halostatue:
I have run tests on my local system with this merged and things are just fine.

PR #4328, opened 2025-03-07
chore: Fix flaky test on macOS GitHub Actions runner

### LINKED ISSUE #4301: `test-macos-3` fails a lot
## Describe the bug

chezmoi's CI runs a bunch of tests. The `test-macos-3` test often (~50% of the time) fails with a timeout after 10 minutes, but passes when the test is re-run. For an example failure, see https://github.com/twpayne/chezmoi/actions/runs/13487705954/job/37681029733.

## To reproduce

I cannot reproduce this, but see "Additional Context" below.

## Expected behavior

chezmoi's CI should pass reliably.

n/a

## Output of `chezmoi doctor`

n/a. This has been happening for a while.

## Additional context

At this point, I have only a few observations:

* The problem did not occur a few months ago, and I don't believe that the problem is related to the CI infrastructure. It's always the `test-macos-3` job that fails with a timeout. If it were related to the CI infrastructure then I would expect all `test-macos-*` jobs to fail equally often.
* Each time, the job fails at a different point. There is no consistency.
* Looking at the logs, tests that should be really fast (i.e. take less than a second) instead take several seconds to run. Clearly, there's something wrong with this set of tests. If it were a simple "noisy neighbor" (i.e. overloaded CI) problem, then all jobs would fail equally often.
* I've made various efforts to tweak the macOS test sharding (by reducing the last first letter of the test), but it's consistently `test-macos-3` that fails.
* Running some kind of `git bisect` on the test shards would likely find the problem, but it would take someone with a lot of patience to run this.

cc @KapJI because you've helped so much in the past and might have insight here.

twpayne:
This should eventually fix #4301. As I can't duplicate the problem locally this will be a bit noisy as I push various commits to find the flaky test.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -131,7 +131,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test-index: [0, 1, 2, 3]
+        test-index: [0, 1, 2, 3, 4]
     needs: changes
     if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
     runs-on: macos-15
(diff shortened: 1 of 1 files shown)


twpayne:
Well, that went quicker than expected. Looks like `test-macos-3` was taking over 10 minutes most of the time, and splitting it into two solved the problem. I suspect that most of the time taken was due to setup time, as the split on S/T was in the middle and both halves take about the same amount of time.

PR #4347, opened 2025-03-16
feat: Set LESS and LV env vars the same way that git does

### LINKED ISSUE #4346: auto-fix less colors when diffing

The same issue as in https://github.com/twpayne/chezmoi/issues/1332 .
When less is set as the `$PAGER`, chezmoi will send ANSI control characters to it to display color, which will lead to broken display on most setups. 

Note that git does not have the same issue, when using git diff or git log, and color works just fine. Apparently they do this:
https://www.reddit.com/r/git/comments/1d1make/comment/l5ut3lr/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button
Image

Which seems to work as a fine default for most people. It would be nice if chezmoi did this, fixing an issue for a relatively common setup.

twpayne:
Fixes #4346.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -1426,6 +1426,15 @@ func (c *Config) getDiffPagerCmd() (*exec.Cmd, error) {
 	}
 	pagerCmd.Stdout = c.stdout
 	pagerCmd.Stderr = c.stderr
+	// If the LESS or LV environment variables are not set, set them to sensible
+	// defaults the same way the git does. See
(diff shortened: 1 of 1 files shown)


k4lizen:
btw i don't know if this is relevant in chezmoi but setenv is not thread safe when using cgo
https://github.com/golang/go/issues/63567

twpayne:
This doesn't use `os.Setenv`. It modifies the environment that the pager will run in, not chezmoi's own environment.

PR #4359, opened 2025-03-19
feat: completion improvements

scop:


halostatue:
Thanks for your contribution. I'll want to test this with fish, but is there any specific command you found lacking that prompted this update?

I don't really use bash or zsh, so I can't test those.

twpayne:
Thanks for this. The commit messages don't match chezmoi's format, but I'll merge this now and fix the commit messages with a force push to `master`.

scop:
Thanks. No specific command prompted an update, checking out state of completions is just something I tend to do when starting to poke around commands new to me.

PR #4368, opened 2025-03-21
feat: Remove 1Password SDK integration

twpayne:
As far as I can tell, including the current 1Password SDK pulls in a full WASM runtime and about 8MB of opaque WASM code.

You can test this by building chezmoi locally and analyzing the built binary with https://gsa.zxilly.dev/. Please compare the built binary with an earlier version of chezmoi.

cc @carey404 refs https://github.com/twpayne/chezmoi/issues/3852


halostatue (approved):
I see no evidence that anyone is using this, so I am in favour of this.

svenjacobs:
Hello @twpayne, does this in any way impact the 1P integration? Probably not because the CLI `op` is used and not the SDK, right?

twpayne:
You're correct. The existing 1Password integration remains unchanged. This removed only the `onepasswordSDKItemsGet` and `onepasswordSDKSecretsResolve` template functions.

halostatue:
It's also worth noting that the 1Password SDK integration was _extremely_ limited, partially because the SDK was in beta and partially because of limitations involving the SDK and the types of account / interactions that could be achieved with it.

PR #4407, opened 2025-04-15
fix: Set .chezmoi.sourceFile in execute-template command

twpayne:
Fixes #4406.

PR #4408, opened 2025-04-15
fix: Set .chezmoi.sourceFile in execute-template command

### LINKED ISSUE #4406: `chezmoi execute-template -f FILENAME` does not set `.chezmoi.sourceFile` to `FILENAME`
## Describe the bug

`chezmoi execute-template` can be used to simulate and debug chezmoi templates. The addition of the `--file` option in [v2.62.0](https://github.com/twpayne/chezmoi/releases/tag/v2.62.0) allows running `.tmpl` files directly from the source directory. Templates that utilize the `.chezmoi.sourceFile` variable will fail, though, because `chezmoi execute-template` sets `.chezmoi.sourceFile` to `argN`, where `N` is the number of the command line argument. A simple example would be a `symlink_*.tmpl` that uses `osDir` to find the parent directory:
```
{{ joinPath ( osDir .chezmoi.sourceFile ) "real_file.ext" }}
```

## To reproduce

1. Create a template at `~/.local/share/chezmoi/test_template.tmpl` with the contents:
    ```
    {{ .chezmoi.sourceFile }}
    ```
2. Run `chezmoi execute-template -f ~/.local/share/chezmoi/test_template.tmpl`

## Expected behavior

**Expected output**
```console
$ chezmoi execute-template -f ~/.local/share/chezmoi/test_template.tmpl
test_template.tmpl
```

**Actual output**
```console
$ chezmoi execute-template -f ~/.local/share/chezmoi/test_template.tmpl
arg1
```

overhacked:
If a template file's path is in the chezmoi source directory, when
running `chezmoi execute-template -f FILENAME`, set
`.chezmoi.sourceFile` to the same source directory-relative path that a
template executing via `chezmoi apply` would have.

If a template's path is *not* in the source directory, then set
`.chezmoi.sourceFile` to the filename exactly as it is given on the
command line.

Previously, all template files executed via `chezmoi execute-template`
would get `.chezmoi.sourceFile` set to the string `"argN"`, where N is
the argument position on the command line, which breaks testing any
templates that use `sourceFile`.

Fixes #4406

### DIFF
--- a/internal/cmd/testdata/scripts/executetemplate.txtar
+++ b/internal/cmd/testdata/scripts/executetemplate.txtar
@@ -114,6 +114,13 @@ stdout 'alphabeta'
 exec chezmoi execute-template --file $HOME${/}file.tmpl
 stdout '# contents of file.tmpl'
 
+exec chezmoi execute-template --file $HOME${/}.local${/}share${/}chezmoi${/}path${/}to${/}template.tmpl
+stdout '# contents of path[/\\]to[/\\]template.tmpl'
+
+exec chezmoi execute-template --file $HOME${/}path${/}to${/}template.tmpl
+# \S+ stands in here for the opaque $WORK directory
+stdout '# contents of \S+[/\\]home[/\\]user[/\\]path[/\\]to[/\\]template.tmpl'
+
 chhome home2/user
 
 # test that files in .chezmoidata must have known extensions
@@ -160,7 +167,11 @@ last:
 {{ "invalid template"
 -- home/user/.local/share/chezmoi/.chezmoitemplates/partial --
 {{ cat "hello" "world" }}
+-- home/user/.local/share/chezmoi/path/to/template.tmpl --
+# contents of {{ .chezmoi.sourceFile }}
 -- home/user/file.tmpl --
 # contents of file.tmpl
+-- home/user/path/to/template.tmpl --
+# contents of {{ .chezmoi.sourceFile }}
 -- home2/user/.local/share/chezmoi/.chezmoidata/unknown --
 -- home3/user/.local/share/chezmoi/.chezmoidata/.chezmoiignore --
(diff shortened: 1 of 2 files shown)


twpayne:
Thanks for this.

Please update the PR to use the commit message from #4407.

You can also run `make format` to fix the order of files in the test.

internal/cmd/testdata/scripts/executetemplate.txtar
```diff
@@ -114,6 +114,9 @@ stdout 'alphabeta'
 exec chezmoi execute-template --file $HOME${/}file.tmpl
 stdout '# contents of file.tmpl'
 
+exec chezmoi execute-template --file $HOME${/}.local/share/chezmoi/path/to/template.tmpl
+stdout '# contents of path/to/template.tmpl'
```
halostatue, line 118: Should we have a test for `.chezmoi.sourceFile` relative to `$HOME` instead of just relative to `.local/share/chezmoi`?

halostatue: I'm really curious what happens if I do `chezmoi execute-template --file /tmp/…`, but I don't think that the script execution here will do well with that, so having something for `$HOME${/}tmp/external.tmpl` with a `.chezmoi.sourceFile` inside of it would be good. We _may_ need to have that be a regex test because the `$HOME…` values are variable depending on operating system.

Also, and this is really for @twpayne, should all of the `/` in the path above be `${/}` for cross-platform testing?

overhacked: Alright, I updated the stdout patterns to be cross-platform with regard to path slashes. I checked, and `$HOME` is an artificial value created by the test harness, so it won't be different on Windows: https://github.com/twpayne/chezmoi/blob/b46fe4e5d87cc868867eb1bef5c59b0c994aa7c7/internal/cmd/main_test.go#L829-L858

I'm not sure how much the slashes actually vary in the test harness, and modern Windows accepts Unix-style slashes most places, anyways, but I modified the stdout pattern to accept either a forward or back-slash.

overhacked: @twpayne, this is ready for CI, thanks!

twpayne: > Also, and this is really for @twpayne, should all of the `/` in the path above be `${/}` for cross-platform testing?

No, it's not necessary. There are plenty of other places where this is tested.

twpayne:
Thank you for this :)

Issue #3834, opened 2024-06-26
promptString is not yielding the expected result


0xjams:
## What exactly are you trying to do?

Using the example case of defining a variable during the init step. I'm currently testing by executing the template.

```console
$ cat .chezmoi.toml.tmpl
{{- $email := promptStringOnce . "email" "Email address" -}}

[data]
    email = {{ $email | quote }}  
$ chezmoi --verbose execute-template --init --promptString email=me@home.org < ~/.local/share/chezmoi/.chezmoi.toml.tmpl

[data]
    email = "Email address"        
```

The expected result should be that the template should have been evaluated to `me@home.org`

## What have you tried so far?

- I've tried using promptStringOnce and promptString in the template file
- I've tried sending the command line parameter promptString between quotes
- I've tried using the verbose flag
- I also thought that I was defining the template function's parameters in the wrong order, but they are OK.

## Where else have you checked for solutions?

* [x] I have read [chezmoi's user guide](https://chezmoi.io/user-guide/command-overview/), and not found the answer.
* [x] I have searched [chezmoi's reference guide](https://chezmoi.io/reference/), and not found the answer.

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT    CHECK                       MESSAGE
ok        version                     v2.49.1, commit b4b55659c69fb13a502903c16dcdd566b988eece, built at 2024-06-23T12:28:09Z, built by goreleaser
ok        latest-version              v2.49.1
ok        os-arch                     linux/amd64 (Kali GNU/Linux 2024.2)
ok        uname                       Linux kali 6.8.11-amd64 #1 SMP PREEMPT_DYNAMIC Kali 6.8.11-1kali2 (2024-05-30) x86_64 GNU/Linux
ok        go-version                  go1.22.4 (gc)
ok        executable                  /usr/local/bin/chezmoi
ok        upgrade-method              sudo-upgrade-package
ok        config-file                 no config file found
warning   source-dir                  ~/.local/share/chezmoi is a git working tree (dirty)
ok        suspicious-entries          no suspicious entries
warning   working-tree                ~/.local/share/chezmoi is a git working tree (dirty)
ok        dest-dir                    ~ is a directory
ok        umask                       002
ok        cd-command                  found /usr/bin/zsh
ok        cd-args                     /usr/bin/zsh
info      diff-command                not set
ok        edit-command                found /usr/bin/vi
ok        edit-args                   /usr/bin/vi
[...]

PR #4431, opened 2025-05-07
docs: correct the usage of execute-template with the prompt string in the user guide setup

Ghost-LZW:
correct the usage of execute-template with the prompt string in the user guide setup.

according to this issue https://github.com/twpayne/chezmoi/issues/3834#issuecomment-2201211679


Ghost-LZW:
I am new to using chezmoi. When I followed the user guide, I found that the execute-template command with the prompt string does not execute as expected. After searching through the issues, I discovered that this is a documentation problem.

twpayne:
Thanks!

PR #4461, opened 2025-05-22
docs: fix link

issenn:


halostatue:
Looks good, but the commit title needs to be capitalized (`docs: Fix link`).

We can merge after that.

PR #4477, opened 2025-06-02
fix: Don't cache path lookup results on Windows systems

### LINKED ISSUE #4476: Automate the installation of Winget packages and detect available executables using lookPath

Automating some packages installation on Windows and detect them with `lookPath` template function

I'm using chezmoidata to install some Winget packages (similar to this [repository](https://github.com/jwnmulder/dotfiles/blob/1f45200d3f9a3b1f8c8013349e14442b1ea5b42f/home/.chezmoidata/packages.yaml#L12), and the installation script that comes with it.

In my PowerShell profile template, I then use the `lookPath` function to determine which executables have been installed.

```
{{ if (lookPath "xmake") -}}
Register-ArgumentCompleter -Native -CommandName xmake -ScriptBlock {
    param($commandName, $wordToComplete, $cursorPosition)
    $complete = "$wordToComplete"
    if (-not $commandName) {
        $complete = $complete + " "
    }
    $oldenv = $env:XMAKE_SKIP_HISTORY
    $env:XMAKE_SKIP_HISTORY = 1
    $results = xmake lua "private.utils.complete" $cursorPosition "nospace-json" "$complete" | ConvertFrom-Json | Sort-Object -Property value
    $results | ForEach-Object {
        $hasdesc = [bool] $_.psobject.Properties['description']
        if ($hasdesc) {
            $desc = " - $($_.description)"
        } else {
            $desc = ""
        }
        [System.Management.Automation.CompletionResult]::new($_.value, "$($_.value)$desc", 'ParameterValue', $_.value)
    }
    $env:XMAKE_SKIP_HISTORY = $oldenv
}

{{- end }}
```

After installing the package, running `chezmoi apply` for the first time results in the `lookPath` function not returning true (it works when applied a second time). I have two theories about what might be causing this:
1. Winget creates a symlink in the 'AppData\Local\Microsoft\WinGet\Links' folder, which is added to the PATH. So maybe lookPath only looks for the executable file directly.
2. The PATH env was reloaded after running the script (but `AppData\Local\Microsoft\WinGet\Links` should be present anyway).

Thank you in advance for any help or suggestions.

* [x] I have read [chezmoi's user guide](https://chezmoi.io/user-guide/command-overview/), and not found the answer.
* [x] I have searched [chezmoi's reference guide](https://chezmoi.io/reference/), and not found the answer.
* [ ] Other, please give details.

twpayne:
Might fix #4476. Draft for now. Will wait for input from Windows experts before merging.

### DIFF
--- a/internal/chezmoi/lookpath.go
+++ b/internal/chezmoi/lookpath.go
@@ -2,6 +2,7 @@ package chezmoi
 
 import (
 	"os/exec"
+	"runtime"
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


twpayne:
I don't think this fix is correct, so closing.

PR #4483, opened 2025-06-06
docs: Clarify `modify_` documentation

unlimitedsola:
I found that the documentation of the `modify_` target type [in the reference](https://www.chezmoi.io/reference/target-types/#modify-file) is a bit harder to comprehend and inconsistent with what is described [in the user guide](https://www.chezmoi.io/user-guide/manage-different-types-of-file/#manage-part-but-not-all-of-a-file) ("contains a line with the text `chezmoi:modify-template`" has a very different meaning from "containing the string").

This commit replaces the current explanation in the reference with the one that is in the user guide.

### DIFF
--- a/assets/chezmoi.io/docs/reference/target-types.md
+++ b/assets/chezmoi.io/docs/reference/target-types.md
@@ -32,14 +32,16 @@ unchanged.
 Files with the `modify_` prefix are treated as scripts that modify an existing
 file.
 
-If the file contains a line with the text `chezmoi:modify-template` then that
-line is removed and the rest of the script is executed template with the
-existing file's contents passed as a string in `.chezmoi.stdin`. The result of
-executing the template are the new contents of the file.
+If the file contains the string `chezmoi:modify-template`, then all lines
+containing that string will be removed, and the rest of the file will be
+interpreted as a template. The template is executed with the existing file's
+contents passed as a string in `.chezmoi.stdin`. The result of the template
+execution becomes the new contents of the file.
 
-Otherwise, the contents of the existing file (which maybe empty if the existing
-file does not exist or is empty) are passed to the script's standard input, and
-the new contents are read from the script's standard output.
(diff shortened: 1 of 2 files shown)


bradenhilton:
The reference is correct here, the user guide should be updated instead.

unlimitedsola:
> The reference is correct here, the user guide should be updated instead.

Hi, when I read the reference version. I understood it as a line with exact "chezmoi:modify-template" and nothing else, not in "comments" like the example shown in the user guide.

I was thinking:
```
chezmoi:modify-template
{{- .chezmoi.stdin | replaceAllRegex "old" "new" }}
```

not  as in the user guide

```
{{- .chezmoi.stdin | replaceAllRegex "old" "new" }}
```

unless the example in the user guide was not intended?

bradenhilton:
Yes, that could be improved, but the reference is still more correct. The line containing `chezmoi:modify-template` is removed before executing the rest as a template, which the user guide does not mention.

The reference should read something like "If the file contains a line containing the string `chezmoi:modify-template`, then that line is removed..." and the user guide should also mention that the line is removed as it's important context.

unlimitedsola:
> Yes, that could be improved, but the reference is still more correct. The line containing `chezmoi:modify-template` is removed before executing the rest as a template, which the user guide does not mention.
> 
> The reference should read something like "If the file contains a line containing the string `chezmoi:modify-template`, then that line is removed..." and the user guide should also mention that the line is removed as it's important context.

Thank you for the details! I will redo this later today.

unlimitedsola:
Sorry for the delay. I've adjusted the wording accordingly. Please take another look. Thanks!

The updated documentation is based on the logic here: https://github.com/twpayne/chezmoi/blob/031214d11f0afff65b467c783dd7b09d6d606cd6/internal/chezmoi/sourcestate.go#L1959-L1980

twpayne:
That's perfect, thank you very much!

PR #4488, opened 2025-06-10
fix: Remove dependency on git in packages

### LINKED ISSUE #4487: RPM package git dependency is too broad
## Describe the bug

The RPM version of chezmoi depends on the `git` package, which includes unused files such as man pages. Instead, the package should depend on `git-core`, which is purely the git tool.

## To reproduce

Attempt to install the chezmoi RPM with dnf:

`sudo dnf install chezmoi-2.62.6-x86_64.rpm`

dnf then reports having to install various other packages that are not in use by chezmoi.

## Expected behavior

The installation should not attempt to install extra dependencies (given that fedora preinstalled `git-core`),
and the following packages are not installed as a dependency:

- `git-core-doc`
- `perl-Error`
- `perl-File-Find`
- `perl-Git`
- `perl-TermReadKey`
- `perl-lib`
- etc...

## Additional context

The dependency names seem to be shared across multiple package types, but just overriding the rpm package spec to depend on `git-core` should be enough to solve the issue.

twpayne:
Fixes #4487.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -164,8 +164,6 @@ nfpms:
   - archlinux
   - deb
   - rpm
-  dependencies:
-  - git
   bindir: /usr/bin
   overrides:
     deb:


CelDaemon:
Is that possible?
I'd assume you'd still want to make sure git is installed.

twpayne:
chezmoi already has builtin git functionality sufficient to clone your dotfiles repo and pull changes, so a `git` binary is not strictly required.

Issue #4494, opened 2025-06-16
`chezmoi init` silently fails when using .chezmoiroot


mwinters0:
## Describe the bug

I'm a new user and copied this pattern from other repos on Github:

```console
% tree -a
.
├── .chezmoiroot
├── .chezmoi.toml.tmpl
├── .git
│   ├── ...
└── home
    ├── ...

% cat .chezmoiroot 
home
```

This seemingly worked previously.  But after some iteration on my repo, `chezmoi init` now does not create `~/.config/chezmoi/*`.  It instead silently exits with error code 0.  I have tried `chezmoi purge --force` with no luck, and am not sure why this seems to have stopped working.
 
This repo works if I:
- `mv home/* . && rm .chezmoiroot`
- or
- `mv .chezmoi.toml.tmpl home`

## What I've tried

None of these approaches have allowed my standard repo to work:
- **Use official binary**: I downloaded v2.62.6 from Github as `chez` (which you'll see in my output below).
- **Explicit source dir**: `git clone` into `~/myrepo` and `chezmoi init -S ~/myrepo`.

## Output of `chezmoi init --debug`

<details>

```console
% chez init --debug
time=2025-06-16T22:01:52.698Z level=INFO msg=persistentPreRunRootE version.version=2.62.6 version.commit=031214d11f0afff65b467c783dd7b09d6d606cd6 version.date=2025-06-06T18:27:16Z version.builtBy=goreleaser args="[chez init --debug]" goVersion=go1.24.4
time=2025-06-16T22:01:52.698Z level=ERROR msg=Stat component=system err="stat /home/m/.config/chezmoi/chezmoistate.boltdb: no such file or directory" name=/home/m/.config/chezmoi/chezmoistate.boltdb
time=2025-06-16T22:01:52.698Z level=ERROR msg=Mkdir component=system err="mkdir /home/m/.local/share/chezmoi: file exists" name=/home/m/.local/share/chezmoi perm=511
time=2025-06-16T22:01:52.698Z level=INFO msg=Stat component=system name=/home/m/.local/share/chezmoi
time=2025-06-16T22:01:52.698Z level=ERROR msg=Mkdir component=system err="mkdir /run/user/1000: file exists" name=/run/user/1000 perm=448
time=2025-06-16T22:01:52.698Z level=INFO msg=Stat component=system name=/run/user/1000
[...]

PR #4495, opened 2025-06-16
docs: add warnings to .chezmoiroot

mwinters0:
Per https://github.com/twpayne/chezmoi/issues/4494


twpayne:
Thank you very much! Please don't close #4494 just yet because I want to add the `chezmoi doctor` check.

PR #4516, opened 2025-07-02
feat: Add `getRedirectedURL` template function

felipecrs:
This function will allow me to drop the dependency on `curl` and `git` in my chezmoi templates, making them more portable and quicker to execute.

See the documentation for example uses of such function.

~I did not find any Go library that does this (like [Rust](https://docs.rs/follow-redirects/latest/follow_redirects/) or [JavaScript](https://www.npmjs.com/package/follow-redirects)), so I tried to keep the implementation simple.~ EDIT: this leverages [`net/http`'s redirection mechanism](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/net/http/client.go;l=456).

Disclaimer: I used Claude Sonnet 4 through GitHub Copilot to help me build this PR.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/getRedirectedURL.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/getRedirectedURL.md
@@ -1,0 +1,25 @@
+# `getRedirectedURL` *url*
+
+`getRedirectedURL` returns the final URL after following any HTTP redirects
+from the given *url*. If the *url* does not redirect, it returns the original
+*url*.
+
+`getRedirectedURL` is not hermetic: its return value depends on the state of
+the network and the remote server at the moment the template is executed.
+Exercise caution when using it in your templates.
+
+!!! example
+
+    ```
+    {{ getRedirectedURL "https://github.com/twpayne/chezmoi/releases/latest" }}
+    {{ getRedirectedURL "https://github.com/twpayne/chezmoi/raw/HEAD/README.md" }}
+    {{ getRedirectedURL "https://git.io/chezmoi" }}
+    ```
+
+    This will return something like:
+
+    ```
+    https://github.com/twpayne/chezmoi/releases/tag/v2.62.7
+    https://raw.githubusercontent.com/twpayne/chezmoi/aa57d1d773715e02103e87f78c58b99f9b91fc0c/README.md
+    https://raw.githubusercontent.com/twpayne/chezmoi/master/assets/scripts/install.sh
+    ```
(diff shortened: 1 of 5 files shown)


twpayne:
Thanks, but the default Go HTTP client will follow up to ten redirects:

https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/net/http/client.go;l=75-76

Is there a particular URL that is not covered by this case? i.e. where does the current code fail?

felipecrs:
@twpayne I added a negative test case to prove that it is not following more than one redirect.

My understanding is that this behavior is because of [ErrUseLastResponse](https://pkg.go.dev/net/http#ErrUseLastResponse).

---

However, I'm investigating if I can improve this. Hold on please.

felipecrs:
Thanks for your feedback, @twpayne. The previous approach didn't make sense indeed. I refactored it to leverage net/http's redirection logic.

Would you mind if I squash all commits before marking this PR as ready to review?

twpayne:
Please do squash the commits. I see that this PR is now in draft. Mention me here or change the PR's status to "Ready for review" when you're ready. Note that you'll also need to replace `Url` with `URL` in the Markdown documentation.

felipecrs:
@twpayne it's ready to be reviewed.

> Note that you'll also need to replace `Url` with `URL` in the Markdown documentation.

I intentionally kept the template function named with `Url` following examples like `toYaml`. I only changed the Go function because the linter complained about it.

If you think I should update the template function name to `URL` too, please let me know.

twpayne:
> Note that you'll also need to replace `Url` with `URL` in the Markdown documentation.

To clarify: chezmoi uses Go's naming conventions where possible, which includes capitalizing abbreviations. A bunch of existing template functions from [Minorminds/sprig](https://github.com/Masterminds/sprig) capitalize only the first letter (e.g. `Json`, `Toml`, and `Url`). Please use full capitals in the template function name.

felipecrs:
> > Note that you'll also need to replace `Url` with `URL` in the Markdown documentation.
> 
> To clarify: chezmoi uses Go's naming conventions where possible, which includes capitalizing abbreviations. A bunch of existing template functions from [Minorminds/sprig](https://github.com/Masterminds/sprig?rgh-link-date=2025-07-03T00%3A49%3A06Z) capitalize only the first letter (e.g. `Json`, `Toml`, and `Url`). Please use full capitals in the template function name.

Perfect. Very clear. I changed it.

twpayne:
Thank you for your very nice contribution :)

Issue #4522, opened 2025-07-04
chezmoi.exe panics when running diff against a WSL-hosted source directory


yu64n:
## Describe the bug

When the Windows-native `chezmoi.exe` binary is asked to operate on a source directory that lives inside WSL (via the `-S \\wsl.localhost\<Distro>\...` flag), the command panics instead of producing a diff.  

## To reproduce

1. On a Windows host, install chezmoi via WinGet (winget install twpayne.chezmoi).  
2. Ensure you have an existing chezmoi source in WSL, e.g.  
   `\\wsl.localhost\Ubuntu\home\<user>\.local\share\chezmoi`  
3. From PowerShell, run:

   ```powershell
   C:\Path\To\chezmoi.exe -S \\wsl.localhost\Ubuntu\home\<user>\.local\share\chezmoi diff
   ```
4. Observe the panic.

## Expected behavior

chezmoi diff should list the differences between the source-state repository in WSL and the Windows target files without throwing a panic.

## Output of command with the `--verbose` flag

```console
$ chezmoi -S \\wsl.localhost\Ubuntu\home\<user>\.local\share\chezmoi -v diff
chezmoi: warning: config file template has changed, run chezmoi init to regenerate config file
panic: /WSL.LOCALHOST/UBUNTU/home/<user>/.local/share/chezmoi/.Brewfile: not in //WSL.LOCALHOST/UBUNTU/home/<user>/.local/share/chezmoi

goroutine 1 [running]:
github.com/twpayne/chezmoi/internal/chezmoi.AbsPath.MustTrimDirPrefix(...)
        /home/runner/work/chezmoi/chezmoi/internal/chezmoi/abspath.go:84
github.com/twpayne/chezmoi/internal/chezmoi.(*SourceState).Read.func2({0xc000592cc0, 0x40}, {0x1b6dd88, 0xc00017b2d0}, {0x0?, 0x0?})
        /home/runner/work/chezmoi/chezmoi/internal/chezmoi/sourcestate.go:970 +0x1515
github.com/twpayne/chezmoi/internal/chezmoi.walkSourceDirHelper({0x1b77020, 0xc0003a2050}, {0xc000592cc0, 0x40}, {0x1b6dd88, 0xc00017b2d0}, 0xc0001ab0a8)
        /home/runner/work/chezmoi/chezmoi/internal/chezmoi/system.go:203 +0x6c
[...]

PR #4525, opened 2025-07-05
fix: Use filepath.Join() to join abspath

yu64n:
Fixes https://github.com/twpayne/chezmoi/issues/4522

The method `AbsPath.JoinString` eventually calls `path.Join`, which strips the leading `\\` from UNC paths. In this case, `AbsPath.JoinString` is invoked here:

https://github.com/twpayne/chezmoi/blob/5188a2e950a2c36aeb5d1c7ae8059989e2b3e276/internal/chezmoi/system.go#L229


twpayne:
Thank you very much for both reporting the problem and providing the fix!

I'll release a new version of chezmoi that includes your fix in the middle of next week (I'll wait for the release of Go 1.24.5, then do the release).

PR #4526, opened 2025-07-06
feat: Add ensureLinePrefix template function

### LINKED ISSUE #4518: `comment` template function should not comment commented lines
## Describe the bug

The `comment` function should not add the prefix if the line already has it.

## To reproduce

```console
$ chezmoi execute-template '{{ "# commented line" | comment "# " }}'
# # commented line
```

## Expected behavior

```diff
$ chezmoi execute-template '{{ "# commented line" | comment "# " }}'
- # # commented line
+ # commented line
```

N/A

## Output of `chezmoi doctor`

N/A (chezmoi 2.62.7)

## Additional context

I want to use `comment` function in some my `modify_vscode.sources`:

```

{{- comment "# " .chezmoi.stdin -}}
```

Original file:

```
### THIS FILE IS AUTOMATICALLY CONFIGURED ###
# You may comment out this entry, but any other modifications may be lost.
Types: deb
URIs: https://packages.microsoft.com/repos/code
Suites: stable
Components: main
Architectures: amd64,arm64,armhf
Signed-By: /usr/share/keyrings/microsoft.gpg
```

Expected output:

```
# ### THIS FILE IS AUTOMATICALLY CONFIGURED ###
# You may comment out this entry, but any other modifications may be lost.
# Types: deb
# URIs: https://packages.microsoft.com/repos/code
# Suites: stable
# Components: main
# Architectures: amd64,arm64,armhf
# Signed-By: /usr/share/keyrings/microsoft.gpg
```

Note however a nitpick: what if the line starts with `###`?

I was thinking if it would be ok to tolerate existing lines that starts with `commentPrefix.TrimSuffix(" ")`, so that the output of the example above would be:

```
### THIS FILE IS AUTOMATICALLY CONFIGURED ###
# You may comment out this entry, but any other modifications may be lost.
# Types: deb
# URIs: https://packages.microsoft.com/repos/code
# Suites: stable
# Components: main
# Architectures: amd64,arm64,armhf
# Signed-By: /usr/share/keyrings/microsoft.gpg
```

I think it may be ok to have some gymnastics, considering the function is named `comment` and not `addPrefixToEveryLine`. I am not aware of any language or dialect whose lack of an empty space ` ` after comment symbols would disqualify a line from being considered a comment.

What do you think?

twpayne:
Fixes #4518.

Draft for now, as the documentation could be much better.

cc @felipecrs for feedback.

### DIFF
--- a/assets/chezmoi.io/docs/reference/templates/functions/comment.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/comment.md
@@ -2,8 +2,13 @@
 
 `comment` returns *text* with each line prefixed with *prefix*.
 
+`comment` is typically used to comment out blocks of multi-line text
+unconditionally. In contrast, the [`ensureLinePrefix` template
+function](ensureLinePrefix.md) can be used to only comment out lines that are
+not already comments.
+
 !!! example
 
     ```
-    {{ "Header" | comment "# " }}
+    {{ "Line 1\nLine 2\n" | comment "# " }}
     ```
--- a/assets/chezmoi.io/docs/reference/templates/functions/ensureLinePrefix.md
+++ b/assets/chezmoi.io/docs/reference/templates/functions/ensureLinePrefix.md
@@ -1,0 +1,17 @@
+# `ensureLinePrefix` *prefix* [*prefix-to-add*] *text*
+
+`ensureLinePrefix` ensures that each line of *text* starts with *prefix*. If any
+line does not start with *prefix* then *prefix-to-add* is prepended to that
+line.
+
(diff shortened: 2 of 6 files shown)


felipecrs:
@twpayne this looks perfect and is exactly what I need.

I agree with not changing the existing comment function as that would be a potential breaking change. 

One small feedback documentation-wise is to reference this function in comment function's documentation, and maybe clarify the difference between the two. I'm sure it will generate some confusion otherwise.

twpayne:
Good idea. I've add more details in the documentation.

PR #4536, opened 2025-07-10
docs: Fix minor error in example hook

tylerbutler:
I think the example hook is incorrect. Since it's a script, not a command, the config should use `script`.


twpayne:
This is not quite correct. If the script has the executable bit set then it can be executed directly. Using `command`, rather than `script`, also avoids the complexities of shell expansion of arguments.

tylerbutler:
> This is not quite correct. If the script has the executable bit set then it can be executed directly. Using `command`, rather than `script`, also avoids the complexities of shell expansion of arguments.

@twpayne thanks for the explanation. Might this be worth a note in the docs? I found it confusing, anyway. I can open a PR with some doc suggestions if it would be helpful.

twpayne:
Yes, please do open a PR with doc suggestions.

PR #4521, opened 2025-07-04
feat: Backup replaced files option added to init, apply, update, and edit

arran4:
This change enables a `--backup=file.tar` flag on `edit` `init`, `apply` and `update` (strictly speaking it's a global flag.)

I believe i have covered testing and documentation which were the things I neglected last time.

`make format` gave me this but I don't believe that this is a code issue:

```
$ make format
if [ ! -x bin/golangci-lint ] || ( ./bin/golangci-lint version | grep -Fqv "version 2.1.6" ) ; then \
        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- v2.1.6 ; \
fi
ERRO Can't read config: can't unmarshal config by viper: 1 error(s) decoding:

* 'Version' expected a map, got 'string' 
./bin/golangci-lint fmt
ERRO Can't read config: can't unmarshal config by viper: 1 error(s) decoding:

* 'Version' expected a map, got 'string' 
make: *** [Makefile:130: format] Error 3
```

And lint

```
$ make lint
if [ ! -x bin/actionlint ] || ( ./bin/actionlint --version | grep -Fqv "v1.7.7" ) ; then \
        GOBIN=/home/arran/Documents/Projects/chezmoi/bin go install "github.com/rhysd/actionlint/cmd/actionlint@v1.7.7" ; \
fi
go: downloading github.com/rhysd/actionlint v1.7.7
go: downloading github.com/bmatcuk/doublestar/v4 v4.8.0
go: downloading golang.org/x/sys v0.29.0
go: downloading github.com/mattn/go-shellwords v1.0.12
go: downloading github.com/robfig/cron/v3 v3.0.1
if [ ! -x bin/editorconfig-checker ] || ( ./bin/editorconfig-checker --version | grep -Fqv "v3.3.0" ) ; then \
        curl -sSfL "https://github.com/editorconfig-checker/editorconfig-checker/releases/download/v3.3.0/ec-linux-amd64.tar.gz" | tar -xzf - "bin/ec-linux-amd64" ; \
        mv "bin/ec-linux-amd64" bin/editorconfig-checker ; \
fi
if [ ! -x bin/golangci-lint ] || ( ./bin/golangci-lint version | grep -Fqv "version 2.1.6" ) ; then \
        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- v2.1.6 ; \
fi
[...]

PR #4548, opened 2025-07-14
feat: Added ignore query and modification command line arguments

arran4:
This adds some ignore file modification code:
```
chezmoi ignore add '*.log'
chezmoi ignore remove README.md
chezmoi ignore activate '*.tmp'
chezmoi ignore deactivate notes.md
chezmoi ignore query foo.log
```

It does assume a simple configuration. So advanced .toml usages might be an issue. 

It doesn't yet support `!` negative matches. However the usecase for this is that I do a 'managed' and 'unmanaged' loop which I was hoping to get a faster way to iterate on that, without jumping into vim. -- It has been a while it was on the back of my mind to do this.

Parts of the code were generated but I've gone over everything. Testing: Txtar is all mine.

Tests work on my fork and machine.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/ignore.md
+++ b/assets/chezmoi.io/docs/reference/commands/ignore.md
@@ -1,0 +1,50 @@
+# `ignore`
+
+Add, remove, activate, or deactivate patterns in `.chezmoiignore`.
+
+## Subcommands
+
+### `add [pattern]...`
+
+Add patterns to `.chezmoiignore`. If the pattern already matches an ignored
+entry, the command fails unless `-f`/`--force` is specified. If the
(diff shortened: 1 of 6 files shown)


twpayne:
As with #4521, please read the [contributing guide](https://www.chezmoi.io/developer-guide/contributing-changes/) before making contributions guide. Specifically:

> If you want to make a more significant change, please first [open an issue](https://github.com/twpayne/chezmoi/issues/new/choose) to discuss the change that you want to make. Dave Cheney gives a [good rationale](https://dave.cheney.net/2019/02/18/talk-then-code) as to why this is important.

`.chezmoiignore` files are templates so they cannot be safely manipulated by adding and removing lines.

PR #4551, opened 2025-07-14
chore: Un-shard tests on GitHub Actions

twpayne:
With the startup time improvements, sharding is no longer strictly needed.


twpayne:
Un-sharding on Windows does not work for reasons that I do not understand, so I kept sharding there.

PR #4568, opened 2025-07-23
fix: Fix bad skip instruction in test script issue4479 

FilippoBonazziSUSE:
If Git is not found in `$PATH`, the test looks like it should be skipped with the message 'git not found in $PATH'. Instead, the test fails with:

```go
[   45s]         testscript.go:584: > [!exec:git] 'skip git not found in $PATH'
[   45s]             FAIL: testdata/scripts/issue4479.txtar:1: unknown command "skip git not found in $PATH"
```


twpayne:
Good catch, thanks!

PR #4576, opened 2025-07-31
feat: Add automatic Bitwarden CLI unlock

twpayne:


halostatue (approved):
LGTM, but I don't have bitwarden installed to test it.

halostatue:
> It works! ❤️

Good to hear!

> Worth noting that `bw login` and `bw unlock` are different commands, and if the user has never run `bw login` on the machine, `bw unlock` fails:

This is as it should be and matches the behaviour of the 1Password integration. chezmoi should not be connecting accounts, but should only be unlocking already connected accounts.

I believe that a `bw login` can be done in a `run_once_before_` script, but I wouldn't use it personally (I manually configure `op` for the first time with 1Password).

twpayne:
Thank you for the review @halostatue and testing @mkasberg :)

I wondered about automating `bw login` but, as there seem to be multiple possible login flows (e.g. using SSO, an email, or an API key), and I'm not a Bitwarden user, I'm not sure that I'd get this right. So, let's keep that for a possible follow-up PR.

PR #4589, opened 2025-08-05
chore: Update dependencies

twpayne:


twpayne:
Actually, I'll wait for Go 1.24.6 (will be released tomorrow during US business hours) before merging this.

PR #4612, opened 2025-08-21
feat: Add `podman` and `ssh` commands

### LINKED ISSUE #1410: Remote install dotfiles over ssh without installing chezmoi on the machine

It would be nice if chezmoi could install my dotfiles on a machine that I can ssh into, without me having to install chezmoi on that machine. This would be very helpful for ephemeral virtual machines and dev containers, for example.

I would like to able to run:
```console
$ chezmoi apply --remote=machine.example.com
```
and have chezmoi install my dotfiles over SSH on `machine.example.com` without me having to install chezmoi on `machine.example.com`.

chezmoi already has much of the functionality to make this possible, it can roughly be simulated with:
```console
$ chezmoi archive | ssh machine.example.com -- tar -xf -
```

However, `chezmoi archive` will generate dotfiles for the local machine, not the remote machine. This can be mitigated with
```console
$ chezmoi --config=remote-chezmoi-config.toml archive | ssh machine.example.com -- tar -xf -
```
where `remote-chezmoi-config.toml` is a chezmoi configuration file for the remote machine. However, the use of any non-hermetic template variables, e.g. `.chezmoi`, or the use of non-hermetic template functions, e.g. `env`, `output`, or `lookPath` will still apply to the local machine, not the remote machine.

Possible ways around this include:
* Requiring the user's templates to not use any non-hermetic template variables or functions. This is possible now.
* Using a two-phase SSH: in the first phase a SSH connection is used to retrieve the value of `.chezmoi`  from the remote machine, and in the second phase the archive is generated using the value of `.chezmoi` and extracted on the remote machine. Non-hermetic template functions are not likely to be supported.

## Additional context

* The remote install functionality should also work for local Docker containers.
* One-shot installation of dotfiles is probably sufficient. There is no need to support `diff` for remotes.

### LINKED ISSUE #3676: Add `chezmoi ssh` command to ssh into a machine with your personal configuration

I'm always frustrated when I log into a machine that I've never logged into before and don't have my personal environment set up, even if that machine is shared with other people.

I would like to run

```console
$ chezmoi ssh somemachine.com
```

and SSH into `somemachine.com` with my personal setup activated on `somemachine.com`. For example, I should be in my favorite shell with my personal config files active.

#1410 describes installing dotfiles remotely over ssh.

#3675 describes wanting your own configuration on a machine shared with other people.

## Additional context

This is a stretch goal for chezmoi. Go has [excellent support for ssh](https://pkg.go.dev/golang.org/x/crypto/ssh), and this is mostly possible by setting environment variables like `$XDG_CONFIG_HOME` for the ssh session.

It's technically possible, but please don't expect anything soon.

cc @old-pigeon

twpayne:
Fixes #1410.
Fixes #3676.

This is a draft PR, but basically allows you to connect to a remote machine or container and get your shell with your dotfiles installed all in a single command:

```bash
# start a new container from $IMAGE_ID, install chezmoi,
# run chezmoi init --apply $GITHUB_USERNAME, and exec your shell
chezmoi podman run $IMAGE_ID $GITHUB_USERNAME

# start a new shell in an existing container, install chezmoi,
# run chezmoi init --apply $GITHUB_USERNAME, and exec your shell
chezmoi podman exec $CONTAINER_ID $GITHUB_USERNAME

# ssh into $HOSTNAME, install chezmoi,
# run chezmoi init --apply $GITHUB_USERNAME, and exec your shell
chezmoi ssh $HOSTNAME $GITHUB_USERNAME
```

This means, that, for example, I can test my dotfiles in a new, Ubuntu container with:

```bash
chezmoi podman run ubuntu:latest twpayne
```

Or, if I want to login to some host and have my shell and my dotfiles, I can do:

```bash
chezmoi ssh hostname twpayne
```

**Please be really careful with the `chezmoi ssh` command**. This will overwrite your dotfiles in the machine you SSH into, and may do other damage to the target machine. Please, only use it on machines that you don't mind destroying!

Credit to @cdown's https://github.com/cdown/sshrc (originally authored by @Russell91) for the idea of running a shell script to install your dotfiles on the remote machine. The major difference with sshrc is that sshrc only handles a few static files, whereas the `chezmoi ssh` and `chezmoi podman` give you the full power of chezmoi.

This PR is ready for testing, but still needs a few more things completed before it's ready to merge:

* [ ] Testing on more UNIX-y distributions (I've only really test Ubuntu and Alpine in podman containers so far)
* [x] Complete documentation
* [ ] Add a `chezmoi docker` command, equivalent to `chezmoi podman`, except using Docker

All feedback welcome!

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/generate.md
+++ b/assets/chezmoi.io/docs/reference/commands/generate.md
@@ -2,14 +2,16 @@
 
 Generates *output* for use with chezmoi. The currently supported *output*s are:
 
-| Output               | Description                                                           |
-| -------------------- | --------------------------------------------------------------------- |
-| `git-commit-message` | A git commit message, describing the changes to the source directory. |
-| `install.sh`         | An install script, suitable for use with GitHub Codespaces            |
+| Output                  | Description                                                                   |
+| ----------------------- | ----------------------------------------------------------------------------- |
+| `git-commit-message`    | A git commit message, describing the changes to the source directory.         |
+| `install.sh`            | An install script, suitable for use with GitHub Codespaces                    |
+| `install-init-shell.sh` | A script which installs chezmoi, runs `chezmoi init`, and executes your shell |
 
 ## Examples
 
 ```sh
 chezmoi generate install.sh > install.sh
 chezmoi git commit -m "$(chezmoi generate git-commit-message)"
+chezmoi generate install-init-shell.sh $GITHUB_USERNAME
 ```
--- a/assets/chezmoi.io/docs/reference/commands/podman.md
+++ b/assets/chezmoi.io/docs/reference/commands/podman.md
@@ -1,0 +1,50 @@
+# `podman`
+
+!!! Warning
+
+    `podman` is an experimental command.
+
+Install chezmoi, run `chezmoi init --apply`, and optionally execute your shell
+in podman containers.
+
+## Subcommands
+
+### `exec` *container-id* *init-args*...
+
+Install chezmoi, run `chezmoi init --apply *init-args*`, and execute your shell
+in the existing podman container *container-id*.
+
+#### Flags
+
+#### `-i`, `--interactive`
+
+Keep stdin open even if not attached.
+
+#### `-p`, `--package`
+
+Install chezmoi using the distribution's package manager, if possible.
+Otherwise, fall back to `curl` or `wget` installation. If neither `curl` nor
+`wget` are installed then install them with the distribution's package manager.
+
+#### `-s`, `--shell`
+
+After installing chezmoi, initializing your dotfiles, execute your shell. This
+is the default.
+
+### `run` *image-id* *init-args*...
+
+Create a new podman container using *image-id*, and in it, install chezmoit, run
+`chezmoi init --apply *init-args*`, and execute your shell.
+
+#### `-p`, `--package`
+
+Install chezmoi using the distribution's package manager, if possible.
+Otherwise, fall back to `curl` or `wget` installation. If neither `curl` nor
+`wget` are installed then install them with the distribution's package manager.
+
+## Examples
+
+```sh
+chezmoi podman exec $CONTAINER_ID $GITHUB_USERNAME
+chezmoi podman run alpine:latest $GITHUB_USERNAME
+```
--- a/assets/chezmoi.io/docs/reference/commands/ssh.md
+++ b/assets/chezmoi.io/docs/reference/commands/ssh.md
@@ -1,0 +1,28 @@
+# `ssh` *host* *init-args*...
+
+!!! Warning
+
+    `ssh` is an experimental, potentially destructive, command.
+
+SSH to *host*, install chezmoi, run `chezmoi init --apply *init-args*`, and
+executes your shell.
+
+## Flags
+
+### `-p`, `--package`
+
+Install chezmoi using the distribution's package manager, if possible.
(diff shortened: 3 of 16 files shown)


13steinj:
> All feedback welcome!

I want to go out of my way to explicitly comment and say thank you for adding this functionality (the podman/docker support in particular)! I have long been using a much more manual set of scripts to do this.

Some feedback:

* The mechanism by which the package manager is selected is not particularly robust, and some people may explicitly want to use a "system" package manager that is not it. For example, my org's dev machines are rhel9; but devs install packages for themselves via nix / home-manager instead. The initial chezmoi install is performed via `nix run our_caching_proxy/...nixpkgs...25.05.tar.gz#chezmoi...`. At home, I run linuxbrew and get chezmoi from there instead of from snap/apt. Some level of customizability here would be appreciated-- I think simply being able to concretely specify _which_ package manager as an argument would suffice.
* It would be nice if I could provide the path to a Dockerfile / Containerfile / docker-bake file to build the image from
* it would be nice if more than docker and podman are supported; as there are a few CLI tools that work here. I would suggest simply keeping the `docker` subcommand (it is my understanding that while officially everything is the "OCI ecosystem" everyone in common parlance just says "docker"), and then people can simply `chezmoi -c <config file with docker information if their own does not have it, e.g. in CI environments> docker ...`
  * it may be nice if `-c <config string>` worked as well as `-c <config path>` for things like smoke tests in CI (but I don't care too much about this, `-c path` is good enough for me).
  
Happy to contribute any/all of the suggestions above when I have the time (hopefully the weekend) if desired.

twpayne:
Thank you very much for the feedback and I'm glad that you like the feature :)

> * I think simply being able to concretely specify which package manager as an argument would suffice.

Great idea! I implemented this in #4638.

> * It would be nice if I could provide the path to a Dockerfile / Containerfile / docker-bake file to build the image from
> * it would be nice if more than docker and podman are supported; as there are a few CLI tools that work here.

> Happy to contribute any/all of the suggestions above when I have the time (hopefully the weekend) if desired.

Please do!

Some notes:
* Please base your changes on #4638.
* Don't hesitate to delete code if it makes things simpler. Specifically, for supporting other CLI tools, please remove all traces of the `podman` command.

> * it may be nice if -c <config string> worked as well as -c <config path> for things like smoke tests in CI (but I don't care too much about this, -c path is good enough for me).

You can achieve this already with bash's process substitution. Run this:

```console
$ echo <(echo foo)
/dev/fd/12
```

Here bash replaces the first argument to the outer echo with a temporary filename of a file containing the output of the command inside the `<(`...`)`. Technically, the filename is actually the name of a pipe connected to the standard output of the command inside the `<(`...`)`, but the effect is mostly the same. Of course, you can replace `echo foo` with whatever command you like.

Thanks again for the feedback and offer to contribute. Much appreciated.

twpayne:
> * it would be nice if more than docker and podman are supported; as there are a few CLI tools that work here. I would suggest simply keeping the `docker` subcommand (it is my understanding that while officially everything is the "OCI ecosystem" everyone in common parlance just says "docker"), and then people can simply `chezmoi -c <config file with docker information if their own does not have it, e.g. in CI environments> docker ...`

Good idea. I implemented this in #4643.

> * It would be nice if I could provide the path to a Dockerfile / Containerfile / docker-bake file to build the image from

I'm not sure how to do this. What would the arguments to `docker` look like for the `exec` and `run` commands to achieve this?

PR #4640, opened 2025-09-07
feat: Add support for specifying a SHA with git-repo

parsiad:
I ran into a performance issue while trying to use `.chezmoiexternal.toml`. I wanted to use a tarball at a specific git SHA:

```
[".icons/dracula-icons"]
exact = true
type = "archive"
stripComponents = 1
url = "https://github.com/m4thewz/dracula-icons/archive/de2a8edd94608ba0ac4dcf5a187af0ffaa511ebc.tar.gz"
```

However, I think the `archive` implementation is creating individual `SourceStateFile` entries for every single file in the archive. Then during `chezmoi apply`, each file gets processed individually with its own diff calculation. The tarball above is around 70MB and can take minutes to apply.

The easiest fix was to allow cloning and checking out at a specific git SHA:

```
[".icons/dracula-icons"]
commit = "de2a8edd94608ba0ac4dcf5a187af0ffaa511ebc"
type = "git-repo"
url = "https://github.com/m4thewz/dracula-icons.git"
```

This is implemented in the PR.

Note: I am not particularly familiar with the chezmoi code and wanted to make my change as minimal as possible. As such, I strung `git clone` and `git checkout` commands in a shell. The reliance on `sh` makes it platform-dependent.

### DIFF
--- a/assets/chezmoi.io/docs/reference/special-files/chezmoiexternal-format.md
+++ b/assets/chezmoi.io/docs/reference/special-files/chezmoiexternal-format.md
@@ -48,6 +48,7 @@ Entries may have the following fields:
 | `checksum.sha512`            | string   | *none*        | Expected SHA512 checksum of data                                 |
 | `checksum.size`              | int      | *none*        | Expected size of data                                            |
 | `clone.args`                 | []string | *none*        | Extra args to `git clone`                                        |
+| `commit`                     | string   | *none*        | Git commit SHA to checkout for `git-repo` type                   |
 | `filter.command`             | string   | *none*        | Command to filter contents                                       |
 | `filter.args`                | []string | *none*        | Extra args to command to filter contents                         |
 | `pull.args`                  | []string | *none*        | Extra args to `git pull`                                         |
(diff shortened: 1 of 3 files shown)


halostatue:
This is not necessary:

```toml
[".icons/dracula-icons"]
clone.args = ["--revision", "de2a8edd94608ba0ac4dcf5a187af0ffaa511ebc"]
type = "git-repo"
url = "https://github.com/m4thewz/dracula-icons.git"
```

```toml
[".icons/dracula-icons"]
clone.args = ["--branch", "circle"] # This can be a branch or a tag
type = "git-repo"
url = "https://github.com/m4thewz/dracula-icons.git"
```

If a shorthand for this is desired (I haven't used those parameters for `git clone` in the past), it should be `clone.ref` where `ref` can be a branch, tag, or revision hashref. The heuristic would be that if the `ref` is all hex (uppercase or lowercase), it would convert into `["--revision", ref]` otherwise it would convert into `["--branch", ref]`.

parsiad:
Thanks! I didn't realize I could use `clone.args = ["--revision", "de2a8edd94608ba0ac4dcf5a187af0ffaa511ebc"]`. I'm closing the PR.

PR #4642, opened 2025-09-09
docs: add Fedora and RHEL install commands

mikelolasagasti:
Packages have been built for all current Fedora releases and EPEL9 & EPEL10, they will land in stable in a few days: https://bodhi.fedoraproject.org/updates/?packages=chezmoi


twpayne:
Fantastic! Thank you so much for your work in getting chezmoi into Fedora and EPEL!

PR #4660, opened 2025-09-17
feat: don't request persistent state lock for `{source,target}-path` cmds

vergenzt:

### DIFF
--- a/internal/cmd/sourcepathcmd.go
+++ b/internal/cmd/sourcepathcmd.go
@@ -16,7 +16,7 @@ func (c *Config) newSourcePathCmd() *cobra.Command {
 		ValidArgsFunction: c.targetValidArgs,
 		RunE:              c.runSourcePathCmd,
 		Annotations: newAnnotations(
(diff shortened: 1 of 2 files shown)


twpayne:
I don't think this is correct. Evaluating the source path and target paths requires executing templates, which might update the persistent state.

vergenzt:
> I don't think this is correct. Evaluating the source path and target paths requires executing templates, which might update the persistent state.

Thanks for that context! How might evaluating source or target path require executing a template? (Not doubting you, I just don't understand where that would happen currently.)

twpayne:
The source or target path might come from an external. The external is defined in `.chezmoiexternal.toml`, which is a template. The template might call a function like `gitHubLatestRelease` which stores data in the persistent state.

vergenzt:
> The source or target path might come from an external. The external is defined in `.chezmoiexternal.toml`, which is a template. The template might call a function like `gitHubLatestRelease` which stores data in the persistent state.

Got it. Thanks!

PR #4672, opened 2025-09-22
chore: update snap base

### LINKED ISSUE #4671: snap: update to newer base
chezmoi's snap release is currently using the archaic `core` base from 2016.

If it needs no dependencies at all (not even libc), then I think it could be built with `base: bare`.

Otherwise, `core24` is current.

References:
- https://goreleaser.com/customization/snapcraft/
- https://documentation.ubuntu.com/snapcraft/stable/how-to/crafting/specify-a-base/

keturn:
remove snap dependency on obsolete core base

Using the `bare` base for this Works On My Machine. I don't know enough about go builds to know if that's sufficient. `core24` would be a more conservative choice.

See https://documentation.ubuntu.com/snapcraft/stable/reference/bases/#base-snap-reference

Fixes #4671

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -288,6 +288,7 @@ snapcrafts:
   grade: stable
   confinement: classic
   license: MIT
+  base: bare
   apps:
     chezmoi:
       command: chezmoi


keturn:
maybe it's that a `nocgo` build could use a `bare` base but a `cgo-libc` build should use `core24`?

halostatue:
I don't use Linux regularly, but I think that `bare` should be good because chezmoi is statically linked. If we choose something else, it should be `core22` for greater compatibility with more libc versions.

bradenhilton:
@keturn FYI, the failing check is because the first letter of the commit message needs to be uppercase. If you amend the commit message it should pass.

twpayne:
Where is it documented that `bare` is no longer supported? As far as I can tell it's still on the list of supported bases in https://documentation.ubuntu.com/snapcraft/stable/reference/bases/#base-snap-reference.

twpayne:
Ah sorry, I misread the initial issue. Yes, using `bare` makes sense.

twpayne:
Thank you all!

PR #4673, opened 2025-09-23
fix rcm details

docwhat:
RCM is written in bash. However, the build system uses lots of perl, which is why it shows up in GitHub as being mostly perl.

In addition, RCM uses symlinks. It's a fancier version of GNU Stow.


halostatue:
Please look at the contribution document on the website to fix your commit title: all commits must be in conventional style.

"docs: Fix RCM details" would be better.

twpayne:
Thanks, I've merged and then edited the commit message.

PR #4678, opened 2025-09-27
feat: add --less-interactive to only prompt for conflicts

dnnr:


internal/cmd/config.go
```diff
@@ -1074,64 +1075,64 @@ func (c *Config) defaultPreApplyFunc(
...
-			case choice == "quit":
-				return chezmoi.ExitCodeError(1)
-			default:
-				panic(choice + ": unexpected choice")
-			}
+	default: // Prompt in *some* cases of a dirty target:
+		switch {
+		case targetEntryState.Overwrite():
+		case targetEntryState.Type == chezmoi.EntryStateTypeScript:
+		case lastWrittenEntryState == nil:
```
twpayne, line 1102: In Go, `switch` cases `break` by default unless you add an explicit `fallthrough` statement. This unlike C or C++ where `switch` cases fall through by default unless you add an explicit `break` statement. So, most of the cases in this `switch` have no effect. Is this correct?

twpayne:
Thank you! This is a nice improvements. I'll make some minor non-functional changes to the code and add some tests in a follow-up PR. Thank you for the contribution!

PR #4681, opened 2025-09-28
Better help

twpayne:
This groups the commands in the output of `chezmoi help` to emphasize popular commands.

After this PR the output of `chezmoi help` looks like:

```
...

Documentation commands:
  doctor               Check your system for potential problems
  help                 Print help about a command
  license              Print license

Daily commands:
  add                  Add an existing file, directory, or symlink to the source state
  apply                Update the destination directory to match the target state
  chattr               Change the attributes of a target in the source state
  diff                 Print the diff between the target state and the destination state
  edit                 Edit the source state of a target
  forget               Remove a target from the source state
  init                 Setup the source directory and update the destination directory to match the target state
  merge                Perform a three-way merge between the destination state, the source state, and the target state
  merge-all            Perform a three-way merge for each modified file
  re-add               Re-add modified files
  status               Show the status of targets
  update               Pull and apply any changes

Template commands:
  cat                  Print the target contents of a file, script, or symlink
  data                 Print the template data
  execute-template     Execute the given template(s)

Advanced commands:
  cd                   Launch a shell in the source directory
  edit-config          Edit the configuration file
  edit-config-template Edit the configuration file template
  generate             Generate a file for use with chezmoi
  git                  Run git in the source directory
  ignored              Print ignored targets
  managed              List the managed entries in the destination directory
  unmanaged            List the unmanaged files in the destination directory
  verify               Exit with success if the destination state matches the target state, fail otherwise

Encryption commands:
  age                  Interact with age
  age-keygen           Generate an age identity or convert an age identity to an age recipient
  decrypt              Decrypt file or standard input
  encrypt              Encrypt file or standard input

Remote commands:
  docker               Use your dotfiles in a Docker container
  ssh                  SSH to a host and initialize dotfiles

Migration commands:
  archive              Generate a tar archive of the target state
  destroy              Permanently delete an entry from the source state, the destination directory, and the state
  import               Import an archive into the source state
  purge                Purge chezmoi's configuration and data
  upgrade              Upgrade chezmoi to the latest released version

Internal commands:
  cat-config           Print the configuration file
  completion           Generate shell completion code
  dump                 Generate a dump of the target state
  dump-config          Dump the configuration values
  secret               Interact with a secret manager
  source-path          Print the source path of a target
  state                Manipulate the persistent state
  target-path          Print the target path of a source path

Flags:
...
```

### DIFF
--- a/internal/cmd/addcmd.go
+++ b/internal/cmd/addcmd.go
@@ -39,6 +39,7 @@ type addCmdConfig struct {
 
 func (c *Config) newAddCmd() *cobra.Command {
 	addCmd := &cobra.Command{
+		GroupID: groupIDDaily,
 		Use:     "add targets...",
 		Aliases: []string{"manage"},
 		Short:   "Add an existing file, directory, or symlink to the source state",
--- a/internal/cmd/agecmd.go
+++ b/internal/cmd/agecmd.go
@@ -25,9 +25,10 @@ type ageCmdConfig struct {
 
 func (c *Config) newAgeCmd() *cobra.Command {
 	ageCmd := &cobra.Command{
-		Use:   "age",
-		Args:  cobra.NoArgs,
-		Short: "Interact with age",
+		GroupID: groupIDEncryption,
+		Use:     "age",
+		Args:    cobra.NoArgs,
+		Short:   "Interact with age",
 		Annotations: newAnnotations(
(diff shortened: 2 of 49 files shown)


twpayne:
I would like to update the structure of chezmoi.io to use the same grouping, but I'll do that in a follow-up PR.

Issue #4628, opened 2025-08-31
Folder containing both regular files and external files shows up twice in chezmoi managed


fleetingbytes:
## Describe the bug

See [this discussion](https://github.com/twpayne/chezmoi/discussions/4620#discussion-8804822)

## To reproduce

Have both regular and external files in a folder. Run `chezmoi managed`. The name of the folder appeasr twice

## Expected behavior

The name of the folder appears only once

## Output of command with the `--verbose` flag

See [this discussion](https://github.com/twpayne/chezmoi/discussions/4620#discussion-8804822)

PR #4683, opened 2025-09-29
fix: duplicate occurrences on managed cmd (#4628)

LiviaLelis:
Hi, I was looking at the issues and found #4628, this is a simple solution for it based on de-duplicating the paths list after it's built. I also though about checking it on the initial creation, but I wasn't sure if it was a stable approach for this scenario.

I didn't setup reproduction of the original issue for now, but it seems to be no regression from this addition.

I added tests for the `Distinct` function I created for this purpose, but I haven't added new tests for the `managed` command, I tried to do so but didn't understand how to replicate the external files scenario on there. Although I believe it's a non-blocker, help on that would be appreciated :) (I'm also curious on how that would look).


LiviaLelis:
There seems to be some `go.sum` changes resulting from my `go tidy` run, if you'd like me to remove it from the commit, just let me know :)

twpayne:
Thanks for this, but I'd like to understand why the duplicate paths are occurring rather than just patching over the problem.

The changes to the `go.sum` file are likely because you ran `go mod tidy` and didn't run `make test-release` afterwards. Building for different OS/architectures pulls in a extra dependencies which `go mod tidy` (which only knows about your current OS/architecture) does not know about.

LiviaLelis:
So, about the root cause, I just reproduced the issue locally (and now know how to write tests for it :D) and added some debug prints to understand the state of the tree, here are my findings:

```
[DEBUG] Checking path '.local/share' (*chezmoi.SourceStateDir): *chezmoi.SourceStateDir
  chezmoi.SourceStateDir {
    Attr: chezmoi.DirAttr {
        TargetName: string("share")
        Exact: bool(false)
        External: bool(false)
        Private: bool(false)
        ReadOnly: bool(false)
        Remove: bool(false)
      }
    origin: string("<redacted>/.local/share/chezmoi/home/dot_local/share")
    sourceRelPath: chezmoi.SourceRelPath {
        relPath: chezmoi.RelPath {
            relPath: string("dot_local/share")
          }
        isDir: bool(true)
      }
    targetStateEntry: *chezmoi.TargetStateDir
        chezmoi.TargetStateDir {
          perm: uint32(493)
          sourceAttr: chezmoi.SourceAttr {
              Condition: string("")
              Encrypted: bool(false)
              External: bool(false)
              Template: bool(false)
            }
        }
  }
[DEBUG] Checking path '.local/share' (*chezmoi.SourceStateDir): *chezmoi.SourceStateDir
  chezmoi.SourceStateDir {
    Attr: chezmoi.DirAttr {
        TargetName: string("")
        Exact: bool(false)
        External: bool(false)
        Private: bool(false)
        ReadOnly: bool(false)
        Remove: bool(false)
      }
    origin: *chezmoi.External
        chezmoi.External {
          Type: string("git-repo")
          Encrypted: bool(false)
          Exact: bool(false)
          Executable: bool(false)
          Private: bool(false)
          ReadOnly: bool(false)
          Checksum: chezmoi.externalChecksum {
              MD5: chezmoi.HexBytes[]
              RIPEMD160: chezmoi.HexBytes[]
              SHA1: chezmoi.HexBytes[]
              SHA256: chezmoi.HexBytes[]
              SHA384: chezmoi.HexBytes[]
              SHA512: chezmoi.HexBytes[]
              Size: int(0)
            }
          Clone: chezmoi.externalClone {
              Args: []string[
                  [0]: string("-b")
                  [1]: string("master")
                ]
            }
          Decompress: string("")
          Exclude: []string[]
          Filter: chezmoi.externalFilter {
              Command: string("")
              Args: []string[]
            }
          Format: string("")
          Archive: chezmoi.externalArchive {
              ExtractAppleDoubleFiles: bool(false)
            }
          Include: []string[]
          ArchivePath: string("")
          Pull: chezmoi.externalPull {
              Args: []string[]
            }
          RefreshPeriod: int64(604800000000000)
          StripComponents: int(0)
          URL: string("https://github.com/zap-zsh/zap.git")
          URLs: []string[]
          sourceAbsPath: string("<redacted>/.local/share/chezmoi/home/.chezmoiexternal.toml")
        }
    sourceRelPath: chezmoi.SourceRelPath {
        relPath: chezmoi.RelPath {
            relPath: string("dot_local/share")
          }
        isDir: bool(false)
      }
    targetStateEntry: *chezmoi.TargetStateDir
        chezmoi.TargetStateDir {
          perm: uint32(493)
          sourceAttr: chezmoi.SourceAttr {
              Condition: string("")
              Encrypted: bool(false)
              External: bool(false)
              Template: bool(false)
            }
        }
  }
  ```
  
I don't have enough understanding still of the overall intentions here, but this seems to be the expected behavior (to have the `.local/share` appearing twice), as one is tracking the local file and the other is tracking the external repo, having different `origin` fields on the respective nodes for the directory.

I don't think working around that in the tree creation to be a very feasible approach, as it could affect behavior in other regions, but I might be wrong on that.

What do you think? Should I try updating the tree creation / iterator to handle this scenario, or keep an afterwards de-duplication approach?

twpayne:
Thank you very much for your further investigative work here! This is excellent.



> I don't have enough understanding still of the overall intentions here, but this seems to be the expected behavior (to have the `.local/share` appearing twice), as one is tracking the local file and the other is tracking the external repo, having different `origin` fields on the respective nodes for the directory.

You're right that this means that the same directory has multiple origins, but I think chezmoi should only have a single directory represented in its source state. Maybe chezmoi should use a set of strings of each entry's origin, rather than the single string that chezmoi uses at the moment.



> I don't think working around that in the tree creation to be a very feasible approach, as it could affect behavior in other regions, but I might be wrong on that.

chezmoi has a large test suite, so please be bold and try big changes. The only weakness that I know of in chezmoi's test suite is interactive functionality (where chezmoi prompts for input from the user), which is not relevant to the issue here.



> What do you think? Should I try updating the tree creation / iterator to handle this scenario, or keep an afterwards de-duplication approach?

Please do try updating the tree creation. It's been a while since I worked on this part of the code, and this goes deep into [chezmoi's internals](https://pkg.go.dev/chezmoi.io/chezmoi/internal/chezmoi), but I think it's worth making the proper fix.

I'll make some changes to chezmoi's code to make it more discoverable. The short term pointer is that it will probably require changing https://pkg.go.dev/chezmoi.io/chezmoi/internal/chezmoi#SourceStateOrigin to return a slice of strings, and populating that slice correctly, maybe somewhere in https://github.com/twpayne/chezmoi/blob/2c439175be2c64132a3d36bf60b2d2947a947aca/internal/chezmoi/sourcestatetreenode.go.

I'm very happy to work with you on this bug so please don't hesitate to ask questions!

twpayne:
I created #4686 to improved chezmoi's internal docs.

twpayne:
As the PR to fix the duplicate occurrences will look quite different from this one, I'll close this PR for now. Please open a new PR for the fix outlined in https://github.com/twpayne/chezmoi/pull/4683#issuecomment-3353975182.

Issue #4730, opened 2025-10-23
Docs in "User guide -> tools -> merge -> use vscode as the merge tool" don't work for paths with spaces in them


Lorite:
## Describe the bug

- I tried adding the configuration for the merge command defined in the docs in "User guide -> tools -> merge -> use vscode as the merge tool".
- When the file path has spaces in it, VS Code doesn't open the merge editor correctly.

## To reproduce

- Modify your chezmoi config file (toml or yaml) with the suggestions from the docs.
- Run the `chezmoi merge <file>` command using a file path with spaces in it.

## Expected behavior

- The file diff should open on VS Code.

## Output of command with the `--verbose` flag

```console
$ chezmoi --verbose $COMMAND
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
RESULT   CHECK                       MESSAGE
ok       version                     v2.66.1, commit 94b95eb0b6bf91e48dd7b584a80ad60d44d8de4d, built at 2025-10-19T22:03:21Z, built by goreleaser
ok       latest-version              v2.66.1
ok       os-arch                     linux/amd64 (Ubuntu 25.04 (Plucky Puffin))
ok       uname                       Linux lori-ThinkPad-P15-Gen-2i 6.14.0-34-generic #34-Ubuntu SMP PREEMPT_DYNAMIC Wed Sep 17 09:21:29 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
ok       go-version                  go1.25.3 (gc)
ok       executable                  /snap/chezmoi/1105/chezmoi
ok       upgrade-method              snap-refresh
ok       config-file                 found ~/.config/chezmoi/chezmoi.toml, last modified 2025-10-23T11:18:01+02:00
ok       source-dir                  ~/.local/share/chezmoi is a git working tree (clean)
ok       suspicious-entries          no suspicious entries
ok       working-tree                ~/.local/share/chezmoi is a git working tree (clean)
ok       dest-dir                    ~ is a directory
error    hardlink                    failed creating hardlink from ~/.local/share/chezmoi to /tmp: link ~/.local/share/chezmoi/.chezmoi-doctor-hardlink-test /tmp/chezmoi-doctor3880247687/.chezmoi-doctor-hardlink-test: invalid cross-device link
ok       umask                       002
ok       cd-command                  found /usr/bin/zsh
ok       cd-args                     /usr/bin/zsh
ok       diff-command                found /usr/bin/code-insiders
[...]

PR #4731, opened 2025-10-23
fix: Docs in "User guide -> tools -> merge -> use vscode as the merge tool" don't work for paths with spaces in them

Lorite:
Solves #4730

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/tools/merge.md
+++ b/assets/chezmoi.io/docs/user-guide/tools/merge.md
@@ -67,7 +67,7 @@ To use [VSCode][vscode] as the merge tool, add the following to your config:
     command = "bash"
     args = [
         "-c",
-        "cp {{ .Target }} {{ .Target }}.base && code --new-window --wait --merge {{ .Destination }} {{ .Target }} {{ .Target }}.base {{ .Source }}",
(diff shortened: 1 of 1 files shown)


assets/chezmoi.io/docs/user-guide/tools/merge.md
```diff
@@ -67,7 +67,7 @@ To use [VSCode][vscode] as the merge tool, add the following to your config:
     command = "bash"
     args = [
         "-c",
-        "cp {{ .Target }} {{ .Target }}.base && code --new-window --wait --merge {{ .Destination }} {{ .Target }} {{ .Target }}.base {{ .Source }}",
+        "cp \"{{ .Target }}\" \"{{ .Target }}.base\" && code --new-window --wait --merge \"{{ .Destination }}\" \"{{ .Target }}\" \"{{ .Target }}.base\" \"{{ .Source }}\"",
```
twpayne: Thanks, for this. However it's better to use the `quote` template function, e.g. `{{ .Target | quote }}`, than adding quotes manually at each end.

halostatue:
Your commit will need to be updated; the title should be "fix: Docs…". Note the lowercase value for the change type.

twpayne:
Thanks for this. I enabled the `quote` template function in arguments and updated the doc in #4748, with you as co-author.

PR #4749, opened 2025-11-02
feat: Switch to znkr.io/diff for builtin diff

znkr:
This is a redo of 4db365b (or a revert of 7beeeac) to switch from github.com/sergi/go-diff to znkr.io/diff.

I am the author of znkr.io/diff and I saw that chezmoi was using it and then stopped using it and wondered what I could learn from this. Unfortunately, I don't really know if znkr.io/diff should provide a different output format now, or if go-git should fix its handling of edits with the same operation. For now, I adapted the format that go-git seems to need inside Chezmoi and left a comment.

Note that I only tested this with the existing unit tests.


znkr:
There was a a failing check about the commit message that I just fixed (before I stop sitting in front of a screen).

twpayne:
Firstly, thank you very much for both znkr.io/diff and for this PR! The Go ecosystem needs a modern, well-maintained diff library and znkr.io/diff is clearly this future.

There were two reasons why I reverted the original switch to znkr.io/diff:

1. I was using znkr.io/diff incorrectly in #4641. Specifically, I missed the use of znkr.io/diff/textdiff for text differences, and didn't spot the undocumented behavior of go-git that you address in this PR. Your PR resolves everything here.

2. By default, chezmoi includes colored diffs if they are supported by the terminal. I added support for colored diffs to go-git in https://github.com/go-git/go-git/pull/40. I didn't find a way to generate colored diffs with znkr.io/diff.

Apologies for not raising these issues in the znkr.io/diff repo. They've been on my to-do list for a while, but you were faster :)

znkr:
> Firstly, thank you very much for both znkr.io/diff and for this PR! The Go ecosystem needs a modern, well-maintained diff library and znkr.io/diff is clearly this future.

I am very happy to hear that. I did have the same wish for a modern, well-maintained diff and I am glad that you think znkr.io/diff is going in the right direction. ❤️ 

> I was using znkr.io/diff incorrectly in https://github.com/twpayne/chezmoi/pull/4641. Specifically, I missed the use of znkr.io/diff/textdiff for text differences, and didn't spot the undocumented behavior of go-git that you address in this PR. Your PR resolves everything here.

It's probably worth calling znkr.io/diff/textdiff out in the package documentation of znkr.io/diff: https://github.com/znkr/diff/pull/13#issue-3583504392. If you don't mind, I'd like to get your feedback.

> By default, chezmoi includes colored diffs if they are supported by the terminal. I added support for colored diffs to go-git in https://github.com/go-git/go-git/pull/40. I didn't find a way to generate colored diffs with znkr.io/diff.

Yeah, I decided against colored output initially to keep the scope small. But I believe there are good reasons to include it: https://github.com/znkr/diff/issues/14

> Apologies for not raising these issues in the znkr.io/diff repo. They've been on my to-do list for a while, but you were faster :)

No worries, but please don't be shy with issues like that. I really want feedback like this :)

Edit: I went ahead and implemented basic coloring support in znkr.io/diff. It's not a v1 yet and there's still time to change it, but generally I think the API for coloring should be okay.

twpayne:
Thank you for adding coloring support to znkr.io/diff! Would you like to update this PR to use it? If you don't have time, I'll happily do the work. No rush.

znkr:
I considered changing this PR to use the newly added coloring support, but it requires a larger refactoring and I don't have the time for that unfortunately.

twpayne:
It turns out that color was already supported, as the color is added by go-git when it processes the hunks. I've created #4771 that is based on this PR and includes a bump of znkr.io/diff to the latest version.

twpayne:
Closing in favor of #4771.

PR #4768, opened 2025-11-20
chore: Update dependencies

twpayne:


twpayne:
The failures on macOS seem to be due to a bug in the GitHub Action Runners. I've opened https://github.com/actions/runner/issues/4134.

Issue #2668, opened 2023-01-02
Replace github.com/masterminds/sprig template functions


twpayne:
## Is your feature request related to a problem? Please describe.

`github.com/masterminds/sprig`, used by chezmoi,  has unfortunately become a popular library of template functions despite several critical issues:

* It is unmaintained.
* The argument order of its functions is incompatible with Go template's pipelining (see #2215).
* Its naming convention does not match Go's (e.g. `toJson` should be `toJSON`).

## Describe the solution you'd like

chezmoi's template functions should be up-to-date, compatible with Go template pipelining, and match Go's naming conventions.

## Describe alternatives you've considered

Keeping minorminds/sprig template functions as-is.

## Additional context

A smooth transition should be provided, with both old and replacement template functions being available to users for some time. Users can use `.chezmoiversion` to help the transition.

bradenhilton:
Perhaps we could also take the opportunity to implement a higher-order function (which would add to the `funcMap`, not something done at runtime) similar [to the one in this issue](https://github.com/Masterminds/sprig/issues/91)?

As you know, there is the [`quoteList`](https://www.chezmoi.io/reference/templates/functions/quoteList/) template function, which was implemented explicitly. So far, there hasn't been much of a demand sufficient to add similar functions, which would probably be necessary to add an explicit function (and thus create clutter and perhaps some complexity).

An `each` function separates the iteration logic from the action being taken on the item in the list. Going forward, implementing `stringFuncEach` would likely be as simple as implementing `stringFunc` and then setting `funcMap["stringFuncEach"]` to something like `eachFunc(stringFunc)`.

The `each{Func}` functions could be as generic as possible or type specific if needed.

Maybe this could even be done dynamically with reflection? So we could do something like:

1. Group functions by type if needed
2. Iterate through each function/group
3. Get function name (with reflection?)
4. `funcMap[funcName + "Each"] = eachFunc{Type}(func)` (probably not syntactically correct)

Separating the logic would remove the need for explicit functions in all but the most specialized cases, which would remove clutter and be trivial to extend as new functions are added. 

The alternative, which would be better, but also likely increase complexity, would be to add iteration logic to every requisite function, so that you can pass it a single item or a list of items and get the expected output, e.g.

```
$string | quote
"foobarbaz"


$list | quote
["foo", "bar", "baz"]
```

This approach *would* be easier if we had control over how each function is implemented, as sprig is essentially a black box in this case.

[...]

PR #4814, opened 2025-12-07
chore: Switch to github.com/go-sprout/sprout for internal tools

twpayne:
Refs #2668.

This should be fully backwards-compatible.

Draft for now.

### DIFF
--- a/internal/cmd/testdata/scripts/sprig.txtar
+++ b/internal/cmd/testdata/scripts/sprig.txtar
@@ -1,0 +1,8 @@
+# test that sprig template functions work as expected
+# see https://docs.atom.codes/sprout/migration-from-sprig
+exec chezmoi execute-template '{{ b64dec "." }}'
+stdout '^illegal base64 data at input byte 0$'
+exec chezmoi execute-template '{{ dict "a" (dict "b" 2 ) | dig "a" "b" }}'
+stdout '^map\[b:2]$'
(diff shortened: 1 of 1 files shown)


marmitar:
I decided to test it and the biggest change I felt is with [`dig`](https://docs.atom.codes/sprout/migration-from-sprig#dig). In sprig you had `dig ...PATH DEFAULT DICT` and `DEFAULT` was required, but in sprout it's `dig ...PATH DICT` which returns `nil` if missing.

I think sprout's implementation is more natural, since I actually have forgotten about the default quite a few times before, writing something like `dig ...PATH DICT | default DEFAULT`. Anyway, just wanted to let you know about this change.

twpayne:
Thank you very much for the testing! This PR needs more work, specifically:

* Currently this PR uses sprout's `sprigin` package, which should provide 100% compatibility with sprig. Therefore, there should be no change to the documentation in this PR.
* To migrate to sprout properly, users should be able to switch from sprig to sprout in their templates at their own pace. I'm not sure if this should be configured globally or on a per-file basis.

I'll update the PR to migrate to `sprigin` internally, which should have no user-visible changes.

twpayne:
Thanks again for the testing. This PR now only touches internal tools and adds some tests for sprig's behavior. Actually migrating to sprout will require more work, as described in https://github.com/twpayne/chezmoi/issues/2668#issuecomment-3618899286.

PR #4822, opened 2025-12-11
feat: Add unmanaged.ignore configuration variable

### LINKED ISSUE #4813: Feature request: `.chezmoiunmanagedignore` file to filter `chezmoi unmanaged` output
### Is your feature request related to a problem?

When using `chezmoi unmanaged` to find dotfiles that should be added to the source state, the output includes many paths that are intentionally not managed:

- XDG data/state directories (`.local/share/`, `.local/state/`)
- Tool-managed environments (`.local/pipx/`, virtual environments)
- Application runtime state (`.config/gcloud/`, `.config/gh/`)
- Downloaded plugins/assets (`Spoons/`, `node_modules/`)

These clutter the output and make it hard to spot files that actually need attention. Currently, there's no way to filter `chezmoi unmanaged` output without external tooling.

### Describe the solution you'd like

A `.chezmoiunmanagedignore` file in the source directory that filters the output of `chezmoi unmanaged`. It would use the same gitignore-style patterns as `.chezmoiignore`.

Example `.chezmoiunmanagedignore`:
```
# XDG data directories (not config)
.local/share/
.local/state/

# Tool environments
.local/pipx/

# Runtime state
.config/gcloud/
.config/gh/

# Nested logs everywhere
**/logs/
```

Running `chezmoi unmanaged` would then exclude these paths from output.

### Describe alternatives you've considered

1. **Using `.chezmoiignore`** - This prevents files from being *applied*, not from appearing in `unmanaged` output. Different purpose.

2. **Global gitignore + external filtering** - I implemented a wrapper that filters through both global gitignore and a custom ignore file. Works but requires shell-specific tooling.

3. **Piping through grep** - Manual and error-prone for complex patterns.

### Additional context

I've implemented this as a fish shell wrapper that:
1. Calls `chezmoi unmanaged`
2. Filters output against a `.chezmoiunmanagedignore` file using gitignore patterns

This works well but feels like functionality that belongs in chezmoi itself, especially since it mirrors the existing `.chezmoiignore` pattern.

Related: #3608 (about empty directories in unmanaged output - different but complementary)

twpayne:
Fixes #4813.

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -386,6 +386,10 @@ sections:
     command:
       default: '`vault`'
       description: Vault CLI command.
+  unmanaged:
+    ignore:
+      type: '[]string'
+      description: Patterns to ignore.
   update:
     apply:
       type: bool
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -185,6 +185,7 @@ type ConfigFile struct {
 	Git        gitCmdConfig        `json:"git"        mapstructure:"git"        yaml:"git"`
 	Merge      mergeCmdConfig      `json:"merge"      mapstructure:"merge"      yaml:"merge"`
 	Status     statusCmdConfig     `json:"status"     mapstructure:"status"     yaml:"status"`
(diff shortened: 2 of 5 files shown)


twpayne:
This PR is not needed. See https://github.com/twpayne/chezmoi/issues/4813#issuecomment-3640038346.

PR #4837, opened 2025-12-21
feat: Add support for transcrypt and git-crypt

### LINKED ISSUE #4836: Support integration with transparent git encryption tools

I use the transparent git encryption tool [transcrypt](https://github.com/elasticdog/transcrypt), and it works absolutely fine with Chezmoi - it would just be great if there could be greater integration with `chezmoi add --encrypt`.

Using transcrypt, I could set up my `.gitattributes` file to encrypt all files that have the prefix `encrypted_`. Therefore, an option could be to add some `encryption = "none"` mode, that does no encryption but allows you to use `chezmoi add --encrypt` and create files with the `encrypted_` prefix.

Alternatively, https://github.com/twpayne/chezmoi/issues/3503 would still be really useful - it would allow me to write a script that, for example, added the exact file path of an added file to the `.gitattributes` file, and then do nothing when it comes to encrypting and decrypting the file itself.

twpayne:
Fixes #4836.

Draft for now as it needs documentation and more testing.

### DIFF
--- a/assets/chezmoi.io/docs/index.md.tmpl
+++ b/assets/chezmoi.io/docs/index.md.tmpl
@@ -18,7 +18,7 @@ including:
 * templates (to handle small differences between machines)
 * password manager support (to store your secrets securely)
 * importing files from archives (great for shell and editor plugins)
-* full file encryption (using gpg or age)
+* full file encryption (using age, gpg, git-crypt, or transcrypt)
 * running scripts (to handle everything else)
 
 With chezmoi, pronounced /ʃeɪ mwa/ (shay-mwa) (listen [1][listen1] [2][listen2]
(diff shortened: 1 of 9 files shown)


j-waters:
Thanks for looking into this so quickly! I've tested this branch locally and it works perfectly with my use-case. I've written down my testing methodology in the form of a tutorial if you'd like to include any of this in the documentation:

Initialise [transcrypt](https://github.com/elasticdog/transcrypt):

```bash
chezmoi cd
transcrypt
```

Add the following pattern to `.gitattributes`:

```
# .gitattributes
encrypted_*  filter=crypt diff=crypt merge=crypt
```

Add the file and commit it:
```bash
$ chezmoi add --encrypt ~/.config/sensitive_file

$ git add dot_config/sensitive_file

$ git commit -m "Add sensitive file"
```

Confirm that the file is indeed encrypted:
```bash
$ git ls-crypt
# dot_config/sensitive_file
```

Note that commands like `git show`, `git diff`, etc. will also show the cleartext form of the file.

Use `transcrypt --display` to show instructions for how to setup transcrypt after cloning the repository elsewhere. It will involve running a command like:

```bash
$ transcrypt -c aes-256-cbc -p '<your password>'
```

PR #4858, opened 2026-01-04
chore: use doc links to point at stdlib symbols

babakks:
Fixes #4856

Tested with a local build of Golangci-lint, cloned from latest `main`.

Unrelated to this, but seems like `golines` in my build of Golangci-lint complains about some long lines. Expand to see more. I can try fixing them as well as they seem to be all cosmetic changes. Let me know, anyway.

<details><summary>Details</summary>
<p>

```
internal/chezmoi/sourcestate.go:666:1: File is not properly formatted (golines)
                        if err := PersistentStateSet(persistentState, EntryStateBucket, sourceUpdate.destAbsPath.Bytes(), sourceUpdate.entryState); err != nil {
^
internal/cmd/githubtemplatefuncs.go:170:1: File is not properly formatted (golines)
        if err := chezmoi.PersistentStateSet(c.persistentState, gitHubVersionReleaseStateBucket, gitHubVersionReleaseKey, &gitHubLatestReleaseState{
^
internal/cmd/inittemplatefuncs_test.go:338:1: File is not properly formatted (golines)
                                assert.Equal(t, tc.expected, config.promptMultichoiceInteractiveTemplateFunc(tc.prompt, tc.multichoices, tc.args...))
^
internal/cmd/mergecmd.go:97:1: File is not properly formatted (golines)
        if targetStateEntry, err = sourceStateEntry.TargetStateEntry(c.destSystem, c.DestDirAbsPath.Join(targetRelPath)); err != nil {
^
4 issues:
* golines: 4
```

</p>
</details>

PR #4860, opened 2026-01-04
chore: fix `golines` issues

babakks:
This PR fixes the issues reported by `golines` (See #4858 description). The linter is not yet released (as part of Golangci-lint), though.

The changes are the result of running `golangci-lint fix` (built from latest `main`).


twpayne:
Thank you!

Issue #4223, opened 2025-01-24
chezmoi adds too many files


itsaphel:
## What exactly are you trying to do?

I want to slowly start using chezmoi, eg initially using it for `~/.config/nvim/` only

## What have you tried so far?

I used the idea at the `oh-my-zsh` example at: https://www.chezmoi.io/reference/commands/add/#-i-include-types:

`chezmoi add ~/.config/nvim/ --exact --recursive`

When I do that, chezmoi starts trying to manage my entire `~/.config` it seems:

```
> chezmoi managed
.config
.config/.DS_Store
.config/chezmoi
.config/fish
...
.config/zed
```

Consequently, when I do `chezmoi diff`, it wants to replace all of my files in other directories (aside from `~/.config/nvim`) with null.

## Additional context

I suspect I'm doing something obviously wrong above, as it seems like a simple use case.

The issue doesn't happen if I don't use `--exact`, but presumably I do want that flag.

May I suggest more examples in the quick start, and the add command, of adding entire directories to be managed with chezmoi? I'd think this is as common, if not more common, than adding individual files like `~/.zshrc`. But https://www.chezmoi.io/quick-start has no examples of whole directories, and the `add` command reference only has one (`~/.oh-my-zsh`)

halostatue:
There may be a bug in how `--exact` interacts with `--recursive` (I’ll leave that to @twpayne), but I'm not sure why you'd want to use `--exact`. What I would check is that you now have `exact_dot_config/exact_nvim` etc rather than `dot_config/exact_nvim`. However, the `exact_` attribute tells chezmoi to remove anything _not managed_ by chezmoi on `chezmoi apply`.

twpayne:
What I think is happening here is:

The initial state is that there are no files currently managed by chezmoi.

When you run

    chezmoi add ~/.config/nvim/ --exact --recursive

you're asking to chezmoi to manage `~/.config/nvim`, but to manage `~/.config/nvim`, chezmoi also has to manage the `~/.config` directory.

Since you specified the `--exact` flag, chezmoi also applies this to the `~/.config` directory, so it has to create entries for every existing subdirectory of `~/.config`, which is why you see so much in the output of `chezmoi managed`.

itsaphel:
I think that makes sense. 

> but to manage `~/.config/nvim`, chezmoi also has to manage the `~/.config` directory.

Why does it need to manage `~/.config` to manage `~/.config/nvim`? (is this true even when not using `--exact`?)

[...]

PR #4892, opened 2026-01-21
fix/adding exact sub-directory adds exact parent directory

aaldebs99:
Hello again!

I recently noticed what I assume is a a weird logic bug. In a nutshell, when adding exact sub-directories, if the parent isn't already created and doesn't exist in the source directory, that parent gets added as an exact directory when it shouldn't be.

I'm creating this draft PR to make sure this logic aligns with what the maintainers believe should happen in this case. I created a quick test case for the failed logic. Naturally, this new test currently fails.

Please let me know if you agree that this should be fixed and I'll fix it!

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -340,21 +340,43 @@ type ReplaceFunc func(targetRelPath RelPath, newSourceStateEntry, oldSourceState
 
 // AddOptions are options to SourceState.Add.
 type AddOptions struct {
-	AutoTemplate      bool                 // Automatically create templates, if possible.
-	Create            bool                 // Add create_ entries instead of normal entries.
-	Encrypt           bool                 // Encrypt files.
-	EncryptedSuffix   string               // Suffix for encrypted files.
-	Errorf            func(string, ...any) // Function to print errors.
-	Exact             bool                 // Add the exact_ attribute to added directories.
-	Filter            *EntryTypeFilter     // Entry type filter.
-	OnIgnoreFunc      func(RelPath)        // Function to call when a target is ignored.
-	PreAddFunc        PreAddFunc           // Function to be called before a source entry is added.
-	ConfigFileAbsPath AbsPath              // Path to config file.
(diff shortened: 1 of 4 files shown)


halostatue:
I can't say what the _correct_ behaviour related to parent directories should be, but I did extensive research on this: https://github.com/twpayne/chezmoi/issues/4223#issuecomment-2614574590 and subsequent.

aaldebs99:
> I can't say what the _correct_ behaviour related to parent directories should be, but I did extensive research on this: [#4223 (comment)](https://github.com/twpayne/chezmoi/issues/4223#issuecomment-2614574590) and subsequent.

Gotcha, I see, it definitely is a weird edge case....
In my experience, I would have preferred for it to NOT add the parent as an exact when adding a subdirectory as exact.

I guess how do you usually decide weird edge cases like this?
Would you accept this change if I put it together?

halostatue:
We don't have a standard answer to this.

My personal rule of thumb is whether the behaviour is easy to explain relative to the complexity it's trying to implement. Right now, I find `exact_` hard to explain and don't use it.

I think that this is slightly easier to explain, but I'm not sure it makes the code for state management easy.

aaldebs99:
Welp I'm not entirely sure how that decision is going to be made, but in my opinion, this is how I would expect it to behave.

I would love to see this merged but I totally understand if you decide to not follow this approach

twpayne:
Thank you for this PR! I haven't had time to review it yet, but I assure you that it is not forgotten. I particularly like that you've added new tests for the new behavior and the existing tests continue to pass.

twpayne:
Thank you, this is excellent.

So I can merge, please can you squash all the commits into a single commit with the commit message:

    fix: Adding exact subdir won't add exact parent

aaldebs99:
> Thank you, this is excellent.
> 
> So I can merge, please can you squash all the commits into a single commit with the commit message:
> 
> ```
> fix: Adding exact subdir won't add exact parent
> ```

Happy to help!
Done!

PR #4893, opened 2026-01-22
Feat/ssh agent encryption

thekoma:
## Motivation

To use `chezmoi` effectively on multiple machines, one inevitably needs to manage encryption keys. Using GPG or age requires copying private key files to every machine, which increases the attack surface.

This PR introduces a new encryption backend: `ssh`.
It leverages the SSH Agent protocol to perform decryption. This enables a **Zero-Trust** workflow via **SSH Agent Forwarding**: you can decrypt your dotfiles on a remote server without your private key ever leaving your local machine.

## Implementation Details

It utilizes a "Sign-Challenge KDF" scheme:
1.  **Encryption**: A random 32-byte challenge and salt are generated. The `ssh-agent` signs the challenge using the configured identity. `scrypt` derives a symmetric key from the signature to encrypt the file (using `chacha20poly1305`).
2.  **Decryption**: The challenge is read and sent to the agent for signing. If the agent holds the correct private key, the signature matches, enabling key derivation.

### Security Hardening
- **Domain Separation**: All challenges are prefixed with `chezmoi-ssh-signature-v1:` before signing. This prevents "signing oracle" attacks where an attacker might trick the agent into signing a valid authentication payload for another protocol.
- **Algorithm Enforcement**: The backend strictly enforces allowed signature algorithms (Ed25519, RSA-SHA2, ECDSA) and rejects weak ones (like SHA-1/ssh-rsa).

## Configuration

In `chezmoi.toml`:
```toml
encryption = "ssh"
[ssh]
    identity = "ssh-ed25519 AAA..." # Output from ssh-add -L
```    
    
--Edited to add Security Hardening

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
+++ b/assets/chezmoi.io/docs/reference/configuration-file/variables.md.yaml
@@ -369,6 +369,9 @@ sections:
       description: Extra args to secret CLI command.
     command:
       description: Generic secret CLI command.
+  ssh:
+    identity:
+      description: SSH public key identity.
   status:
     exclude:
       type: '[]string'
--- a/assets/chezmoi.io/docs/user-guide/encryption/index.md
+++ b/assets/chezmoi.io/docs/user-guide/encryption/index.md
@@ -1,7 +1,7 @@
 # Encryption
 
 chezmoi supports encrypting files with [age][age], [git-crypt][gitcrypt],
-[gpg][gpg], and [transcrypt][transcrypt].
+[gpg][gpg], [ssh][ssh], and [transcrypt][transcrypt].
 
 Encrypted files are stored in ASCII-armored format in the source directory with
 the `encrypted_` attribute and are automatically decrypted when needed.
@@ -18,4 +18,5 @@ re-encrypt it afterwards.
 [age]: https://age-encryption.org
 [gitcrypt]: https://github.com/AGWA/git-crypt
 [gpg]: https://www.gnupg.com/
+[ssh]: ssh.md
 [transcrypt]: https://github.com/elasticdog/transcrypt
--- a/assets/chezmoi.io/docs/user-guide/encryption/ssh.md
+++ b/assets/chezmoi.io/docs/user-guide/encryption/ssh.md
@@ -1,0 +1,35 @@
+# SSH
+
+chezmoi supports encrypting files using SSH keys via `ssh-agent`. This allows you to use your existing SSH identity for encryption without needing GPG or Age keys.
+
+## Configuration
+
+Specify `ssh` encryption in your configuration file, and provide the public key identity you wish to use:
+
+```toml title="~/.config/chezmoi/chezmoi.toml"
+encryption = "ssh"
+[ssh]
+    identity = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAA..."
+```
+
+You can obtain your public key string by running:
+
+```console
+$ ssh-add -L
+```
+
+## How it works
+
+chezmoi uses a "Challenge-Response" mechanism to derive a secure encryption key from your SSH key:
+
+1.  **Encryption**: chezmoi generates a random challenge. Your `ssh-agent` signs this challenge. The signature is used to derive the encryption key (using scrypt).
+2.  **Decryption**: chezmoi asks the agent to sign the challenge again. If the correct key is loaded, the signature matches, and the key is derived.
+
+## Portability
+
+*   **Agent Forwarding**: You can decrypt files on a remote server without copying your private key by using SSH Agent Forwarding (`ssh -A`).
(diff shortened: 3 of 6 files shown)


twpayne:
This is an interesting idea, but please follow [talk, then code](https://dave.cheney.net/2019/02/18/talk-then-code) before opening large PRs, as described in the [contributing guide](https://www.chezmoi.io/developer-guide/contributing-changes/).

Although I can see that this is a potentially convenient feature, I have a number of objections:
1. This PR re-uses signing keys for encryption, which is generally considered bad security practice.
2. [age](https://age-encryption.org/) already supports encrypting with SSH keys as a convenience, but explicitly recommends against using it, which is why chezmoi does not implement it.
3. There is no easy mechanism for rotating keys. If you change your SSH key, you also have to re-encrypt all your encrypted files.
4. I am not a cryptography expert and am not able to review this code for security correctness.

I'm happy to hear counter-arguments to the above, but especially given point 2, I would chose age's support for encryption with SSH keys over what is implemented in this PR.

twpayne:
BTW, the PR description looks like it was generated by an LLM. Did you use an LLM when creating this PR?

thekoma:
Absolutely, English is not my first language so to deliver the message in the best way possible I feeded it to Gemini to summarize and translate it.

I also used the LLM to scan the style of coding you used in the project and to build the tests to adhere to the project as much as possible. 

It does make sense to me to use modern tools to enhance open source projects adding features that I crave. 
Sorry regarding the noise, I decided to consolidate my previous replies into this single comment (and I'll remove the others).

First, apologies for skipping the "talk then code" phase. Since 90% of the PR is a new isolated module and only touches ~3 lines of core code, I thought a working PoC would demonstrate the value better than just ranting about the idea in an issue!

Regarding your technical objections:

**Security (Signing Reuse)**: You are absolutely right, I overlooked the risk of creating a signing oracle. I have a fix ready for this: I will maximize security by implementing domain separation:
prepending a unique constant prefix (e.g., chezmoi-ssh-signature-v1:) to the challenge before signing. 
This ensures that signatures generated by chezmoi are namespaced and can never be accepted by an SSH server for authentication.

**Why not Age?**: This is the "killer feature" validation for me. 
age supports SSH keys on disk, but not the agent. 
My specific use case is SSH Agent Forwarding. 
I want to be able to `ssh -A` into a server and use chezmoi there without ever copying my private key file to that machine. 
This backend allows exactly that: securely decrypting via the forwarded socket. 
Age cannot do this currently.

**Key Rotation**: It is a valid concern, but I feel like it's a feature that belongs effectively at the application level (e.g. a command to decrypt-reencrypt with a new key) rather than blocking the encryption method itself.
I'm polishing the fix for point 1 and will push it shortly.

thekoma:
The point of using the ssh-agent is to not put the private part of the key on every machine but instead keep it in something like bitwarden desktop app that can act as ssh-agent in a simple and seamless way in your primary machine. 

This reduces the attack surface greatly. 

You can have multiple keys and the configuration permits to define a specific key to be used as encryption. This means that you can use a key just for chezmoi. 

The additional security to prevent rainbow tables is implemented by randomic salt and I've set the check to accept only well known strong cyphers. 

I understand that there are better ways to implement Encryption via age or gpg but this way looked enough strong to me at least if used in conjunction with the password manager integrations. 

I'm not a security expert, tho so if this looks too flimsy I will gladly close the PR.

thekoma:
Added the fix to point 1 plus a couple of error message and cleanup here and there.
Obviously the encrypted files by the previous version are not compatible to this one because of the additional signature prefix.

`sshEncryptionSignaturePrefix = "chezmoi-ssh-signature-v1:"`


**Edit:**
Uh by the way the Security concern is tickling my ears:
To enhance the hardening i could:
- Limit to Ed25519 the supported algorithm it will protect against non deterministic signatures
- Remove the fingerprinting of the encrypted file. I added it to auto select the correct ssh key from the agent during the decrypt but it will leak which private key has been used.
- I could memory wipe as soon as the decryption happens so that it won't stay in RAM waiting for Go garbage collector to do his work.
- I could move to Argon2id instead of Scrypt. I don't actually know how the library works at the moment. But i know that is the more hip and hardened solution out there. (Yes i know I've been lazy).


Let me know :)

twpayne:
This PR feels like it was vibe coded. Can you confirm that you didn't use any LLMs when you wrote this code?

thekoma:
@twpayne, this has clearly been used. If you feel the quality isn't up to standard or there's an error in the code, please let me know, and I'll gladly rewrite it.


Hey, this is your repo! I won't go down a rabbit hole trying to convince you to accept a PR you don't feel is right to accept.

twpayne:
When contributing to a project, make sure you read the [contributing guide](https://www.chezmoi.io/developer-guide/) before starting work. chezmoi does not allow LLM-coded contributions.

PR #4905, opened 2026-02-01
feat: Prefer PowerShell Core (pwsh) for .ps1 scripts

### LINKED ISSUE #4888: Default to PowerShell Core (pwsh) for .ps1 Sripts on Windows when Available

When using `run_once_` or `run_ scripts` with a `.ps1` extension on Windows, _chezmoi_ executes them with Windows PowerShell (`powershell.exe`) rather than PowerShell Core (`pwsh.exe`).

Windows PowerShell has poor UTF-8 support, causing scripts with Unicode characters to fail with misleading parse errors:

```posh
# Script containing:
Write-Host "✓ Task completed"

# Fails with:
Missing closing '}' in statement block or type definition.
```

This is annoying and difficult to find root cause because the script is syntactically valid - it is  purely an encoding/parsing issue in the legacy shell.

On Windows, when PowerShell Core (`pwsh.exe`) is available in `PATH`, _chezmoi_ should prefer it over Windows PowerShell for executing `.ps1` scripts.

The desired behaviour would be:

- Default to PowerShell Core (`pwsh.exe`) if available
- Transparently fall back to Windows PowerShell  (`powershell.exe`) 
- Where both PowerShell Core and Windows PowerShell are available the new default could be overridden through 
[Interpreters.ps1](https://github.com/search?q=repo%3Atwpayne%2Fchezmoi%20interpreters.ps1&type=code) if desired

While Windows PowerShell is still shipped as the default with Windows:

- Windows PowerShell 5.1 is in maintenance mode; `pwsh` is the actively developed successor
- `pwsh` handles UTF-8 correctly out of the box
- `pwsh` is cross-platform, so scripts work identically on Windows/Linux/macOS
- Users who explicitly want legacy PowerShell could still configure the  use of `powershell.exe` if required

**Current workaround**: Explicitly configure the interpreter in `.chezmoi.toml`:

```posh
[interpreters.ps1]
    command = "pwsh"
    args = ["-NoLogo", "-NoProfile"]
```

This works but requires users to discover the issue first, and requires the boilerplate to address.

**Alternative**: Restrict scripts to ASCII-only output—but this feels like a step backward.

**Additional context**
- 
- Windows PowerShell ships with Windows but is no longer actively developed
- PowerShell Core (`pwsh`) is the recommended PowerShell for new development
- The encoding issue only manifests with  the legacy default
- Users on Linux/macOS with only `pwsh` installed don't encounter this issue

dgalbraith:
PowerShell Core (pwsh) is now the default interpreter for .ps1 scripts when available, with automatic fallback to Windows PowerShell.

## Benefits
- Better UTF-8 and Unicode handling
- Cross-platform support (Windows, Linux, macOS)
- Modern PowerShell features
- Backward compatible with Windows PowerShell fallback

## Implementation
- Added runtime detection for pwsh/powershell via findExecutable
- Windows: checks pwsh → powershell → empty
- Unix: checks pwsh → empty (no Windows PowerShell on Unix)
- Added unit tests for PowerShell interpreter selection (Windows and Unix)
- Added integration tests for .ps1 script execution via testscript
- Updated documentation to use pwsh.exe in examples
- Added notes about PowerShell Core installation and fallback options
  
## Fixes
- Fixed line ending normalization in scriptinterpreters_windows test

Users can override this behavior in their config if needed:

```posh
[interpreters.ps1]
    command = "powershell"  # explicitly use Windows PowerShell
```
Closes #4888

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/interpreters.md
+++ b/assets/chezmoi.io/docs/reference/configuration-file/interpreters.md
@@ -9,13 +9,13 @@ extensions. Other extensions require an interpreter, which must be in your
 
 The default script interpreters are:
 
-| Extension | Command      | Arguments |
-| --------- | ------------ | --------- |
-| `.nu`     | `nu`         | *none*    |
-| `.pl`     | `perl`       | *none*    |
-| `.py`     | `python3`    | *none*    |
-| `.ps1`    | `powershell` | `-NoLogo` |
-| `.rb`     | `ruby`       | *none*    |
+| Extension | Command      | Arguments       |
+| --------- | ------------ | --------------- |
+| `.nu`     | `nu`         | *none*          |
+| `.pl`     | `perl`       | *none*          |
+| `.py`     | `python3`    | *none*          |
+| `.ps1`    | `pwsh`       | `-NoLogo -File` |
+| `.rb`     | `ruby`       | *none*          |
 
 Script interpreters can be added or overridden by adding the corresponding
 extension (without the leading dot) as a key under the `interpreters`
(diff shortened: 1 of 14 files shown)


twpayne:
Thank you very much!

Issue #4920, opened 2026-02-11
Feature request: could `textconv` filters also affect comparison in `apply` and `status`?


owpac:
Hello! Sorry in advance for the detailed issue.

## Is your feature request related to a problem? Please describe.

> **Edit:** I initially thought `textconv` already affected the comparison logic in `chezmoi apply` — I had a flawed test setup that gave me a false positive. After further investigation, I realized `textconv` only affects diff display. The feature request below is based on that corrected understanding.

Many macOS apps constantly rewrite their plist preference files with volatile metadata (timestamps, window positions, update counters, UUIDs, etc.) even when the user hasn't changed any settings. This is the same root issue as #3650.

When managing these files with chezmoi, `chezmoi apply` always prompts to overwrite them because the raw bytes differ — even though the actual configuration hasn't changed. `chezmoi status` also reports them as `MM`.

Today, `textconv` filters only affect how diffs are **displayed** (in `chezmoi diff` and the diff shown during `chezmoi apply`). They do **not** affect the underlying comparison that determines whether a file needs to be applied. This means there is no way to tell chezmoi "these two files are equivalent for comparison purposes, even if their raw bytes differ."

## Describe the solution you'd like

It would be great if `textconv` output could be used as the basis for **comparison** in `chezmoi apply` and `chezmoi status`, not just for display in `chezmoi diff`.

The idea: if a file matches a `textconv` pattern, chezmoi would run the textconv command on both the source and destination contents, and compare the **filtered outputs**. If they are identical, the file would be considered up-to-date — no overwrite prompt in `apply`, no `MM` in `status`.

This would make the existing `textconv` feature much more powerful while keeping backwards compatibility (users who only use textconv for display would see no change in behavior, since if the raw bytes differ the filtered outputs will typically differ too).

Alternatively (or additionally), a simpler option could be a glob-based exclude list for `chezmoi status`:

```yaml
status:
  exclude:
    - pattern: "**/*.plist"
```

This would be less intrusive to implement but would only address `status`, not the `apply` prompt.

## Describe alternatives you've considered

The alternatives discussed in #3650 all have significant drawbacks for plist files:

- **`modify_` scripts**: require enumerating every config key to set via `plutil -replace`. For complex apps like iTerm2 (~1000 keys) or Raycast (~1400 lines), this is impractical. It also breaks `chezmoi add`/`re-add` since the managed entry becomes a script, not the file itself.
[...]

PR #4924, opened 2026-02-17
feat: Use textconv filters for comparison in apply, status, and verify

owpac:
> [!NOTE]
> This PR was authored with the assistance of AI (Claude). While it has been tested manually and with the existing test suite, please review the implementation carefully.

## Summary

Implements the feature requested in #4920: `textconv` filters now affect the **comparison logic** in `apply`, `status`, and `verify`, not just diff display.

- When textconv-filtered contents of source and destination match, the file is considered up-to-date (no overwrite prompt, no `MM` in status, verify passes)
- When textconv-filtered contents differ, the raw (unfiltered) source contents are written to the destination
- Existing behavior is unchanged for files that don't match any textconv pattern

This solves the long-standing pain point with macOS plist files (and similar config files on other platforms) that are constantly rewritten by apps with volatile metadata (timestamps, window positions, update counters, etc.).

### Implementation

The change threads `TextConvFunc` through chezmoi's comparison pipeline at two levels:

1. **EntryState level** (`SourceState.Apply()`): filters `ContentsSHA256` in-place on both target and actual entry states before the `PreApplyFunc` comparison. This drives `status` output, `apply` prompting, and `verify` results.
2. **`TargetStateFile.Apply()` level**: uses textconv for the SHA256 comparison that decides whether to write, then falls back to raw comparison if no textconv pattern matches.

The existing stale-cache logic (line 831 of `sourcestate.go`) naturally handles first-run migration when textconv is newly configured.

## Test plan

- [x] New test cases in `textconv.txtar`:
  - `chezmoi status` reports no changes when textconv-filtered contents match
  - `chezmoi verify` succeeds when textconv-filtered contents match
  - `chezmoi apply` does not overwrite when textconv-filtered contents match
  - `chezmoi diff` shows no output when textconv-filtered contents match
  - `chezmoi apply` writes raw (unfiltered) contents when textconv-filtered contents differ
- [x] Existing `textconv.txtar` tests still pass (diff display, interactive diffs)
- [x] Full `internal/chezmoi/...` test suite passes
- [x] Tested manually with real macOS plist files and a textconv script stripping volatile keys

### DIFF
--- a/assets/chezmoi.io/docs/reference/configuration-file/textconv.md
+++ b/assets/chezmoi.io/docs/reference/configuration-file/textconv.md
@@ -1,7 +1,14 @@
 # textconv
 
 A section called `textconv` in the configuration file controls how file contents
-are modified before being passed to diff.
+are modified before being compared and passed to diff.
+
+When configured, `textconv` filters affect comparison in `apply` (deciding
+whether to write), `status` (deciding whether to report changes), `verify`
+(deciding whether the destination matches), and `diff` (displaying differences).
+If `textconv`-filtered contents of the source and destination are identical, the
+file is considered up-to-date. If they differ, the raw (unfiltered) source
+contents are written to the destination.
 
 The `textconv` must contain an array of objects where each object has the
 following properties:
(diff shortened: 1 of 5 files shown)


twpayne:
For the record, this PR was closed without review and @owpac was permanently banned for directly violating [chezmoi's contributing guidelines](https://www.chezmoi.io/developer-guide/).

PR #4945, opened 2026-03-07
feat: Add podman as an alias for docker command

twpayne:
Fixes https://github.com/twpayne/chezmoi/commit/5b4c85688b58b932a79da11836f58ec8335f4153#commitcomment-178765404.

### DIFF
--- a/assets/chezmoi.io/docs/reference/commands/list.md
+++ b/assets/chezmoi.io/docs/reference/commands/list.md
@@ -1,3 +1,5 @@
 # `list`
 
-`list` is an alias for `managed`.
+`list` is an alias for [`managed`][managed].
+
+[managed]: /reference/commands/managed.md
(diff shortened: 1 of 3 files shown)


superuser-does:
Thanks Tom, really appreciate you taking this on board.

Just a thought at the moment (not fully formed) if we wanted to take this UI approach one step further:

I was thinking that if the alias Podman is used, it would automatically set `docker.command` to `podman` for that command by default, unless it has been explicitly set. You couldn't do the reverse of course as it could break backwards compatibility, since the feature is already out.

twpayne:
> I was thinking that if the alias Podman is used, it would automatically set `docker.command` to `podman` for that command by default, unless it has been explicitly set. You couldn't do the reverse of course as it could break backwards compatibility, since the feature is already out.

This was the original behavior. There were separate `chezmoi docker` and `chezmoi podman` commands with `chezmoi docker` command invoking `docker` by default and `chezmoi podman` invoking `podman` by default. See #4612.

Later, I merged them into a single `chezmoi docker` command. See #4643.

I'll let this discussion settle for a bit as I'm not sure what the right thing to do here is. On one side, separate `chezmoi docker` and `chezmoi podman` commands better communicate intent. On the other side, I don't want to add a new top-level chezmoi command every time someone releases a new container runtime.

PR #4949, opened 2026-03-08
feat: Support multiple externals to same target in one .chezmoiexternal file

ronlut:
Implemented #4732


twpayne:
Very nice, thank you!

PR #4956, opened 2026-03-13
docs: Make the AI/LLM stance even more obvious

halostatue:
With recent contributions that people have made using LLM, it is possible that people aren't seeing chezmoi's strict no-LLM policy clearly enough. While this doesn't fix *everything*, it makes the policy clearer by repeating it in places where people would be most likely to see it.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,4 +1,13 @@
 # Contributing
 
-See the [contributing
-guide](https://chezmoi.io/developer-guide/contributing-changes/).
+> [!WARNING]
+>
+> If you use an LLM (Large Language Model, like ChatGPT, Claude, Gemini, GitHub
+> Copilot, or Llama) to make any kind of contribution then you will immediately
+> be banned without recourse. See [`CODE_OF_CONDUCT.md`][coc] for more
+> information.
+
+See the [contributing guide][guide] for full details.
+
+[guide]: https://chezmoi.io/developer-guide/contributing-changes/
+[coc]: https://github.com/twpayne/chezmoi/blob/master/.github/CODE_OF_CONDUCT.md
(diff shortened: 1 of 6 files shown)


bradenhilton:
I think the wording could be disambiguated slightly regarding what exactly is meant by "contribution".

Some users might assume that only refers to things like PRs, and will attempt to debug something with an LLM and report their findings in an Issue or Discussion, which could include outdated practices, buggy/sloppy code etc.

Correct me if I'm wrong, but I think things like getting an LLM to help you rewrite your issue from your native language into English (as long as there's absolutely zero LLM code/debugging (diffs, test cases etc.) are allowed, no?
Or, I don't personally mind that so much, at least, but ultimately it's Tom's repo at the end of the day.

Basically, I see a user trying to submit LLM generated content in two different ways:

1. A user who brazenly does not care about the quality of what they are submitting. They either assume it works perfectly or in other cases are doing it maliciously to contribute things like vulnerabilities in extremely large PRs, or general slop. They don't see it as a big deal if it only partially works. They don't care about the potential maintenance burden introduced by trying to reason about and debug.

2. A user who means well and genuinely wants to help, but perhaps does not understand that what they are suggesting is often buggy, severely outdated or won't even compile. They might not necessarily understand that it can very often be a massive waste of time (time spent trying to reason about LLM slop is time not being spent working on proper solutions etc.). They are acting in what they believe to be good faith in order to find a solution, but don't see how it could potentially be disrespectful to the free time of maintainers.

I would maybe phrase it as something like

```text
chezmoi has a strict zero-tolerance policy on LLM contributions. If you use an
LLM (Large Language Model, like ChatGPT, Claude, Gemini, GitHub Copilot, or
Llama) to make any kind of contribution (such as a Pull Request, Code Review,
or Issue/Discussion comment) then you will immediately be banned without
recourse.
```

(and maybe also add that text-only LLM translations are allowed if that is indeed the case, if not, that's good to know as well)

halostatue:
I don't disagree; I only added the "strict zero tolerance" verbiage.

twpayne:
I'll post a follow-up on this in https://github.com/twpayne/chezmoi/discussions/4010. Let's continue the discussion there.

Issue #4954, opened 2026-03-12
Documentation example improvements


bradenhilton:
## Is your feature request related to a problem? Please describe.

Refs https://github.com/twpayne/chezmoi/issues/4927#issuecomment-4035402811, specifically:

> I plan to make the following changes to chezmoi:
>
>   3. Ideally, ensure that all example configs in the documentation have valid and equivalent JSON, TOML, and YAML examples. I'd like to automate this, and it might take some time.

## Describe the solution you'd like

How do you feel about one of these potential solutions?

1. Add new files for all config formats, each with **all** examples in a centralized location, then including them where needed with something like https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#snippet-sections.
    Pros:
        - no additional processing required (other than naming and using the snippets)
    Cons:
        - adds a layer of abstraction: if a user happens to read the raw markdown files, they won't see the examples, only the snippet syntax
        - multiple sources of truth, so potential sync issues between files and more error prone
        - multiple files to update when the documentation needs to be changed
2. Write all examples in each page in something like TOML, then use templating to round-trip them into the other formats.
    Pros:
        - single source of truth for each example, so no sync issues
        - the contents of the examples *should* be equivalent, so less error prone
    Cons:
        - the round-tripping *may* make examples slightly less readable, depending on how the template functions format the snippets
        - may potentially also be a lot slower to build? Honestly not sure...
3. A hybrid approach: write a single centralized "examples file" in TOML, section into snippets and apply the snippets to relevant pages, then round-trip the snippets into different formats (or alternatively, round-trip TOML file into other formats first, then apply snippets from each file?).
    Pros:
        - single source of truth
        - only need to update one file
        - every format example should be equivalent
    Cons:
        - adds a layer of abstraction
        - round-tripped examples may be slightly less readable
        - may potentially be *even slower*

## Describe alternatives you've considered

NA

## Additional context

NA

KapJI:
Second option is pretty smart, I like it.

twpayne:
I had option 2 in mind. I expect this will require writing a plugin for material-mkdocs.

@bradenhilton if you want to tackle this, please go ahead and assign this issue to you. I'll not have time to work on it for a while.

bradenhilton:
I actually tried something similar to this a while ago (but I was just manually adding every format for each example), but I got busy with other things and the documentation changed quite drastically in some areas to the point where it would have been easier to just start over again.

[...]

PR #4957, opened 2026-03-13
chore: Update Python and Python dependencies

KapJI:
As Python is only used for docs and managed by uv, I think it makes sense to use the latest version. Also remove dev dependency group with duplicated dependencies.

In response to @bradenhilton comment on #4954 that stdlib `tomllib` isn't available in Python 3.10.

Also added step to run Ruff in `test-website` job. It's a bit out of scope of this PR, I can move it to separate one if you want.


twpayne:
Thank you!

PR #4993, opened 2026-04-03
chore: Exclude Python dependencies newer than 5 days

KapJI:
With all recent supply chain attacks I think it makes sense to have it.

It makes newer versions impossible to use even if added manually to `pyproject.toml`.

https://docs.astral.sh/uv/reference/settings/#exclude-newer

Discussion about similar feature in Go: https://github.com/golang/go/issues/76485


twpayne:
Agreed, thanks!

PR #5003, opened 2026-04-10
feat Add spaces before question marks in prompts

### LINKED ISSUE #4983: Remove ?question mark at the and of the Apply suggestion

If you try to find the file path suggested by the Apply command, double clicking on the path also selects the extraneous `?` mark, which makes it harder to find the file as the path is now incorrect

Remove the `?` or move it to the word itself
`?Apply .config/alsd/file.cfg`

Manual cleanup

## Additional context

n/a

twpayne:
Fixes #4983.

### DIFF
--- a/internal/cmd/inittemplatefuncs_test.go
+++ b/internal/cmd/inittemplatefuncs_test.go
@@ -21,7 +21,7 @@ func TestPromptBoolInteractiveTemplateFunc(t *testing.T) {
 			name:              "response_without_default",
 			prompt:            "bool",
 			stdinStr:          "false\n",
-			expectedStdoutStr: "bool? ",
+			expectedStdoutStr: "bool ? ",
 			expected:          false,
 		},
 		{
@@ -29,7 +29,7 @@ func TestPromptBoolInteractiveTemplateFunc(t *testing.T) {
 			prompt:            "bool",
 			args:              []bool{true},
 			stdinStr:          "no\n",
-			expectedStdoutStr: "bool (default true)? ",
+			expectedStdoutStr: "bool (default true) ? ",
 			expected:          false,
 		},
 		{
@@ -37,7 +37,7 @@ func TestPromptBoolInteractiveTemplateFunc(t *testing.T) {
 			prompt:            "bool",
 			args:              []bool{true},
 			stdinStr:          "\n",
-			expectedStdoutStr: "bool (default true)? ",
+			expectedStdoutStr: "bool (default true) ? ",
 			expected:          true,
 		},
 		{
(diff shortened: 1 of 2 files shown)


KapJI:
Are you sure this one is worth fixing? After paths, it makes at least some sense, but here it seems you're adding spaces in many other places, which makes messages look less clean.

Although `?` is allowed in filenames on some filesystems, it's quite rare. A better fix would probably be to configure the terminal to ignore a trailing question mark for paths. At least with iTerm2, editing these filters is simple enough.

To me, this seems more an issue with the terminal than chezmoi.

twpayne:
The reason for adding spaces everywhere is to be consistent, but you're right that this can be fixed in the terminal configuration. I had not thought of that.

@eugenesvk what are your thoughts on this?

eugenesvk:
Oh, there is nothing simple about those filters, precisely because paths can be arbitrary. Besides, how would you filter out `?` only for "paths" and not for natural/log texts that need to select the mark?
 
Also, because of that double click isn't the only selection method, manual **click&drag** happens as well, especially for shorter paths or paths that have **spaces**, so no filter can save you there.
Though for that manual selection extra space is of less help as it's easy to manually overdrag, hence the suggestion to move the`?` to the previous word `Apply?   /pure/path/without/extra/symbols` so your ending selection is more fault-tolerant.

For double clicks, fundamentally, think that the only reliable solution is for chezmoi to mark paths using using enhanced **OSC semantic marks**, then you don't need brittle **text** parsing/filters because terminals would see

KapJI:
@eugenesvk what terminal are you using?

Here is my smart selector in iTerm2 which copies paths (I'm pretty sure it's a default one): `\~?/?([[:alnum:]._-]+/+)+[[:alnum:]._-]*[[:alnum:]_-]/?`. It doesn't match paths with question marks.

And WezTerm also doesn't match trailing question mark by default (in quick select mode).

eugenesvk:
mostly wezterm, but also sometimes native ones. Though the issue is generic: how can you exclude `?` at the end of a path if your terminal doesn't know it's a path? You'd need to exclude them universally, thus blocking non-path selects. Same with spaces

> And WezTerm also doesn't match trailing question mark by default (in quick select mode).

Doesn't work with with spaces, so that's an extra modal barrier that's not universally reliable
But that's the one I meant with my first remark, I had dozens of regex patterns trying to solve this and other issues and even asked for an fancy regex crate to be added to wezterm hoping it [would solve issues like this](https://github.com/wezterm/wezterm/issues/3906)... But it's not powerful enough.

(though maybe with specific chezmoi keywords `apply` I could get a working pattern that avoids the need to find the universal solution)

halostatue:
Just as an FYI, I do not believe that this is currently configurable in ghosTTY.

I'm not able to check for a bit as I'm headed to ✈️.

twpayne:
I don't think this PR is the right fix, so I'll close it.

PR #5012, opened 2026-04-13
chore: Preview docs changes on PRs via Cloudflare Pages

### LINKED ISSUE #4960: Preview doc changes

It'd be nice to have a way to preview docs changes on PR.

I found https://github.com/rossjrw/pr-preview-action which does exactly that. It creates subdir in `gh-pages` branch, so changes will be available at `https://chezmoi.io/pr-preview/pr-[number]/`. Then it can post a comment with link to PR and clean it up when PR is closed. This step can be added to `test-website` job.

It will likely require modifying `site_url` to make links and search work correctly inside subdir and disable `CNAME` copy in `hooks.py`. But this should be pretty trivial.

There are alternative solutions like Netlify but this one uses Github Pages similarly as main website, so no extra platform dependency.

What do you think?

KapJI:
# Motivation

Allow everyone to preview docs changes easily without checking out code and running MkDocs build. For example it makes it easy to check how docs look on mobile devices.

Example: https://github.com/KapJI/chezmoi/pull/2

# Design

Two new workflows are added.

1. `build-docs.yml` (`pull_request` trigger) - runs on PRs that modify docs-related files. Builds the docs and uploads the result as an artifact.
2. `preview-docs.yml` (`workflow_run` trigger) - runs after `build-docs` completes successfully, with access to repo secrets. Downloads the artifact, looks up the PR number via GitHub API, deploys to Cloudflare Pages and creates or updates a comment on the PR with preview link and commit SHA.

Preview URL: `https://pr-<number>.chezmoi-preview.pages.dev`

# Security

- `build-docs` runs on `pull_request` events, which means it executes code from the fork. It has no access to repo secrets. Its outputs like `pr-number.txt` are not trusted.
- `preview-docs` uses `workflow_run` trigger, so its workflow file is always read from the default branch - fork PRs cannot modify the deploy logic. It never checks out fork code, only handles the pre-built static artifact.
- PR number is looked up via GitHub API (`commits/{sha}/pulls`) instead of reading from artifact, so a fork PR cannot spoof the PR number to comment on unrelated PRs.

# Other changes

Added `GITHUB_TOKEN` to `test-website` job so it can be used on forks where `CHEZMOI_GITHUB_TOKEN` isn't available. `GITHUB_TOKEN` has a higher API rate limit than unauthenticated requests. `CHEZMOI_GITHUB_TOKEN` takes priority when available so same-repo behavior is unchanged.

# Considered alternatives

## Uploading artifact from test-website job

Another option was uploading the artifact from the existing `test-website` job in `main.yml`. The problem is that `preview-docs` workflow only starts after the whole `main` workflow finishes, which can take a long time. The new `build-docs` workflow duplicates `test-website` a bit but it's fast (~40s), and I think it's better to have the preview ready 10 minutes earlier.

## GitHub Deployments instead of PR comments

GitHub Deployments were also considered, but the "View deployment" button disappears when a new commit is pushed until the deployment is updated. With a PR comment, the link stays visible at all times and the commit SHA clearly shows whether the preview is up to date.

Closes #4960

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,50 @@
+name: build-docs
+on:
+  pull_request:
+    branches:
+    - master
+    paths:
+    - 'assets/chezmoi.io/**'
+    - 'pyproject.toml'
+    - 'uv.lock'
+    - '.github/workflows/build-docs.yml'
+    - '.github/workflows/preview-docs.yml'
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+env:
+  GO_VERSION: 1.26.2 # https://go.dev/doc/devel/release
+  PYTHON_VERSION: '3.14' # https://www.python.org/downloads/
+  UV_VERSION: 0.11.3 # https://github.com/astral-sh/uv/releases
+jobs:
+  build-docs:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd   # v6.0.2
+      with:
+        persist-credentials: false
+    - uses: ./.github/actions/setup-go
+      with:
+        cache-prefix: website-go
+        go-version: ${{ env.GO_VERSION }}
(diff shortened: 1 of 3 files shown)


twpayne:
This is fantastic. Thank you!

twpayne:
Worked perfectly in #5014, thank you 🚀

PR #5017, opened 2026-04-14
chore: Suppress MkDocs 2 upgrade warning

KapJI:
Suppress the MkDocs 2 upgrade warning by setting `NO_MKDOCS_2_WARNING=1` in task definitions. 
The warning is styled in red, making it look like an error.

Warning is now gone in CI jobs.

The only issue is that because of bash syntax this won't work on Windows, not sure how important is this. I can move it just to CI if you prefer it this way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,9 +15,9 @@ dependencies = [
 ]
 
 [tool.taskipy.tasks]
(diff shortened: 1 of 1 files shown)


twpayne:
Thanks. I'm happy with this being non-Windows-only for now.

PR #5018, opened 2026-04-14
chore: Fix preview-docs PR lookup for forks

KapJI:
Two fixes for `preview-docs` workflow:

1. I noticed `commits/{sha}/pulls` API doesn't find commits that only exist on fork branches. Replaced with the pulls list API filtered by head branch and source repository, which works for both fork and same-repo PRs.

   Verified fields with:

   ```
   gh api repos/twpayne/chezmoi/pulls --jq ".[] | {number, draft, head_ref: .head.ref, head_repo: .head.repo.full_name}"
   ```

2. Add `--paginate` to the PR comment lookup to fetch all comment pages (30 comments per page). Initially I thought docs preview comment is always in the beginning but then I realised that docs changes theoretically can be added after lengthy discussion. Performance difference is negligible anyway.

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -30,11 +30,13 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         GH_REPO: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


twpayne:
Thank you!

PR #5022, opened 2026-04-16
feat: Add .chezmoi.flags template variable

### LINKED ISSUE #5020: Support `.chezmoi.flags` in templates

In my dotfiles, I need to detect when the user is running with `--force` or not to decide when to show prompts, for example.

Here's what I currently do:

```
{{- $chezmoiForce := or (has "--force" .chezmoi.args) (has "--force=true" .chezmoi.args) -}}
```

But as [pointed](https://github.com/twpayne/chezmoi/issues/5010#issuecomment-4246217593) by @twpayne, this doesn't capture all possible cases of `--force`, and is a bit fragile.

Another issue is that flags defined by default in the config file wouldn't be caught as well, although I don't think `--force` can be defined there anyway.

It would be great if chezmoi exposed the resolved flags (probably not just `--force`, although I personally only need `--force`) in a `.chezmoi.flags` variable for users to template with in their scripts.

```
{{- $chezmoiForce := .chezmoi.flags.force -}}
```

This would eliminate the fragility of the alternative method.

See above.

## Additional context

N/A

twpayne:
Fixes #5020.


felipecrs:
This is really awesome. Thanks a lot!

PR #5023, opened 2026-04-16
chore: Only suppress mkdocs 2 warning in CI

twpayne:
Refs #5017.

Thinking about this more, it's nicer if the tasks work on Windows too.

@KapJI does this look correct to you?


KapJI:
Yes, looks good to me.

PR #5030, opened 2026-04-17
chore: Skip preview deploy when docs output is unchanged

KapJI:
I noticed that docs preview gets deployed on PRs that don't actually change anything in the docs. E.g. dependency update triggers the path filter and runs a preview deploy. This causes unnecessary noise and uses Cloudflare quota.

## How it works

`test-website` job already builds the same docs on every master push. It now also uploads a digest of `site/` as a long-retention artifact.

On a PR, `build-docs` computes the same digest from its build. Then `preview-docs` resolves the PR's base SHA, fetches the digest artifact from that specific master run (with fallback to the latest successful master run), and compares. If they match there are no rendered changes so we skip the Cloudflare deploy and post no comment. If they don't match or digest is missing it deploys and comments as before.

Happy to switch the silent skip to an explicit "no docs changes, skipping deploy" comment if you prefer that to be more visible.

There is also a new comment on failure with a link to the run. `preview-docs` runs from the default branch via `workflow_run` so its status doesn't show up as a PR check - without this comment preview failures would be silent on the PR.

Verified both paths on my fork:

- Skip path (no rendered changes): https://github.com/KapJI/chezmoi/actions/runs/24577584704 - digests match, no deploy, no comment.
- Deploy path (rendered change): https://github.com/KapJI/chezmoi/pull/4 - digests differ, preview deployed, comment posted.

## Why `PYTHONDONTWRITEBYTECODE=1`

Consecutive master builds initially gave different digests. MkDocs imports `hooks.py` which creates `__pycache__/` next to it, and mkdocs copies that into `site/`. `.pyc` files are not deterministic and caused different digest each time.

Setting `PYTHONDONTWRITEBYTECODE=1` on the build step skips bytecode generation entirely, which also stops [chezmoi.io](https://chezmoi.io) from serving `.pyc` files: https://chezmoi.io/__pycache__/hooks.cpython-314.pyc

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -43,9 +43,14 @@ jobs:
       env:
         CHEZMOI_GITHUB_TOKEN: ${{ secrets.CHEZMOI_GITHUB_TOKEN }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        PYTHONDONTWRITEBYTECODE: 1
+    - name: compute-docs-digest
+      run: echo "${{ hashFiles('assets/chezmoi.io/site/**') }}" > assets/chezmoi.io/digest.txt
     - name: upload-docs-preview
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a   # v7.0.1
       with:
         name: docs-site-preview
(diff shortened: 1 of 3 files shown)


twpayne:
Thank you again - this is really nice!

PR #5052, opened 2026-05-10
chore: Include v2 in module name

### LINKED ISSUE #5049: availability on pkg.go.dev?
The latest version of chezmoi that I could find on

https://pkg.go.dev/github.com/twpayne/chezmoi/v2

is 2.62.4. Triggering an update doesn't work, probably due to the package having been renamed?

Is there a way to make chezmoi available on pkg.go.dev again? It would help a lot in updating the OpenBSD port. Thank you for considering.

twpayne:
Fixes #5049.

### DIFF
--- a/.config/capslock-summary.json
+++ b/.config/capslock-summary.json
@@ -1,5 +1,5 @@
 {
-  "chezmoi.io/chezmoi": [
+  "chezmoi.io/chezmoi/v2": [
     "CAPABILITY_ARBITRARY_EXECUTION",
     "CAPABILITY_CGO",
     "CAPABILITY_EXEC",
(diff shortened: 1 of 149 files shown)


VlkrS:
Something seems to be still missing ... this gives me a 404 error:

https://pkg.go.dev/chezmoi.io/chezmoi/v2@v2.70.4

twpayne:
Yeah, I have no idea what's missing.

Also, visiting https://pkg.go.dev/chezmoi.io/chezmoi/v2 redirects to https://pkg.go.dev/github.com/twpayne/chezmoi, which is clearly incorrect behavior.

VlkrS:
> Yeah, I have no idea what's missing.
> 
> Also, visiting https://pkg.go.dev/chezmoi.io/chezmoi/v2 redirects to https://pkg.go.dev/github.com/twpayne/chezmoi, which is clearly incorrect behavior.

Maybe this needs to be added to chezmoi.io's index.html?
https://github.com/golang/gddo/wiki/Source-Code-Links

twpayne:
Ah, something in the website generation code changed and so the `go-import` metadata was no longer included. #5091 should fix it.

PR #5063, opened 2026-05-18
install.sh: map armv6/armv7 uname to chezmoi armv6 release

### LINKED ISSUE #5047: Bug: Install script fails with 404 on ARMv6 (Raspberry Pi 1) - Incorrect architecture mapping
**Describe the bug**
The one-line installation script (`sh -c "$(curl -fsLS [https://get.chezmoi.io](https://get.chezmoi.io))"`) fails on Raspberry Pi 1 devices. The script incorrectly attempts to download a generic `linux_arm` tarball which does not exist in the GitHub releases, leading to a 404 error.

**To Reproduce**
Run the install script on an ARMv6 device (e.g., Raspberry Pi 1 or Zero):
```
sh -c "$(curl -fsLS https://get.chezmoi.io) -d"
```

**Actual Output**
```text
debug checking GitHub for tag latest
debug http_download https://github.com/twpayne/chezmoi/releases/latest
debug found tag v2.70.2 for latest
info found chezmoi version 2.70.2 for latest/linux/arm
debug http_download https://github.com/twpayne/chezmoi/releases/download/v2.70.2/chezmoi_2.70.2_linux_arm.tar.gz
curl: (22) The requested URL returned error: 404
debug http_download_curl received HTTP status 404
```

**System Information**
*   **OS:** Raspbian GNU/Linux 13 (trixie)
*   **Kernel:** 6.12.75+rpt-rpi-v6
*   **CPU:** ARMv6-compatible processor rev 7 (v6l)
*   **`uname -m` output:** `armv6l`

**Root Cause Analysis**
The script detects the system as `linux/arm`. However, `chezmoi` releases are specifically named `linux_armv6` and `linux_armv7`. There is no `linux_arm` (v7 by default) asset available, causing the 404.
The latest available/found artifact with the suffix `_linux_arm.tar.gz` is for release [v2.69.2](https://github.com/twpayne/chezmoi/releases/tag/v2.69.2).

**Workaround**
Manually downloading the `linux_armv6` asset works perfectly:
[https://github.com/twpayne/chezmoi/releases/download/v2.70.2/chezmoi_2.70.2_linux_armv6.tar.gz](https://github.com/twpayne/chezmoi/releases/download/v2.70.2/chezmoi_2.70.2_linux_armv6.tar.gz)

**Side note**
I really like this tool, I may only use 5-10% of its capacities but already so great and fits my requirements! ❤️

c-tonneslan:
Fixes #5047. The install script mapped every `armv*l` uname value to `goarch=arm`, which matched the `linux_arm` tarball that existed up to v2.69.2. Starting with v2.70.0 the release naming switched to per-version suffixes (`linux_armv5`, `linux_armv6`) and the bare `linux_arm` tarball is gone, so the install fails with a 404 on Raspberry Pi 1/Zero (`armv6l`) and on `armv7l` hardware too.

This maps `armv6l` to `armv6`, and `armv7l` to `armv6` as well — there's no `linux_armv7` tarball to point at, and armv7 hardware runs the armv6 binary fine.

Fixes #5047

### DIFF
--- a/assets/scripts/install-local-bin.sh
+++ b/assets/scripts/install-local-bin.sh
@@ -138,7 +138,9 @@ get_goarch() {
 	arch="$(uname -m)"
 	case "${arch}" in
 	aarch64) goarch="arm64" ;;
-	armv*) goarch="arm" ;;
+	arm64) goarch="arm64" ;;
+	armv5*) goarch="armv5" ;;
+	arm*) goarch="armv6" ;;
 	i386) goarch="386" ;;
 	i686) goarch="386" ;;
 	i86pc) goarch="amd64" ;;
@@ -158,9 +160,11 @@ check_goos_goarch() {
 	freebsd/amd64) return 0 ;;
 	freebsd/arm) return 0 ;;
 	freebsd/arm64) return 0 ;;
+	freebsd/riscv64) return 0 ;;
 	linux/386) return 0 ;;
 	linux/amd64) return 0 ;;
-	linux/arm) return 0 ;;
+	linux/armv5) return 0 ;;
+	linux/armv6) return 0 ;;
 	linux/arm64) return 0 ;;
 	linux/loong64) return 0 ;;
 	linux/mips64) return 0 ;;
--- a/assets/scripts/install.sh
+++ b/assets/scripts/install.sh
@@ -137,7 +137,9 @@ get_goarch() {
 	arch="$(uname -m)"
 	case "${arch}" in
 	aarch64) goarch="arm64" ;;
-	armv*) goarch="arm" ;;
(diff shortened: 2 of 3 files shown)


halostatue:
The install.sh file is a generated file. This isn't the correct fix.

c-tonneslan:
Good point, didn't realize install.sh was generated. Pushed a fix that edits install.sh.tmpl and regenerates both scripts.

KapJI:
What's the point of this if there is already https://github.com/twpayne/chezmoi/pull/5051?

c-tonneslan:
Closing — didn't realize #5051 was already open touching the templated install.sh source (the right place for the fix). Sorry for the noise.

PR #5064, opened 2026-05-18
fix: Run git lfs pull in working copy directory

overhacked:
When `chezmoi init --git-lfs` runs `git lfs pull`, it should run it in the working copy directory, or Git fails with exit status 128: `Not in a Git repository.`


twpayne:
Thank you!

PR #5070, opened 2026-05-19
chore: Correct #5064

twpayne:
Refs #5064.

### DIFF
--- a/internal/cmd/initcmd.go
+++ b/internal/cmd/initcmd.go
@@ -133,15 +133,15 @@ func (c *Config) runInitCmd(cmd *cobra.Command, args []string) error {
 
 	useBuiltinGit := c.UseBuiltinGit.Value(c.useBuiltinGitAutoFunc)
 
-	gitDirAbsPath := c.WorkingTreeAbsPath.JoinString(git.GitDirName)
-	workingTreeRawPath, err := c.baseSystem.RawPath(c.WorkingTreeAbsPath)
-	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


overhacked:
@twpayne, sorry I got the `RawPath` thing wrong in #5064. I think I misunderstood the concept, thinking that the "raw path" was the fully resolved path to the working copy. I did read through the code of the various storage providers trying to understand. Without taking too much of your time, could you explain or point me to somewhere in the chezmoi code that might help me understand? Interested because I do plan to keep contributing in the future!

twpayne:
There is some information on path handling in https://www.chezmoi.io/developer-guide/architecture/#path-handling, but this doesn't mention `RawPath`.

Basically, `AbsPath` and `RelPath` are used in chezmoi for paths that might be relative to a temporary test directory (using https://github.com/twpayne/go-vfs). However, when using external tooling, the external tooling doesn't know about the temporary test directory and instead needs `RawPath`s. The conversion from `AbsPath` to `RawPath` is already done in `Config.run` at https://github.com/twpayne/chezmoi/blob/64583685c5eb36e10670bad076d5406a08baf751/internal/cmd/config.go#L2721, so it's not needed in the code that calls `Config.run`.

Hope this helps, and thanks for following up with the question!

overhacked:
Thank you, and that actually makes me feel better. I did understand correctly that `RawPath` is the concrete path on disk, but I missed that `Config.run` already does the conversion. Cheers.

PR #5073, opened 2026-05-21
docs: `pager` -> `command` for delta; remove old "Customize" section

hivehand:
I attempted to make `delta` my diff tool by following the documentation instructions and setting `diff.pager` to "delta", but `chezmoi doctor` continued to report that `diff-command` was `not set`. When I followed the other examples, and used `command` instead of `pager`, `chezmoi doctor` reported the tool as found, and `chezmoi diff` used it.

I'm guessing that `pager` was changed to `command` at some point, and the `delta` section was overlooked.

I've also removed the **Customize the diff pager** section, as that also refers to `diff.pager` and seems superseded by the **Use a custom diff tool** section at the top of the page.

### DIFF
--- a/assets/chezmoi.io/docs/user-guide/tools/diff.md
+++ b/assets/chezmoi.io/docs/user-guide/tools/diff.md
@@ -36,13 +36,13 @@ To use [VSCode][vscode] as the diff tool, add the following to your config:
 
 ## Use delta as the diff tool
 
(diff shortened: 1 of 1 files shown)


KapJI:
Using it as `pager` is correct. More context here: https://github.com/twpayne/chezmoi/pull/4349

twpayne:
Closing as the PR is not correct.

PR #5104, opened 2026-06-22
chore: generate sigstore bundles rather than separate signature

scop:
The latter is deprecated as of cosign 3.1.1,
ref https://github.com/sigstore/cosign/releases/tag/v3.1.1


twpayne:
Thank you very much for this! There a couple of tweaks needed:

* Bump the cosign version in `.github/workflows/main.yml`.
* Fix the format of the commit message.

I'll merge this as-is, and then fix these with an edited commit on `master`.

PR #5115, opened 2026-07-04
chore: Deduplicate script cmd setup

xander-hirsch:
Create a shared helper function to setup the temporary directory/script and build an `exec.Cmd` struct value.

The duplicate implementations in `RealSystem.RunScript` and `SourceState.newModifyTargetStateEntryFunc` are nearly identical. Prefer the `RealSystem.RunScript` where different.


twpayne:
Thank you :)

PR #5119, opened 2026-07-05
chore: Deduplicate read file and apply template

xander-hirsch:
Create a shared helper function to read the contents of a source file and apply the contents as a template if necessary.

Also, the function type `func() (byte[], error)` and `func() (byte[], error)` are used frequently throughout the code base, so make explicit `ProduceBytesFunc` and `Produce32BytesFunc` types.

### DIFF
--- a/internal/chezmoi/actualstateentry.go
+++ b/internal/chezmoi/actualstateentry.go
@@ -32,7 +32,7 @@ type ActualStateDir struct {
 type ActualStateFile struct {
 	absPath      AbsPath
 	perm         fs.FileMode
-	contentsFunc func() ([]byte, error)
+	contentsFunc ContentsFunc
 }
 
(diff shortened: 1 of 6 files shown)


twpayne:
Thanks for this. I agree that the de-duplication is helpful. Please rename `ProducesBytesFunc` to `ContentsFunc` and `Produces32BytesFunc` to `ContentsSHA256Func` though.

xander-hirsch:
I applied the suggested rename and force pushed the commit. Is that the proper way to respond to PR feedback?

twpayne:
Yes, that's perfect for small changes, thank you.

Another option is to add another commit that makes the required change, which is easier to review if there are a lot of changes. Some projects (including chezmoi) will then ask you to create a clean commit history before merging. For small changes, force-pushing combines these two steps, which is fine.

twpayne:
Thanks!

Issue #5110, opened 2026-07-01
Support Variant of `modify_` to Prompt for Overwrite


xander-hirsch:
## Is your feature request related to a problem? Please describe.

I use multiple VS Code derivatives (mostly [Microsoft VS Code](https://code.visualstudio.com/) and [Kiro](https://kiro.dev/)) between my personal and work setup. These editors are very similar and I wish to centralize my configuration to a single source of truth in chezmoi and allow for untracked machine specific variants.

This is complicated by the fact that these applications update `settings.json` without me directly editing it from the chezmoi source, which will cause drift between the source and destination. My current use of `modify_` scripts introduces the risk that `chezmoi apply` wipes the drift that I would want to capture in my chezmoi source.

## Describe the solution you'd like

Add a file source state attribute `derive_` which is similar to `modify_` but with the following distinctions:
- Prompt the user when an destination file overwrite is detected
- No standard input

I believe my proposed attribute can serve two general purpose goals:
- Create certain dotfiles where there is an odd confluence of circumstances (e.g. the application does not support multiple configuration files *and* the application directly edits the file)
- The user wants to maintain the configuration in one format but the application only supports another (e.g. the user uses YAML or Jsonnet building blocks to separate logical concerns but the application only supports a single JSON)

## Describe alternatives you've considered

My [dotfiles](https://codeberg.org/xander-hirsch/dotfiles) already have approaches to handling untracked, machine specific configs.

I have read the documentation for [Manage machine-to-machine differences](https://www.chezmoi.io/user-guide/manage-machine-to-machine-differences/) and I use those techniques where possible. However, there are certain items (such as work specific git configurations) where I do not want to deal with the hassle of making sure I do not inadvertently leak internal company data.

### When the Configuration Tool **Supports** Include

This is the easiest case for me to handle. I simply include an "include local" entry in the chezmoi tracked configuration file.

For example, I have the file `private_dot_config/git/create_empty_config.local` and my `private_dot_config/git/config.tmpl` has the snippet
```
[include]
    path=config.local
```

That allows me to edit `~/.config/git/include.local` which remains outside of chezmoi but is still ultimately picked up by git.

### When the Configuration Tool **Does Not Support** Include

Admittedly, the following pattern seems more like a hack.

Git only allows for one user-global ignore file. My `private_dot_config/git` includes the following three files:
```
.ignore.dist.tmpl
create_empty_ignore.local
modify_ignore.sh
```

[...]

PR #5121, opened 2026-07-05
chore: Deduplicate parse script extensions

xander-hirsch:
Deduplicate the logic for mapping file extensions to interpreters and stripping the extension if it matches an interpreter. This is currently used by run and modify types.

**Review note:** The deduplicated logic here is a smaller scope than #5115 or #5119. However, I felt it would be best to submit this as a standalone PR before attempting a proposal for a `derive_` attribute as discussed in #5110. That implementation will also rely on the logic deduplicated here.

### DIFF
--- a/internal/chezmoi/sourcestate.go
+++ b/internal/chezmoi/sourcestate.go
@@ -2152,6 +2152,34 @@ func (s *SourceState) newSymlinkTargetStateEntryFunc(
 	}
 }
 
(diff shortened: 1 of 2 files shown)


twpayne:
Thanks, but I think the de-duplicated code is more complicated than the code it replaces.

If we end up adding a `derive_` attribute as described #5110 then I'll reconsider this PR, but I'll close this for now.

Issue #5111, opened 2026-07-02
Copyright status of install scripts


daniel-baumann:
Hi,

I've uploaded chezmoi to debian: https://tracker.debian.org/chezmoi

Before the package was alled into Debian, it passed a license review, where it was [pointed out that](https://dfsg-new-queue.debian.org/reviews/chezmoi/2.70.5-1) that in [debian/copyright](https://tracker.debian.org/media/packages/c/chezmoi/copyright-2.70.5-1) should be ammeded by the authors of which the install scripts have been "inspired by".

debian/copyright tracks copyright relevant information. If you have been inspired by the idea of something/someone, this is not copyright revelant (patents are about ideas, copyright is about the implementation). If you have borrowed beyond fair-use, copied or translated parts (e.g. from one script language to another), then that would be copyright relevant and we would need to track down the license of the original works and add the other people (and licenses) to debian/copyright.

I'm aware that it's anoying topic, but would you mind elaborating to what extend you've been "inspired"?

Thanks,
Daniel

twpayne:
Firstly, thank you so much for adding chezmoi to Debian. This is really appreciated!

I'm very happy to add copyright attributions for this.

For the "inspired by":

* https://github.com/client9/shlib is licensed under the "[Unlicense](https://github.com/client9/shlib/blob/master/LICENSE.md)", which, if I understand correctly, means that no copyright attribution is required. I did try to [contribute back to the project](https://github.com/client9/shlib/pull/30) but at the time the project seemed dead and unmaintained. There have been some signs of life some years later.

* https://github.com/goreleaser/godownloader is [MIT licensed](https://github.com/goreleaser/godownloader/blob/master/LICENSE.md). It [uses components from github.com/client9/shlib](https://github.com/goreleaser/godownloader/blob/master/makeshellfn.sh).

For sure, Debian understands when a copyright attribution is needed much better than I do, so please tell me what the right thing to do here is.

larsw:
@twpayne this documentation isn't correct (yet) then? https://www.chezmoi.io/install/#__tabbed_2_3

daniel-baumann:
> Firstly, thank you so much for adding chezmoi to Debian. This is really appreciated!

welcome, but thanks for the excellent chezmoi in the first place! ;)

>     * https://github.com/client9/shlib is licensed under the "[Unlicense](https://github.com/client9/shlib/blob/master/LICENSE.md)", which, if I understand correctly, means that no copyright attribution is required.

[...]

PR #5126, opened 2026-07-10
fix: Adding copyright and license identifiers in install scripts (Closes: #5111).

daniel-baumann:
Hi,

as discussed in #5111, it would be nice to add the copyright and license identifiers to keep the "provenance" selfcontained in the sources.

Regards,
Daniel

### DIFF
--- a/assets/scripts/install-local-bin.sh
+++ b/assets/scripts/install-local-bin.sh
@@ -5,6 +5,10 @@
 # https://github.com/client9/shlib
 # https://github.com/goreleaser/godownloader
 
+# Copyright (C) 2017 Nick Galbreath
+# Copyright (C) 2018 Tom Payne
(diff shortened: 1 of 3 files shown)


twpayne:
Thanks for this! Both install scripts are generated from a template, so you need to modify [`internal/cmds/generate-install.sh/install.sh.tmpl`](https://github.com/twpayne/chezmoi/blob/bd880914c19817b03567573e474254ac2ad7d84a/internal/cmds/generate-install.sh/install.sh.tmpl) and then run `go generate`.

Also, could you change the commit message from `fix:` to `chore:` as this is not a user-visible functionality fix.

daniel-baumann:
sure, thanks!

twpayne:
Thanks very much!

daniel-baumann:
welcome, thanks to you :)

Issue #5124, opened 2026-07-08
Print `chezmoi` version before an upgrade commenced


dimsanius:
## Is your feature request related to a problem? Please describe.

When running `chezmoi upgrade`, there is no indication which version the upgrade happened from. It prints only the version after an upgrade has been completed.

## Describe the solution you'd like

When running `chezmoi upgrade`, it would be good to see the version that was present before an upgrade.
Getting inspiration from the `uv` toolset:

```shell
Upgraded uv from vx.y.z to vx.y.z! https://github.com/link-to-released-version
``` 

Based on that, it could be something like:

```shell
Upgraded from chezmoi version vx.y.z, commit HASH, built at DATETIME, built by goreleaser to chezmoi version vx.y.z, commit HASH, built at DATETIME, built by goreleaser
```

Maybe it could be worthwhile having some kind of a less verbose version print just to ensure output is understood easily:

```shell
Upgraded from chezmoi version vx.y.z to chezmoi version vx.y.z
```

## Describe alternatives you've considered

For now my only option is to run `chezmoi --version` myself before running `chezmoi upgrade`

## Additional context

N/a

axmc1:
I'd be willing to work on this, I believe it's not so complicated.
About the verbose output, while a simple inline switch would do, I believe it's not really the right option. If you forget to append the switch before upgrading, it'll probably be too late to try to get that information afterwards. If we wanted something like that, we'd need a permanent switch in an environment variable or config file.

dimsanius:
Thanks for jumping on this!

I will share my use case to add some clarity to the table:

I have a script that updates various parts of the system in one single script execution, e.g. apt packages, snap packages, omz, uv, chezmoi. I dont run it automatically. I prefer running it manually. I am looking at the terminal output everytime I run thr command just to ensure outputs dont contain any errors. With the current chezmoi output, it might be a bit troublesome to understand at a glance as the output of `chezmoi --version` is quite massive.

In my case, extra flag for `--version` could work to make output of it less verbose. However, I believe it might be worthwhile doing it the other way, where simple`--version` argument gives a less verbose output, while something along the line of `--full-version` or `--version --full` gives a much more verbose output.

Again, I am glad you decided to jump on this ticket! I am already happy with your proposed implementation. I might be nitpicking at this stage with my further proposal.

PR #5129, opened 2026-07-12
chore: Print previous and new version when upgrading

axmc1:
This PR solves #5124.

During `chezmoi upgrade`, it now calls `chezmoi --version` before swapping the executable (old version) and after (new version), getting both version strings to print a summary of the upgrade at the end.

In the issue I talked about switching between a simple and verbose version of the message. I decided to leave that feature out and only use the output from `chezmoi --version` since that has always been the output.

### DIFF
--- a/internal/cmd/upgradecmd.go
+++ b/internal/cmd/upgradecmd.go
@@ -121,12 +121,6 @@ func (c *Config) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 	}
 	c.logger.Info("upgradeMethod", slog.String("executable", c.upgrade.executable), slog.String("method", method))
 
-	// get old chezmoi version for logging later
-	oldChezmoiVersion, err := getChezmoiVersion(executableAbsPath.String())
-	if err != nil {
-		return err
-	}
-
 	// Replace the executable with the updated version.
 	switch method {
 	case upgradeMethodBrewUpgrade:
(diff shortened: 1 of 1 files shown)


internal/cmd/upgradecmd.go
```diff
@@ -121,6 +121,12 @@ func (c *Config) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 	}
 	c.logger.Info("upgradeMethod", slog.String("executable", c.upgrade.executable), slog.String("method", method))
 
+	// get old chezmoi version for logging later
+	oldChezmoiVersion, err := getChezmoiVersion(executableAbsPath.String())
+	if err != nil {
+		return err
+	}
```
twpayne: The "old" version of chezmoi is the version that's currently executing. You don't need to run `chezmoi --version` to get it as it's already available in `c.versionInfo`.

internal/cmd/upgradecmd.go
```diff
@@ -164,11 +170,9 @@ func (c *Config) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 	}
 
 	// Execute the new version.
-	chezmoiVersionCmd := exec.Command(path, "--version")
-	chezmoiVersionCmd.Stdin = os.Stdin
-	chezmoiVersionCmd.Stdout = os.Stdout
-	chezmoiVersionCmd.Stderr = os.Stderr
-	return chezmoilog.LogCmdRun(c.logger, chezmoiVersionCmd)
+	chezmoiVersion, err := getChezmoiVersion(path)
```
twpayne: Errors should be handled immediately.

internal/cmd/upgradecmd.go
```diff
@@ -324,3 +328,15 @@ func getReleaseAssetByName(rr *github.RepositoryRelease, name string) *github.Re
...
 	return nil
 }
+
+func getChezmoiVersion(executablePath string) (string, error) {
+	buf := new(bytes.Buffer)
+	versionCmd := exec.Command(executablePath, "--version")
+	versionCmd.Stdin = os.Stdin
+	versionCmd.Stdout = buf
+	versionCmd.Stderr = os.Stderr
+	if err := chezmoilog.LogCmdRun(nil, versionCmd); err != nil {
```
twpayne: `chezmoilog.LogCmdOutput` already captures the output for you.

twpayne (changes requested):

internal/cmd/upgradecmd.go
axmc1: Replaced `getChezmoiVersion()` with `c.versionStr`.

internal/cmd/upgradecmd.go
axmc1: Handled error immediately.

internal/cmd/upgradecmd.go
axmc1: Replaced original `LogCmdRun` with `LogCmdOutput`, making `getChezmoiVersion()` obsolete, so that was removed with it.

twpayne:
Thanks, but I prefer the solution in #5131.

Issue #4759, opened 2025-11-09
Shift from Material for MkDocs to Zensical


halostatue:
This is mostly a tracking issue since [Zensical](https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/) was announced a few days ago and they have put Material for MkDocs into maintenance mode as they are spending their time on Zensical.

There's a few plugins being used (I added a couple) which may or may not be compatible, so we should plan a migration.

I'll add more to this when I know more.

twpayne:
Thank you for tracking this!

twpayne:
There seems to be a lot of [controversy around mkdocs](https://fpgmaas.com/blog/collapse-of-mkdocs/). Zensical might not be eventual winner.

I agree that chezmoi will need to migrate from Material for MkDocs to something else in the future, but I propose to let the ecosystem stabilize first. For now, the current documentation system works, so I'll close this issue for now.

PR #5145, opened 2026-07-23
chore: Render Mermaid diagrams with mkdocs-material

kytta:
I first wanted to migrate docs to Zensical, but from what I see in #4759 and https://zensical.org/compatibility/features/, it's not yet feature-complete to be able to replace mkdocs-material. Nevertheless, I started looking at the docs infra, and I noticed that the Mermaid diagram rendering plugin is technically not necessary, as Material for MkDocs already includes this functionality.

This PR removes `mkdocs-mermaid2-plugin` and replaces it with mkdocs-material built-in Mermaid diagram renderer. This reduces the amount of dependencies (getting rid of the frankly unneeded `setuptools` dep) and also removes the need in custom light/dark mode handling.

However, this does have one major change: The color of the actors in the sequence diagram is different:

| Before | After |
| ------ | ----- |
| Actors have a purple background | Actors have a white background

This is due to Material having [their own Mermaid styles](https://github.com/squidfunk/mkdocs-material/blob/b3e6dd886a974aa8200759ecfd7db28c598a2894/src/templates/assets/javascripts/components/content/mermaid/index.css). I find the new style to be fine, but if you dislike this change, I could also add custom CSS styles to "fix" this.

### DIFF
--- a/assets/chezmoi.io/mkdocs.yml
+++ b/assets/chezmoi.io/mkdocs.yml
@@ -385,7 +385,7 @@ markdown_extensions:
     custom_fences:
     - name: mermaid
       class: mermaid
-      format: !!python/name:mermaid2.fence_mermaid
+      format: !!python/name:pymdownx.superfences.fence_code_format
 - pymdownx.tabbed:
     alternate_style: true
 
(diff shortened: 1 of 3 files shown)


twpayne:
Thank you very much for this very nice cleanup! The new default Mermaid style is fine.

PR #5150, opened 2026-07-25
chore: Update dependencies

twpayne:


twpayne:
This is really annoying and weird. One of the dependency updates causes the `main / test-macos` Github Action step to time out. I'm trying to find out which one.

PR #5156, opened 2026-07-27
chore: Fix potential code injection in preview-docs.yml

twpayne:
This was found by an independent security audit of chezmoi.

In the `preview-docs.yml` workflow, an attacker controls the branch name in `${HEAD_BRANCH}`. Git branch names can include the `"` character, which allows them to escape the argument passed to `--jq` and get local shell access.

cc @KapJI for information

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -36,8 +36,10 @@ jobs:
         HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
         HEAD_REPO: ${{ github.event.workflow_run.head_repository.full_name }}
       run: |
-        PR=$(gh api "repos/${GH_REPO}/pulls" --paginate \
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-docs.yml
```diff
@@ -36,8 +36,10 @@ jobs:
         HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
         HEAD_REPO: ${{ github.event.workflow_run.head_repository.full_name }}
       run: |
-        PR=$(gh api "repos/${GH_REPO}/pulls" --paginate \
-          --jq ".[] | select(.head.ref == \"${HEAD_BRANCH}\" and .head.repo.full_name == \"${HEAD_REPO}\") | {number, base_sha: .base.sha}")
+        PR=$(gh api "repos/${GH_REPO}/pulls" --paginate | \
```
KapJI, line 39: It's possible to open mutiple PRs from the same branch to different target branches. In that case `head -1` may return wrong PR.

This should make it a bit more robust.
```suggestion
        PR=$(gh api "repos/${GH_REPO}/pulls?base=master" --paginate | \
```

KapJI (commented):
Thanks for fixing it!

PR #5176, opened 2026-08-11
fix: sanitize control characters in diff and status path display

### LINKED ISSUE #5175: chezmoi diff and status don't strip terminal escape sequences from displayed paths, allowing injection via a crafted source file name
Affected versions: confirmed on the tip of `master`, commit d686011 (2026-08-09)

## Summary

`chezmoi diff` and `chezmoi status` print the path of each affected target file with no filtering of control characters. A source state entry's name comes directly from the dotfile repository chezmoi was initialized from (`chezmoi init <repo>`), which has no character restrictions. Since previewing pending changes with `chezmoi diff` or `chezmoi status` before running `chezmoi apply` is chezmoi's own documented, routine workflow, initializing from a repository containing a crafted file name causes any escape sequence embedded in that name to be written to the terminal and interpreted live the next time its status is checked.

## Details

`internal/chezmoi/diff.go`, `gitDiffFile.Path()`, consumed by go-git's `diff.UnifiedEncoder` to build `chezmoi diff`'s header lines:
```go
type gitDiffFile struct {
	hash     plumbing.Hash
	fileMode filemode.FileMode
	relPath  RelPath
}

func (f *gitDiffFile) Path() string            { return f.relPath.String() }
```

`internal/cmd/statuscmd.go`, `runStatusCmd`'s `preApplyFunc`, building `chezmoi status`'s default output:
```go
if x != ' ' || y != ' ' {
	var path string
	switch pathStyle := c.Status.PathStyle.String(); pathStyle {
	case pathStyleAbsolute:
		path = c.DestDirAbsPath.Join(targetRelPath).String()
	case pathStyleRelative:
		path = targetRelPath.String()
	default:
		return fmt.Errorf("%s: invalid path style", pathStyle)
	}

	fmt.Fprintf(&builder, "%c%c %s\n", x, y, path)
}
```
Neither site filters control characters from the path before printing.

## POC

(available upon request)

## Impact

Both `chezmoi diff` and `chezmoi status` write a source state entry's name to the terminal exactly as read from the dotfile repository, with no stripping of control characters or escape sequences. Since initializing from and previewing an unaudited dotfile repository is chezmoi's own documented primary workflow, a crafted file name in that repository causes any embedded escape sequence to be written to the terminal and interpreted live. I demonstrated this with a benign OSC title-set marker rather than a more aggressive payload, since confirming the sequence reaches output unmodified is sufficient to establish the defect; the specific range of what an injected OSC/CSI sequence can achieve depends on the victim's terminal emulator and wasn't enumerated further.

I have a fix ready and will follow up with a PR.

carfeii:
Fixes #5175.

`chezmoi diff` and `chezmoi status` print a target file's path with no filtering of control characters. A source state entry's name comes directly from the dotfile repository chezmoi was initialized from, which has no character restrictions, so a crafted file name can inject terminal escape sequences that get written to the terminal and interpreted live on an ordinary diff or status check, before apply is ever run.

Adds `SanitizePathForDisplay` (`internal/chezmoi/diff.go`), replacing control characters with the Unicode replacement character, applied at `gitDiffFile.Path()` (used by go-git's `diff.UnifiedEncoder` to build `diff`'s header lines) and at the one `fmt.Fprintf` call in `statuscmd.go`'s `preApplyFunc`.

I applied this only at the display point, not by sanitizing `RelPath` itself or where it's first read from the source state, since `RelPath` is used extensively for real filesystem operations throughout the codebase. This is safe because Go passes `RelPath` by value: the caller's own copy used for filesystem operations is unaffected by what `gitDiffFile.Path()` returns.

Verified against a scratch config with a source file named using a raw OSC title-set escape sequence, output captured directly to a file and inspected as a hex dump, for both `diff` and `status`. Confirmed no regression with a normal filename (`dot_bashrc` still renders as `.bashrc` in both outputs). `internal/chezmoi` and `internal/cmd` test suites pass.

One thing worth a follow-up separate from this PR: a few other call sites build interactive confirmation-prompt text from the same underlying path value (`internal/cmd/config.go`'s apply confirmation, `addcmd.go`, `readdcmd.go`, `forgetcmd.go`) and likely share this issue. I didn't touch those here since I wasn't able to verify them the same way (harder to reproduce without driving interactive stdin), so didn't want to claim a fix I hadn't actually confirmed.

### DIFF
--- a/internal/chezmoi/diff.go
+++ b/internal/chezmoi/diff.go
@@ -4,6 +4,7 @@ import (
 	"io/fs"
 	"net/http"
 	"strings"
+	"unicode"
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/filemode"
@@ -11,6 +12,23 @@ import (
 	znkrdiff "znkr.io/diff"
 	znkrtextdiff "znkr.io/diff/textdiff"
 )
+
+// SanitizePathForDisplay replaces control characters (which can include
+// terminal escape sequences) with the Unicode replacement character. A
+// source state entry's name comes directly from a dotfile repository,
+// which has no character restrictions, and chezmoi supports applying a
+// dotfile repository the user hasn't audited yet (chezmoi init followed
+// by chezmoi diff, a routine preview step). Applied only to the path
+// string returned to go-git's diff encoder for display; callers still
+// hold the unmodified RelPath for actual filesystem operations.
+func SanitizePathForDisplay(s string) string {
+	return strings.Map(func(r rune) rune {
+		if unicode.IsControl(r) {
+			return '�'
+		}
+		return r
+	}, s)
+}
 
 var gitDiffOperation = [...]diff.Operation{
 	znkrdiff.Delete: diff.Delete,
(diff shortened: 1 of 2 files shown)


twpayne:
This PR is nonsense and incomplete. See https://github.com/twpayne/chezmoi/issues/5175#issuecomment-5258925779.

PR #5196, opened 2026-09-02
chore: Change XDG_SHARE_HOME to XDG_DATA_HOME for sourceDir

wushenrong:
The sourceDir variable in the docs incorrectly shows that you should set XDG_SHARE_HOME for the directory where user-specific data should be written to while XDG_DATA_HOME is the correct environment variable name for XDG Base Directories.


twpayne:
Thanks!

Issue #3689, opened 2024-04-09
Several issues installing on a vanilla machine


peterbraden:
## Describe the bug

Using the install one liner (`$ sh -c "$(curl -fsLS get.chezmoi.io)" -- init --apply $GITHUB_USERNAME`) chezmoi failed at several steps, which has led to a state where chezmoi isn't installed, but successive runs of the one-liner think they have succeeded.

## To reproduce

On a brand new mac `sh -c "$(curl -fsLS get.chezmoi.io)" -- init --apply peterbraden`

This is on Ventura 13.6.6

This isn't really that reproduceable, so let's work backwards from the end state below.

## Expected behavior
It's expected that the script end with chezmoi installed.

## Output of command with the `--verbose` flag

```console
$ chezmoi --verbose $COMMAND
zsh: command not found: chezmoi
```

## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
zsh: command not found: chezmoi
```

</details>

## Additional context

Ok, let's dive into the meat of this:

When installing, several fatal errors caused the script to abort:

### 1. A permissions error with a git clone.
My dotfiles have subrepos. At the time that I installed, I hadn't setup an SSH key (all my dotfiles are public) thus the clone was without github authentication.

```
Cloning into '/Users/peterbraden/.local/share/chezmoi/home/dot_vim/external_bundle/vim-go.vim'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

This is potentially not a chezmoi issue (although all of the repos are public so this _should_ perhaps work), but the issue is that a subsequent run should have attempted the clone again, rather than assuming it succeeded.

### 2. 
An issue in one of my install scripts (`brew: command not found`) meant that _my_ script aborted with an error - this _isn't_ a chezmoi bug, but a successive run should have run my script again.

### 3.
A subsequent attempt 'assumes' that the former run succeeded: 
```
sh -c "$(curl -fsLS get.chezmoi.io)" -- init --apply peterbraden
info found version 2.47.3 for latest/darwin/arm64
info installed ./bin/chezmoi

echo $?
0
``` 

# Solution
In writing this bug report I think I've found that it's actually probably my issue. ˜/.bin isn't in my path, so obviously chezmoi isn't found. I will close this issue.

Issue #2993, opened 2023-05-11
[Feature] Ensure chezmoi binary is in current `$PATH` while running scripts.


aarondill:
## Is your feature request related to a problem? Please describe.

When loading chezmoi in a new machine, running the chezmoi install script, chezmoi installs into `~/bin/chezmoi`, which is not in my 
PATH, as a new Ubuntu machine includes `~/bin` *only* if it exists when a new shell is loaded. Since my scripts call `chezmoi` (especially `chezmoi source-path`. I know `$CHEZMOI_SOURCE_DIR` is set, but I like to manually run my scripts with `bash SCRIPT`, so running the `chezmoi` binary allows universality). Because `~/bin` is not in my `$PATH`, running `chezmoi apply` fails with a `chezmoi: command not found`.

## Describe the solution you'd like

When running scripts, detect if the binary's parent directory is in the `$PATH`, and if not, then **append** it.

Another solution for potential safety is to create a temp dir and link the binary (there's a reason that not *every* path is in `$PATH`). This is implemented in bash, but hopefully you get the gist:
```bash
set -eu
temp_dir="$(mktemp -d)"
trap 'rm -rf "$temp_dir"' EXIT
ln -sT "$CHEZMOI_EXECUTABLE" "$temp_dir/chezmoi"
PATH="$PATH:$temp_dir"
```

## Describe alternatives you've considered

```bash
"${CHEZMOI_EXECUTABLE:-chezmoi}" ARGS
```

This works, but is not very clear, and would require that I change *every* one of my scripts. I could automate this, but I would prefer that `chezmoi ARGS` would *just work*™️

## Additional context

```console
> chezmoi apply
/tmp/1402241099.whyyyyyyy.sh: line 2: chezmoi: command not found
chezmoi: exit status 127
> chezmoi apply
/tmp/1402241099.whyyyyyyy.sh: line 2: chezmoi: command not found
chezmoi: exit status 127
> PATH="$PATH:$(dirname -- "$(which chezmoi)")" chezmoi apply
....
> # Finally
```

twpayne:
Changing the user's `$PATH` is a no-no for many reasons, including security and transparency.

You can set environment variables for scripts, including `$CHEZMOI_EXECUTABLE`, with the `scriptEnv` configuration variable.

aarondill:
[...]

Issue #2064, opened 2022-05-11
Command not found after running install script


RobThree:
I have installed chezmoi as per [the instructions](https://www.chezmoi.io/install/). However, this doesn't work; running chezmoi results in: `command not found: chezmoi`

## To reproduce

1. Login to machine
2. Run `sh -c "$(curl -fsLS chezmoi.io/get)"`
3. Observe:
```
info found version 2.15.4 for latest/linux/amd64
info installed ./bin/chezmoi
```
4. Run `chezmoi init`
5. Observe:
```
zsh: command not found: chezmoi
```

## Expected behavior

I expect the command to work regardless of the current working directory.

## Output of command with the `--verbose` flag

```
zsh: command not found: chezmoi
```
## Output of `chezmoi doctor`

<details>

```console
$ chezmoi doctor
zsh: command not found: chezmoi
```

</details>

## Additional context

I've ran this command on Debian Bullseye (11) with zsh as default shell from my home directory. This results in chezmoi being installed in `/home/rob/bin/`. When I run this (with sudo) from the root directory then `chezmoi` is placed in `/bin/` where it belongs (I guess?).

I guess either the documentation should state clearly the install script should be run from `/` (undesirable) or the installscript should use `/bin` instead of `./bin`.

RobThree:
I just switched to bash and ran the install script again from my home directory; same problem.

RobThree:
Turns out zsh doesn't run `~/.profile` where `$HOME/bin` is added to path...

PR #5207, opened 2026-09-17
PATH notification

### LINKED ISSUE #5206: Addressing PATH caveats

I tried installing chezmoi on a fresh Ubuntu Studio installation today and became extremely frustrated with the documented installation procedure failed. Specifically, the fact that the `curl | sh` installation installs to `./bin`, which is not in the `PATH` environment variable by default.

This has been filed before in #2064, #2993, and #3689. However, I don't propose changing the user's `PATH`.

I recommend a two-pronged approach:

1. Detect when the chezmoi has been installed in a directory that doesn't exist in `PATH`, then print a message to the user identifying the actual installed path and recommend that they add it to their `PATH` var. Something like:

```txt
Installed chezmoi into /home/example/bin/.
This directory is not in your current PATH.
You can run it now using "/home/example/bin/chezmoi".
See [documentation link] to make it available as "chezmoi".
```

This way, there's no need to automatically edit the user's shell configuration, as #2993 proposed.

2. As part of that [documentation link] above, explain in the installation guide that the installer does not update `PATH` and offer instructions on how to do so so the user can invoke "chezmoi" from anywhere.

Please let me know if you'd like me to open a PR with these minor additions.

Alternative: Not doing anything. Most seasoned Linux veterans will know how to get around the incomplete documentation, but new users who are still learning Linux will get frustrated that chezmoi doesn't work out of the box even while following the instructions to the letter.

Given that this has been an ongoing problem for over four years now, I think the above mentioned steps are minor enough to be incorporated if it means helping out new users.

## Additional context

See above for the links to previous reports about this issue, which should be more than enough additional context.

SturmB:
Closes #5206.

This adds a simple safeguard to ensure users understand that they may need to update their `PATH` variable in their shell's configuration before they can invoke chezmoi from anywhere without specifying the absolute path to it every time. It makes no changes to the functionality of chezmoi itself. It merely prints a message after installation to inform the user to check their `PATH` var.

The testing for this is in the GitHub Actions workflow and ensures that the notification appears when the path to chezmoi is not in their `PATH` var and that it doesn't appear otherwise.

This utilizes shell scripting only, so it's fairly simple. I know PHP and Laravel like the back of my hand, but I'm still learning shell scripting, so please look over these changes and see if I've missed anything. I hope this will also address the concerns raised in #2064, #2993, and #3689.


twpayne:
Thanks! This looks good. Please can you squash all the commits into a single commit with the message `feat: Notify user if chezmoi is not in their PATH in install script`.

twpayne:
Thank you!

